PR #58, opened 2023-02-10
Run unit tests on Windows as well

fnesveda:
Since we test a lot of the local behavior of the SDK in unit tests, we should run those tests on Windows as well.

This PR adds Windows in the matrix by which we run the unit tests, and refactors the workflows a bit to accommodate that.

It also fixes some unit tests which were failing on Windows because the `datetime.now()` resolution is lower there, and if you ran two `datetime.now()` calls in quick succession, they would return the same value, which caused some of the timestamp checking tests in memory storage to fail.

### DIFF
--- a/.github/workflows/lint_and_type_checks.yaml
+++ b/.github/workflows/lint_and_type_checks.yaml
@@ -1,11 +1,11 @@
-name: Lint and test
+name: Lint and type checks
 
 on:
   workflow_call:
 
 jobs:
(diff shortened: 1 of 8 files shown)


jirimoravcik (approved):
maybe rename `release.yml` to `release.yaml`?

fnesveda:
@jirimoravcik Good point, I will do it in a separate change.

PR #59, opened 2023-02-10
Add more logging to flaky request queue integration test

fnesveda:
Every now and then, the integration test for request queue times out. This adds more logging to the test so that we can debug it easier.


jirimoravcik (approved):
maybe it's flaky because request queue is slow :D

fnesveda:
@jirimoravcik Yeah, it's not the fault of the SDK, it's just the first `GET http://api.apify.com/v2/request-queues/QUEUE_ID` call that gets stuck, dunno why.

PR #69, opened 2023-02-15
Add logger

fnesveda:
This PR adds a new submodule, `apify.log`, and in it a `logger`, which is just a standard logger from the `logging` library, with name `apify` as per best practices for Python logging. It also makes it available under the `Actor.log` property, for convenience.

The main magic happens in the `ActorLogFormatter`, which is a log formatter that consumes the logging messages, and formats them so they are pretty:

Screenshot 2023-02-15 at 10 31 01

The log formatter and handler is attached in the actor setup, so in the `__main__.py` file. We'll have to put the setup in all the Python actor templates so that it's applied in the actors by default.

The integration tests and unit tests might be too strict, and any slight change might break them, so if it happens too often, we can rewrite them a bit.

### DIFF
--- a/src/apify/log.py
+++ b/src/apify/log.py
@@ -9,7 +9,7 @@ from colorama import Fore, Style, just_fix_windows_console
 just_fix_windows_console()
 
 
-# Name of the logger used throughout the library
+# Name of the logger used throughout the library (resolves to 'apify')
 logger_name = __name__.split('.')[0]
 
 # Logger used throughout the library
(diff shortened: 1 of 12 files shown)


src/apify/log.py
```diff
@@ -0,0 +1,85 @@
...
+}
+
+# So that all the log messages have the same alignment
+_LOG_MESSAGE_INDENT = ' ' * 6
+
+
+class ActorLogFormatter(logging.Formatter):
+    """Log formatter that prints out the log message nicely formatted, with colored level and stringified extra fields."""
+
+    empty_record = logging.LogRecord('dummy', 0, 'dummy', 0, 'dummy', None, None)
```
jirimoravcik, line 41: wtf is this

jirimoravcik (approved):

src/apify/log.py
fnesveda: Ah yeah this deserves a comment, will add it 😄

drobnikj (approved):

drobnikj:
For some reason, just one test is failing. @fnesveda

PR #70, opened 2023-02-15
Fix handling the `MIGRATING` actor event

fnesveda:
The handler for the `MIGRATING` event was missing the argument for event data, so it was broken. This was very hard to discover because you can't force actor migrations easily.

This fixes it, plus adds a unit test for checking this behavior.

It also makes sure that the final `PERSIST_STATE` event is not emitted twice.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -64,6 +64,7 @@ setup(
         'aiofiles ~= 22.1.0',
         'aioshutil ~= 1.2',
         'cryptography ~= 39.0.1',
+        'colorama ~= 0.4.6',
     ],
     extras_require={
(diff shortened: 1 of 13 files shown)


jirimoravcik (approved):

drobnikj (approved):
Good catch. It is strange that on worker we did not emit persist state during migration same as we do for aborting.

fnesveda:
Yeah, maybe we could just emit it on the worker instead of in the SDK.

But then it would cause all the actors to persist their state twice until they update their SDK version, that is not ideal either.

PR #73, opened 2023-02-17
Add more Flake8 plugins, fix issues they reported

fnesveda:
This adds a few Flake8 plugins I found interesting, and it fixes the issues the plugins reported.

They are:
- `flake8-comprehensions`
	- checks that list and dict comprehensions are used everywhere they should be, to make code more readable
	- this found a few issues but nothing major
- `flake8-datetimez`
	- makes sure that we don't use timezone-naive objects anywhere
	- this found no issues
- `flake8-noqa`
	- makes sure `noqa` comments are formatted correctly
	- this found no issues
- `flake8-pytest-style`
	- enforces some best practices for Pytest
	- this found quite a few issues
		- we were mixing `@pytest.fixture` and `@pytest.fixture()`
		- our `with pytest.raises(...)` calls were too lenient, we should be checking for what the errors are more explicitly
		- the `with pytest.raises(...): ...` blocks should contain only a single statement, if you have multiple statements that each raise an error, only the first one is actually called, the others are not


fnesveda:
The unit tests failing for Python 3.11 is a nice catch actually

fnesveda:
Looks like it's caused by a breaking `Enum.__format__()` change in Python 3.11: https://docs.python.org/3/whatsnew/3.11.html#enum

jirimoravcik (approved):

drobnikj (approved):

PR #74, opened 2023-02-17
Remove `pydantic`

fnesveda:
This PR removes `pydantic` and replaces it with custom argument validation.

It had many drawbacks and it was not super useful for us:
- it broke marking methods as `async` in documentation
- it had ~2.5 MB (30% of all dependencies of the SDK)
- it was a pain to use correctly because it had bugs with validating nested annotated types
- we were only really using it to validate URLs

Now, the only huge dependency is the `cryptography` library, which has 5 MB, but that is much more valuable for us.


jirimoravcik (approved):
much nicer without pydantic

drobnikj (approved):

PR #76, opened 2023-02-22
Make sure local `RequestQueueClient` respects `forefront`

### LINKED ISSUE #66: Forefront parameter for request queue client in MemoryStorage does not work
See https://github.com/apify/crawlee/issues/1787

We currently ignore the `orderNo` property in `list_head`.

TODO:
1. Fix methods to respect the `forefront` parameter
2. Add it to unit tests

fnesveda:
This makes sure that the `RequestQueueClient` in `MemoryStorage` respects forefront, and maintains proper order of inserted requests.

This wasn't so hard in the end, I found [this neat library](https://grantjenks.com/docs/sortedcollections/valuesorteddict.html)
that takes care of keeping the requests sorted in memory.

I've also changed `orderNo` to be a `Decimal`, in milliseconds, and I've added logic that makes sure it's unique, to make the sorting work correctly.

Resolves #66.

### DIFF
--- a/tests/unit/memory_storage/resource_clients/test_request_queue.py
+++ b/tests/unit/memory_storage/resource_clients/test_request_queue.py
@@ -157,3 +157,42 @@ async def test_delete_record(request_queue_client: RequestQueueClient) -> None:
     assert rq_info_after_update['pendingRequestCount'] == 0
     # Does not crash when called again
     await request_queue_client.delete_request(request_info['requestId'])
+
+
+async def test_forefront(request_queue_client: RequestQueueClient) -> None:
+    # this should create a queue with requests in this order:
+    # Handled:
+    #     2, 5, 8
+    # Not handled:
+    #     7, 4, 1, 0, 3, 6
+    for i in range(9):
+        request_url = f'http://example.com/{i}'
+        forefront = i % 3 == 1
+        was_handled = i % 3 == 2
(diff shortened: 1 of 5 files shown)


tests/unit/memory_storage/resource_clients/test_request_queue.py
```diff
@@ -157,3 +157,42 @@ async def test_delete_record(request_queue_client: RequestQueueClient) -> None:
...
+            'url': request_url,
+            'handledAt': datetime.now(timezone.utc) if was_handled else None,
+        }, forefront=forefront)
+
+    # Check that the queue head (unhandled items) is in the right order
+    queue_head = await request_queue_client.list_head()
+    req_unique_keys = [req['uniqueKey'] for req in queue_head['items']]
+    assert req_unique_keys == ['7', '4', '1', '0', '3', '6']
+
+    # Mark request #6 as handled
```
jirimoravcik, line 183:
```suggestion
    # Mark request #1 as handled
```
?

jirimoravcik (approved):

tests/unit/memory_storage/resource_clients/test_request_queue.py
fnesveda: 🤦 good catch

drobnikj (approved):

PR #82, opened 2023-03-10
Fixes to `RequestQueue`

fnesveda:
This fixes three bugs in `RequestQueue`:
- when loading a queue from disk, `orderNo`s were not correctly converted back to `Decimal`s
- when adding requests to the queue, even handled requests were added to the queue head
- when opening an already populated queue, if you first added another request to it before you called `fetch_next_request()`, the queue head would not contain the requests that were in the queue before you opened it

It also adds an E2E test which catches all three of these things.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ Changelog
 
 - started triggering base Docker image builds when releasing a new version
 
+### Fixed
+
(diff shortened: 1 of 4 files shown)


drobnikj (approved):

jirimoravcik (approved):
nice

PR #88, opened 2023-05-03
Add support for terminal status message, allow setting status message directly in exit

mvolfik:
Depends on apify/apify-client-python#122

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ Changelog
 ### Added
 
 - option to add event handlers which accept no arguments
+- added support for `is_terminal` flag in status message update
+- option to set status message along with `Actor.exit()`
 
 ### Fixed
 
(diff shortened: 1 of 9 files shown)


mvolfik:
Hm, I'm trying to create integration tests, so I added this:

```python3
    async def test_fail_status_message(self, make_actor: ActorFactory) -> None:
        async def main() -> None:
            async with Actor:
                input = await Actor.get_input()
                await Actor.fail(**input) if input else await Actor.fail()

        actor = await make_actor('actor-fail', main_func=main)

        run_result = await actor.call(run_input={'status_message': 'This is a test message'})
        assert run_result is not None
        assert run_result['status'] == 'FAILED'
        assert run_result['statusMessage'] == 'This is a test message'
```

to `class TestActorFail`, but it's failing, and I don't know if I have the tests wrong or the code, so I'll need to check this more tomorrow before it's ready to merge.

fnesveda (changes requested):
Looks good, just one idea about the function signature

fnesveda (approved):

PR #90, opened 2023-05-12
feat: use test user token from organisation secrets

drobnikj:
I removed `APIFY_TEST_USER_API_TOKEN` from repository secrets and keep only one from organisation.


fnesveda (approved):
I had the same change ready locally 😄

PR #91, opened 2023-05-17
Save local key-value store records with the correct extension

### LINKED ISSUE #85: Using Actor.set_value('image.png', data) with memory storage saves as .bin
Using Actor.set_value('image.png', data) with memory storage saves the file as 'image.png.bin'

fnesveda:
There was a bug that files in local key-value store were saved with the wrong extension, e.g. `Actor.set_value('image.png', data)` would save the record as `image.png.bin`.

This fixes it by making the filename generation more sensible:
- the `.bin` extension won't be added anymore if the content type is `application/octet-stream`.
- the record metadata is no longer saved as `{record_key}.__metadata__.json`, but as `{record_filename}.__metadata__.json`
	- this makes it a lot easier to find the right metadata for a given file when opening the store
- the metadata no longer contains the `extension` attribute, just the key
- the internal record representation now contains the `filename` attribute pointing to the filename under which the record is saved.

I think the logic is a bit more readable this way, and it works better.

I also wrote a lot of tests to check if it works correctly, they're a bit hacky, but good enough.

Fixes #85.

### DIFF
--- a/src/apify/_memory_storage/resource_clients/key_value_store.py
+++ b/src/apify/_memory_storage/resource_clients/key_value_store.py
@@ -40,7 +40,7 @@ def _filename_from_record(record: KeyValueStoreRecord) -> str:
     if record.get('filename') is not None:
         return record['filename']
 
-    content_type = record['contentType']
+    content_type = record.get('contentType')
     if not content_type or content_type == 'application/octet-stream':
         return record['key']
     else:
--- a/src/apify/_memory_storage/resource_clients/key_value_store.py
+++ b/src/apify/_memory_storage/resource_clients/key_value_store.py
@@ -4,15 +4,18 @@ import json
 import mimetypes
 import os
 import pathlib
-import warnings
 from datetime import datetime, timezone
 from operator import itemgetter
-from typing import TYPE_CHECKING, Any, AsyncIterator, Dict, List, Optional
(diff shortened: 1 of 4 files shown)


src/apify/_memory_storage/resource_clients/key_value_store.py
```diff
@@ -24,13 +27,28 @@
...
 
-DEFAULT_LOCAL_FILE_EXTENSION = 'bin'
+KeyValueStoreRecord = TypedDict('KeyValueStoreRecord', {'key': str, 'value': Any, 'contentType': Optional[str], 'filename': NotRequired[str]})
+
+
+def _filename_from_record(record: KeyValueStoreRecord) -> str:
+    if record.get('filename') is not None:
+        return record['filename']
+
+    content_type = record['contentType']
```
drobnikj, line 43:
```suggestion
    content_type = record.get('contentType')
```
not sure but, contentType can be undefined right?

drobnikj (approved):
Just one thing otherwise looks good 👍

src/apify/_memory_storage/resource_clients/key_value_store.py
fnesveda: I think it should always be defined (but its value can possibly be `None`), but I guess better safe than sorry, I will update it

jirimoravcik (approved):
test parametrization looks scary, but I don't see any nicer way to do it

PR #100, opened 2023-07-25
Start using apify-shared for general consts and utils

### LINKED ISSUE #98: Start using apify-shared for general consts and utils
- In https://github.com/apify/apify-shared-python/issues/1 we introduced `apify-shared` library for sharing general constants and utils across our Python projects. Let's start using it.
  - https://github.com/apify/apify-shared-python
  - https://pypi.org/project/apify-shared/

vdusek:
Closes #98


jirimoravcik (approved):

jirimoravcik:
There's still some errors from type checks, etc.

I'd suggest releasing new client first and then using it in this PR so you don't have to write type ignores...

vdusek:
> There's still some errors from type checks, etc.
>
> I'd suggest releasing new client first and then using it in this PR so you don't have to write type ignores...

Agreed, thanks.

src/apify/consts.py
```diff
@@ -1,208 +1,5 @@
 import re
 from enum import Enum
```
fnesveda, line 2: We'll have to do the reexport magic here, same as in client, so it's not a breaking change

fnesveda (commented):
Sorry about the late review, I was waiting for the tests to be fixed.

Anyway, just one note - we need to reexport the consts in `apify.consts` so that this is not a breaking change.

And please bump the version and add an entry to `CHANGELOG.md`

PR #103, opened 2023-08-01
chore: Code quality improvements

jirimoravcik:
I used https://pypi.org/project/flake8_simplify/ to get some insights and improved the code a bit

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -53,6 +53,7 @@ dev = [
     "flake8-noqa ~= 1.3.1",
     "flake8-pytest-style ~= 1.7.2",
     "flake8-quotes ~= 3.3.2",
+    "flake8-simplify ~= 0.20.0",
     "flake8-unused-arguments ~= 0.0.13",
     "isort ~= 5.12.0",
     "mypy ~= 1.3.0",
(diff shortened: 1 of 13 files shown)


vdusek (approved):
LGTM

fnesveda (approved):
Cool!

vdusek (changes requested):
What about adding the `flake8_simplify` package to the [dev dependencies](https://github.com/apify/apify-sdk-python/blob/master/pyproject.toml#L40)?

jirimoravcik:
> What about adding the `flake8_simplify` package to the [dev dependencies](https://github.com/apify/apify-sdk-python/blob/master/pyproject.toml#L40)?

Yeah, I though about it.  It reports some stuff that I didn't want to change :D (specifically, using context managers for files - we have code that does `open(..).close()` which is shorter than using a context manager. But I think it's fine

jirimoravcik:
Added `flake8-simplify` as dev dependency. Will also do for other 2 repos along with code simplification

jirimoravcik:
> > Yeah, I though about it. It reports some stuff that I didn't want to change :D (specifically, using context managers for files - we have code that does open(..).close() which is shorter than using a context manager.
> 
> I'd use `#noqa` for such cases, up to you, both variants LGTM.

I just did the suggested change, it's better than noqa or adding the rule to flake ignores

PR #106, opened 2023-08-02
chore: Pin apify deps versions

jirimoravcik:
We pin Apify deps versions to prevent newer packages breaking older SDK versions


vdusek (approved):
LGTM

PR #127, opened 2023-10-24
chore: Check whether version in PR is already released

fnesveda:
We have a recurring issue that you often forget to bump the package version number in a PR, and then when you merge it to `master`, the package publication fails because the version you're trying to publish already exists.

This adds a check to CI which detects these issues early and fails the CI, so you can't merge those PRs without bumping the package version number.

It gets skipped if the title of the PR is `docs: ...`, because such PRs don't trigger package publication when merged.

### DIFF
--- a/.github/workflows/check_version_availability.yaml
+++ b/.github/workflows/check_version_availability.yaml
@@ -1,0 +1,25 @@
+name: Check package version availability
+
+on:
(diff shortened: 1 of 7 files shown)


fnesveda:
Yup, once I merge it here and see that it works, I will copy it there too.

PR #133, opened 2023-11-11
Migrate from Flake8 & Autopep8 to Ruff

### LINKED ISSUE #99: Remove underscore prefix from objects that are not private
Currently, we use the underscore prefix in objects that are imported from other modules (e.g. all objects in https://github.com/apify/apify-client-python/blob/master/src/apify_client/_utils.py) - are not private.

This was intended to let users of the library know, that these objects are for internal usage only.

However, this does not correspond to the usage of underscore prefixes in the Python world. We should remove these prefixes from the non-private objects.

We can still use the underscore prefix in the module names, to let users know, this module is only for internal usage and should not be imported by library users.

### LINKED ISSUE #128: Switch to Ruff (linter & formatter)
Ruff is a new extremely fast Python linter written in Rust, which supports many rules from the `flake8` & `pylint` world.

They recently released a formatter, where single quotes are an option :).

### Task list

- [ ] Install the `ruff`, and update the configs & code to be aligned with the linter rules.
- [ ] Update the GH workflows.
- [ ] Do the same for the `apify-client-python` and `apify-shared-python` libraries.

vdusek:
### Description

- Most changes were performed automatically using `ruff format` and `ruff check --fix`.
- Type hints were updated manually.
- To not break anything I tried not to update/fix violations and rather turn off the violation for a particular line (with some exceptions).
- I've also created many new issues addressing either the TODOs in the code or fixing new linter rule violations.
  - https://github.com/apify/apify-sdk-python/issues/137
  - https://github.com/apify/apify-sdk-python/issues/138
  - https://github.com/apify/apify-sdk-python/issues/139
  - https://github.com/apify/apify-sdk-python/issues/140
  - https://github.com/apify/apify-sdk-python/issues/141
  - https://github.com/apify/apify-sdk-python/issues/142
  - https://github.com/apify/apify-sdk-python/issues/143
  - https://github.com/apify/apify-sdk-python/issues/146
  - https://github.com/apify/apify-sdk-python/issues/147
  - https://github.com/apify/apify-sdk-python/issues/148
  - https://github.com/apify/apify-sdk-python/issues/149
  - https://github.com/apify/apify-sdk-python/issues/150

### Issues

- Closes #128 
- Closes #99 

### Problem with `typing.get_type_hints()`

- There is a problem on Python 3.8 and 3.9, `typing.get_type_hints()` throws an error if `|` is used in the type hint, even if `from __future__ import annotations` is used.
- Fortunately, this is not part of our code base and the users of SDK should not have a valid use case for using it for building Actors. However, we use it in the unit tests when mocking methods of [ApifyClientAsync](https://github.com/apify/apify-client-python/blob/master/src/apify_client/client.py#L278) in here [tests/unit/conftest.py#L81](https://github.com/apify/apify-sdk-python/blob/v1.3.0/tests/unit/conftest.py#L81).
- I implemented a workaround, to make it working, check the commit https://github.com/apify/apify-sdk-python/pull/133/commits/e81f8ab889880f1e71f226d4076f9e16922e248f.


src/apify/_memory_storage/resource_clients/dataset.py
```diff
@@ -94,36 +97,36 @@ async def update(self, *, name: Optional[str] = None) -> Dict:
...
             previous_dir = existing_dataset_by_id._resource_directory
 
             existing_dataset_by_id._resource_directory = os.path.join(self._memory_storage_client._datasets_directory, name)
 
-            await _force_rename(previous_dir, existing_dataset_by_id._resource_directory)
+            await force_rename(previous_dir, existing_dataset_by_id._resource_directory)
 
             # Update timestamps
-            await existing_dataset_by_id._update_timestamps(True)
+            await existing_dataset_by_id._update_timestamps(has_been_modified=True)  # type: ignore
```
jirimoravcik, line 125: why is the type ignore needed now?

src/apify/_memory_storage/resource_clients/dataset.py
```diff
@@ -188,47 +191,49 @@ async def list_items(
...
 
             items = []
 
             for idx in range(start, end):
                 entry_number = self._generate_local_entry_name(idx)
-                items.append(existing_dataset_by_id._dataset_entries[entry_number])
+                items.append(existing_dataset_by_id._dataset_entries[entry_number])  # type: ignore
 
-            await existing_dataset_by_id._update_timestamps(False)
+            await existing_dataset_by_id._update_timestamps(has_been_modified=False)  # type: ignore
```
jirimoravcik, line 209: so many type ignores here, why?

src/apify/_memory_storage/resource_clients/key_value_store.py
```diff
@@ -196,7 +213,7 @@ async def list_keys(self, *, limit: int = DEFAULT_API_PARAM_LIMIT, exclusive_sta
         if exclusive_start_key is not None:
             key_pos = next((idx for idx, i in enumerate(items) if i['key'] == exclusive_start_key), None)
             if key_pos is not None:
-                truncated_items = items[key_pos + 1:]
+                truncated_items = items[key_pos + 1 :]
```
jirimoravcik, line 216:
```suggestion
                truncated_items = items[(key_pos + 1):]
```
?

src/apify/_utils.py
```diff
@@ -308,67 +327,68 @@ class LRUCache(MutableMapping, Generic[T]):
...
     _max_length: int
 
-    def __init__(self, max_length: int) -> None:
+    def __init__(self: LRUCache, max_length: int) -> None:
         """Create a LRUCache with a specific max_length."""
         self._cache = OrderedDict()
         self._max_length = max_length
 
-    def __getitem__(self, key: str) -> T:
+    def __getitem__(self: LRUCache, key: str) -> Any:
```
jirimoravcik, line 335: This seems wrong to me. Should return `T`

src/apify/_crypto.py
```diff
@@ -89,35 +91,37 @@ def private_decrypt(
 
     # Slice the encrypted into cypher and authentication tag
     authentication_tag_bytes = encrypted_value_bytes[-ENCRYPTION_AUTH_TAG_LENGTH:]
-    encrypted_data_bytes = encrypted_value_bytes[:len(encrypted_value_bytes) - ENCRYPTION_AUTH_TAG_LENGTH]
+    encrypted_data_bytes = encrypted_value_bytes[: len(encrypted_value_bytes) - ENCRYPTION_AUTH_TAG_LENGTH]
```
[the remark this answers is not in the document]
vdusek: Sure, no problem, `ruff format` will just add additional space after the `:` there.

src/apify/_memory_storage/resource_clients/key_value_store.py
vdusek: Sure, just again, `ruff format` will add a space there (`truncated_items = items[(key_pos + 1) :]`)

jirimoravcik:
pls replace `methonds` with `methods`

src/apify/_utils.py
vdusek: Yeah, I can do that. However, then I'll either have to put there `# type: ignore` or remove `warn_return_any`. Because it says:
> Returning `Any` from function declared to return `"T"` Mypy [no-any-return](https://mypy.readthedocs.io/en/latest/_refs.html#code-no-any-return)

jirimoravcik: how about a `cast`?

src/apify/_memory_storage/resource_clients/dataset.py
vdusek: The problem with all of the new `# type: ignore` in `resource_clients/` is the following...

The `_find_or_create_client_by_id_or_name` method returns either `BaseResourceClient` or `None`.

image

There is a check for a `None` value, so after that, the IDE expects it to be `BaseResourceClient`.

However, `BaseResourceClient` does not have all of the fields that are accessed here (`_file_operation_lock`, `_update_timestamps`, ...). Only its child/children have it. 

image

So that's the problem and probably always has been. The reason why it was not caught before, might be the update of Mypy from 1.5 to 1.7?

src/apify/_memory_storage/resource_clients/dataset.py
vdusek: Please continue with the discussion here https://github.com/apify/apify-sdk-python/pull/133#discussion_r1414245709.

src/apify/_utils.py
vdusek: Yeah, that sounds like a good way to go...

jirimoravcik (approved):

src/apify/_memory_storage/resource_clients/dataset.py
vdusek: Ok, I created a new issue for that - https://github.com/apify/apify-sdk-python/issues/152, resolving.

PR #154, opened 2023-12-05
Add comment why not to use `~=` clause in deps definition

vdusek:
Currently, we use the [inclusive ordered comparison clause](https://peps.python.org/pep-0440/#version-specifiers) in our dependency requirements (except the apify packages). 

I suggest playing safely by using a more strict [compatible release clause](https://peps.python.org/pep-0440/#version-specifiers) as we use it for the dev dependencies or dependencies in the Apify Client.

I install the dependencies using `make install-dev`, then use `pip freeze` and copy the package versions it installed - the current versions we use.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,7 @@
 
 ## [1.4.1](../../releases/tag/v1.4.1) - Unreleased
 
-### Internal changes
-
-- Use compatible release clause in dependencies definition
+...
 
 ## [1.4.0](../../releases/tag/v1.4.0) - 2023-12-05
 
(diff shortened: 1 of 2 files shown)


vdusek:
@fnesveda Oh, thank you, that makes sense...

I'll close the PR without merging.

What do you think about relaxing the dependencies for the `apify-client` as well ([here](https://github.com/apify/apify-client-python/blob/master/pyproject.toml#L24:L27))? Although, there is only HTTPX.

fnesveda (approved):
Cool 👍 I would probably use the ordered clause in the client as well (and for the scrapy extra here too).

Issue #64, opened 2023-02-13
Investigate usage of DataClasses, TypedDicts for e.g. requests in RequestQueue


jirimoravcik:
We should check possible ways how to pass data to storages. The current dicts are not very user-friendly and provide no type-safety or hints. One interesting option is a dataclass, `asdict` https://docs.python.org/3/library/dataclasses.html#dataclasses.asdict can be used to convert data to a dict that can be sent using clients.

Another interesting possibility is a TypedDict, it can work nicely for return types that have fixed members.

jirimoravcik:
Possibly check https://www.attrs.org/en/stable/index.html

fnesveda:
I was playing with this a bit now, because I'm not a fan of how the API client returns generic dictionaries without any types, and because accessing deeply nested values in responses is quite annoying, and I managed to make a custom dataclass which accepts extra argumens (so that if we ever add fields to the API it won't break the dataclass).

Not sure if this is the direction we would want to go with, but it could be a possibility for client v2 (perhaps along with API v3?).

```python
from dataclasses import dataclass, fields
from typing import Dict, List, Optional, get_origin

def maybe_convert_to_generic_api_object(val):
    if isinstance(val, dict):
        return GenericApiObject(**val)
    if isinstance(val, list):
        return [maybe_convert_to_generic_api_object(item) for item in val]
    return val


@dataclass(init=False, repr=False)
class GenericApiObject:
    def __init_subclass__(cls) -> None:
        dataclass(cls, init=False, repr=False)

    def __init__(self, **kwargs: Dict):
        fields_dict = {field.name: field for field in fields(self)}

        for key, value in kwargs.items():
            if key in fields_dict:
                field = fields_dict[key]
                if isinstance(field.type, type) and issubclass(field.type, GenericApiObject):
                    value = field.type(**value)
                elif get_origin(field.type) is list:
                    list_type = field.type.__args__[0]
                    if isinstance(list_type, type) and issubclass(list_type, GenericApiObject):
                        value = [list_type(**item) for item in value]
                elif get_origin(field.type) is dict:
                    dict_type = field.type.__args__[1]
                    if isinstance(dict_type, type) and issubclass(dict_type, GenericApiObject):
                        value = { k: dict_type(**v) for k, v in value.items() }
            else:
                value = maybe_convert_to_generic_api_object(value)

            setattr(self, key, value)

    def __repr__(self):
[...]

PR #158, opened 2023-12-20
Add Scrapy `ApifyHttpProxyMiddleware` for managing proxies

vdusek:
### Issue

- Closes: https://github.com/apify/actor-templates/issues/255

### Other

- It is used in the https://github.com/apify/actor-templates/pull/260. I released an alpha version of the SDK, so we can test it in the template Actor.


honzajavorek:
As soon as it's released, I'm happy to try it out and report any feedback.

src/apify/scrapy/middlewares/apify_retry.py
```diff
@@ -11,14 +11,14 @@
         'To use this module, you need to install the "scrapy" extra. Run "pip install apify[scrapy]".',
     ) from exc
 
-from ..actor import Actor
-from .utils import nested_event_loop, open_queue_with_custom_client, to_apify_request
+from ...actor import Actor
```
janbuchar, line 14: Could we replace this with an absolute import? It is difficult for me to see any advantage to relatively importing something this far away.

src/apify/scrapy/middlewares/apify_proxy.py
```diff
@@ -0,0 +1,156 @@
...
+if TYPE_CHECKING:
+    from scrapy import Request, Spider
+    from scrapy.crawler import Crawler
+
+
+class ApifyHttpProxyMiddleware:
+    """Apify HTTP proxy middleware for Scrapy.
+
+    This middleware enhances request processing by adding a 'proxy' field to the request's meta and an
+    authentication header. It draws inspiration from the Scrapy default `HttpProxyMiddleware`. The proxy URL is
```
janbuchar, line 22: Sorry to be nitpicky, but in what sense is that Scrapy middleware default?

src/apify/scrapy/middlewares/apify_proxy.py
```diff
@@ -0,0 +1,156 @@
...
+    sourced from the settings under the `APIFY_PROXY_SETTINGS` key. The value of this key, a dictionary, should be
+    provided by the Actor input. An example of the proxy settings:
+
+    proxy_settings = {'useApifyProxy': true, 'apifyProxyGroups': []}
+    """
+
+    def __init__(
+        self: ApifyHttpProxyMiddleware,
+        proxy_settings: dict,
```
janbuchar, line 31: Could we use a `TypedDict` here? A plain `dict` is difficult to reason about - is it a mapping? Probably not, what keys does it accept?

src/apify/scrapy/middlewares/apify_proxy.py
```diff
@@ -0,0 +1,156 @@
...
+
+        # Get a new proxy URL and return it
+        new_url = await proxy_cfg.new_url()
+        return urlparse(new_url)
+
+    def _get_basic_auth_header(self: ApifyHttpProxyMiddleware, username: str, password: str) -> bytes:
+        """Generate a basic authentication header for the given username and password."""
+        string = f'{unquote(username)}:{unquote(password)}'
+        user_pass = to_bytes(string, encoding=self._auth_encoding)
+        return b'Basic ' + b64encode(user_pass)
```
janbuchar, line 156: I believe this could be a "global" helper function.

src/apify/scrapy/middlewares/apify_proxy.py
```diff
@@ -0,0 +1,156 @@
...
+    sourced from the settings under the `APIFY_PROXY_SETTINGS` key. The value of this key, a dictionary, should be
+    provided by the Actor input. An example of the proxy settings:
+
+    proxy_settings = {'useApifyProxy': true, 'apifyProxyGroups': []}
+    """
+
+    def __init__(
+        self: ApifyHttpProxyMiddleware,
+        proxy_settings: dict,
+        auth_encoding: str = 'latin-1',
```
janbuchar, line 32: Is there a realistic usecase for overriding this?

src/apify/scrapy/middlewares/apify_retry.py
vdusek: I understand... However, the relative imports are used everywhere in the repository. 

We can ask @jirimoravcik - is there any reason you guys preferred using relative rather than absolute imports? Thanks. 

I'd stick with that for now because of the consistency. And if there is no particular reason for that, we can replace them later in all the places. I've got no problem with that.

src/apify/scrapy/middlewares/apify_proxy.py
vdusek: Scrapy [HttpProxyMiddleware](https://github.com/scrapy/scrapy/blob/master/scrapy/downloadermiddlewares/httpproxy.py#L25C66-L25C66) handles the `auth_encoding` in the same way, so I stick with that.

Would you rather have it as a module/class level constant?

src/apify/scrapy/middlewares/apify_proxy.py
vdusek: Sure, it's moved to the `scrapy.utils`.

src/apify/scrapy/middlewares/apify_proxy.py
vdusek: I moved the `get_basic_auth_header` to a separate module, so the optional argument is moved just to that function.

src/apify/scrapy/middlewares/apify_proxy.py
janbuchar: Oh, I see, thanks. Then I'd probably reword this along the lines of "It draws inspiration from the `HttpProxyMiddleware` included by default in Scrapy projects". What do you think?

src/apify/scrapy/middlewares/apify_retry.py
janbuchar: I guess I already pointed this out before, huh? I agree that if we decide to replace this, we should do it in one go.

src/apify/scrapy/middlewares/apify_proxy.py
janbuchar: And if you decide to use that argument, can you comfortably set the header value in the ApifyHttpProxyMiddleware class?

By the way, `httpx` for example handles this by accepting both `str` and `bytes` for `username` and `password`, and if you want to control the encoding, you just pass pre-encoded `bytes`

src/apify/scrapy/middlewares/apify_proxy.py
vdusek: That would be nice. However, it's just an argument that is passed to the [Actor.create_proxy_configuration](https://github.com/apify/apify-sdk-python/blob/master/src/apify/actor.py#L1282) (and then it is passed further),  where it is just a general dictionary as well. We do not do the processing of the dict here.

If we would want a more specific typing here we should create a typedict (or dataclass) for it and use it across all its occurrences.

janbuchar: Or even some kind of a model class that does validation. I understand if it's out of the scope of this PR, and I also understand that you don't usually instantiate this class manually. But still, it would make the code more readable.

src/apify/scrapy/middlewares/apify_retry.py
vdusek: ok, i opened an issue for that https://github.com/apify/apify-sdk-python/issues/160

src/apify/scrapy/middlewares/apify_proxy.py
vdusek: Sure, done

src/apify/scrapy/middlewares/apify_proxy.py
vdusek: Discussed on Slack... We're gonna leave it as it is for now, if someone in the future needs to set up the encoding, we can provide it by extracting optional value from the `crawler.settings` in the `from_crawler` method and pass it to the default constructor.

src/apify/scrapy/middlewares/apify_proxy.py
jirimoravcik: See https://github.com/apify/apify-sdk-python/issues/64

src/apify/scrapy/middlewares/apify_retry.py
jirimoravcik: I don't remember if there was a particular reason for relative imports. I think it's fine to replace them with absolute imports. But please do it everywhere so we keep the codebase consistent.

janbuchar (approved):

PR #161, opened 2024-01-03
Use absolute imports

honzajavorek:
Implements https://github.com/apify/apify-sdk-python/issues/160, with exception of tests, where the relative imports make some sense and removing them would mean re-architecturing the tests.

I've ran `make check-code` and it did complain about `Import block is un-sorted or un-formatted` after my changes, so I've ran it with `--fix` and committed the changes, but then I've hit the following error caused by circular import:

```
Lint codebase.................................................................Passed
Type-check codebase...........................................................Passed
Run unit tests................................................................Failed
- hook id: unit-tests
- exit code: 2

python3 -m pytest -n auto -ra tests/unit --cov=src/apify
ImportError while loading conftest '/Users/honza/Projects/apify-sdk-python/tests/unit/conftest.py'.
tests/unit/conftest.py:11: in <module>
    from apify import Actor
src/apify/__init__.py:3: in <module>
    from apify.actor import Actor
src/apify/actor.py:27: in <module>
    from apify.storages import Dataset, KeyValueStore, RequestQueue, StorageClientManager
src/apify/storages/__init__.py:1: in <module>
    from apify.storages.dataset import Dataset
src/apify/storages/dataset.py:11: in <module>
    from apify.storages.key_value_store import KeyValueStore
src/apify/storages/key_value_store.py:7: in <module>
    from apify_client.clients import KeyValueStoreClientAsync, KeyValueStoreCollectionClientAsync
venv/lib/python3.11/site-packages/apify_client/__init__.py:3: in <module>
    from .client import ApifyClient, ApifyClientAsync
venv/lib/python3.11/site-packages/apify_client/client.py:6: in <module>
    from .clients import (
venv/lib/python3.11/site-packages/apify_client/clients/__init__.py:1: in <module>
    from .base import (
venv/lib/python3.11/site-packages/apify_client/clients/base/__init__.py:1: in <module>
    from .actor_job_base_client import ActorJobBaseClient, ActorJobBaseClientAsync
venv/lib/python3.11/site-packages/apify_client/clients/base/actor_job_base_client.py:11: in <module>
    from apify._errors import ApifyApiError
E   ModuleNotFoundError: No module named 'apify._errors'
```

I think the circular import wasn't there, but sorting the imports as required by `ruff` probably introduced it. I did not investigate further, as this was supposed to be just a small gift, not a all evening debugging session 😅

I intentionally committed both changes separately, so that you can cherry pick what you like and drop the rest.

PR #167, opened 2024-01-08
Add option to run integration tests from PRs from forks

fnesveda:
With our [first community PR](https://github.com/apify/apify-sdk-python/pull/161) (🎉 ), there was an issue that the integration tests for the PR did not run since they could not access the API token for the test user from the repo secrets.

This changes the check workflows so that the tests run on the `pull_request_target` event, not on the `pull_request` event, which allows passing the secrets to the workflow even in case of forks. To prevent leaking the secrets, running the workflows on a PR from a fork requires approving the workflow run by a maintainer via an environment protection rule. I've put @vdusek, @janbuchar, @jirimoravcik and me as maintainers of that environment, so that we get notified in case of a fork PR and can approve the workflow runs.

The `pull_request_target` event runs in the context of the PR base branch, not the head branch, for security reasons, so to run the tests against the PR code, we have to check it out explicitly instead of checking out the default ref.

### DIFF
--- a/.github/workflows/check_version_availability.yaml
+++ b/.github/workflows/check_version_availability.yaml
@@ -12,6 +12,8 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
+        with:
+          ref: "${{ github.event_name == 'pull_request_target' && (github.event.pull_request.merge_commit_sha || github.event.pull_request.head.sha) || '' }}"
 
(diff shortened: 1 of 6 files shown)


jirimoravcik (approved):

janbuchar (approved):

vdusek (approved):
Good job, thanks :+1:

Issue #176, opened 2024-01-16
Move most of the Scrapy template's logic to Apify SDK


honzajavorek:
I think some logic from [\_\_main\_\_.py](https://github.com/apify/actor-templates/blob/dc5e68805dcf630f35d112a7e113e4f388bbf30a/templates/python-scrapy/src/__main__.py) could be moved to the SDK. I think `new_configure_logging` could be a decorator which is just imported from Apify SDK. I think `configure_logger` and logger names could be imported.

Similarly, [main.py](https://github.com/apify/actor-templates/blob/master/templates/python-scrapy/src/main.py) contains `_get_scrapy_settings`, but it could also be just imported, if turned into this:

```python
def apply_apify_settings(settings: Settings, proxy_config: dict | None = None):
    ...
    return settings
```

Then it wouldn't need to call `get_project_settings()` and would leave space for custom modifications before or after applying Apify-specific settings.

The main benefit of doing this is that the template contains less boilerplate and it's easier to control and maintain updates. If new Scrapy version is published and the logic of the monkey patching or anything else needs to be changed, I could just upgrade Apify SDK with updates. As of now the only way is to watch updates to the template manually and update my code by copy-pasting.

Scrapy is a library and I can upgrade it carefully or pin it to a certain version, but Apify is a SaaS platform. If something changes in Apify and it won't be compatible with the old template code, it can just break my actors out of nowhere. Only then I will be prompted to go and see if the template looks differently than last time. Not ideal.

Follow up to https://github.com/apify/apify-sdk-python/issues/132, vaguely related to https://github.com/apify/actor-templates/issues/264

honzajavorek:
I did some changes in my implementation so that it's more tidied up:

- Introduced CLI where only the logging setup is hoisted before everything else: https://github.com/juniorguru/plucker/blob/6fe0c31097b00339cbc05b2ab40fc1dae23160bd/juniorguru_plucker/cli.py
- Moved all logging setup to a separate file: https://github.com/juniorguru/plucker/blob/main/juniorguru_plucker/loggers.py
- I moved monkey-patching to a decorator - not sure myself if this is nicer or not, but won't go and rewrite it again now: https://github.com/juniorguru/plucker/blob/6fe0c31097b00339cbc05b2ab40fc1dae23160bd/juniorguru_plucker/loggers.py#L38
[...]

Issue #148, opened 2023-12-02
Add integration test for Scrapy integration


vdusek:
https://github.com/apify/apify-sdk-python/tree/v1.3.0/src/apify/scrapy

PR #178, opened 2024-01-22
Add `apply_apify_settings` to Scrapy subpackage

vdusek:
### Issue

- Related to the https://github.com/apify/apify-sdk-python/issues/176, but not solving it as a whole.

### Description

- This PR moves just a single function from a template to SDK - `apply_apify_settings`.
- The rest of the suggested changes are related to the logging. I was not able to make them work at first try. So for now let's solve it just for the settings-related helper function, and the rest will be done later.

### Testing

- The changes were tested both locally and on the platform. 
- Unit tests are ready as well.

### DIFF
--- a/tests/unit/scrapy/utils/test_apply_apify_settings.py
+++ b/tests/unit/scrapy/utils/test_apply_apify_settings.py
@@ -1,107 +1,62 @@
 from __future__ import annotations
 
-from dataclasses import dataclass
-
-import pytest
 from scrapy.settings import Settings
 
 from apify.scrapy.utils import apply_apify_settings
 
 
-@dataclass(frozen=True)
-class TestCase:
-    settings: Settings
-    proxy_config: dict | None
-    expected_output: Settings
+def test__apply_apify_settings__overrides_scheduler() -> None:
+    settings = Settings()
+    new_settings = apply_apify_settings(settings=settings)
+
+    assert new_settings.get('SCHEDULER') == 'apify.scrapy.scheduler.ApifyScheduler'
 
 
-test_cases = [
-    # Test case with default settings and no proxy configuration
-    TestCase(
-        settings=Settings(),
-        proxy_config=None,
-        expected_output=Settings(
-            {
-                'SCHEDULER': 'apify.scrapy.scheduler.ApifyScheduler',
-                'ITEM_PIPELINES': {
-                    'apify.scrapy.pipelines.ActorDatasetPushPipeline': 1000,
-                },
-                'DOWNLOADER_MIDDLEWARES': {
-                    'scrapy.downloadermiddlewares.robotstxt.RobotsTxtMiddleware': None,
-                    'scrapy.downloadermiddlewares.httpproxy.HttpProxyMiddleware': None,
-                    'scrapy.downloadermiddlewares.retry.RetryMiddleware': None,
-                    'apify.scrapy.middlewares.ApifyHttpProxyMiddleware': 950,
-                    'apify.scrapy.middlewares.ApifyRetryMiddleware': 1000,
-                },
-                'APIFY_PROXY_SETTINGS': None,
+def test__apply_apify_settings__update_item_pipelines() -> None:
+    settings = Settings(
+        {
+            'ITEM_PIPELINES': {
+                'scrapy.pipelines.files.FilesPipeline': 1,
             }
-        ),
-    ),
-    # Test case with default settings and proxy configuration
-    TestCase(
-        settings=Settings(),
-        proxy_config={'useApifyProxy': True, 'apifyProxyGroups': []},
-        expected_output=Settings(
-            {
-                'SCHEDULER': 'apify.scrapy.scheduler.ApifyScheduler',
-                'ITEM_PIPELINES': {
-                    'apify.scrapy.pipelines.ActorDatasetPushPipeline': 1000,
-                },
-                'DOWNLOADER_MIDDLEWARES': {
-                    'scrapy.downloadermiddlewares.robotstxt.RobotsTxtMiddleware': None,
-                    'scrapy.downloadermiddlewares.httpproxy.HttpProxyMiddleware': None,
-                    'scrapy.downloadermiddlewares.retry.RetryMiddleware': None,
-                    'apify.scrapy.middlewares.ApifyHttpProxyMiddleware': 950,
-                    'apify.scrapy.middlewares.ApifyRetryMiddleware': 1000,
-                },
-                'APIFY_PROXY_SETTINGS': {'useApifyProxy': True, 'apifyProxyGroups': []},
-            }
-        ),
-    ),
-    # Test case with custom settings and proxy configuration
-    TestCase(
-        settings=Settings(
-            {
-                'SCHEDULER': 'scrapy.core.scheduler.Scheduler',
-                'ITEM_PIPELINES': {
-                    'scrapy.pipelines.files.FilesPipeline': 1,
-                },
-                'DOWNLOADER_MIDDLEWARES': {
-                    'scrapy.downloadermiddlewares.robotstxt.RobotsTxtMiddleware': 123,
-                    'scrapy.downloadermiddlewares.httpproxy.HttpProxyMiddleware': 234,
-                    'scrapy.downloadermiddlewares.retry.RetryMiddleware': 345,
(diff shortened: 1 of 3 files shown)


vdusek:
@honzajavorek FYI; also if you would have any feedback on this, I'll be glad. Thanks :slightly_smiling_face:.

tests/unit/scrapy/utils/test_apply_apify_settings.py
```diff
@@ -0,0 +1,109 @@
...
+        ),
+    ),
+]
+
+
+@pytest.mark.parametrize('tc', test_cases)
+@pytest.mark.only()
+def test_apply_apify_settings(tc: TestCase) -> None:
+    output_settings = apply_apify_settings(settings=tc.settings, proxy_config=tc.proxy_config)
+    print(dict(output_settings))
```
honzajavorek, line 104:
```suggestion
```
Leftover?

tests/unit/scrapy/utils/test_apply_apify_settings.py
```diff
@@ -0,0 +1,109 @@
...
+                },
+                'APIFY_PROXY_SETTINGS': {'useApifyProxy': True, 'apifyProxyGroups': []},
+            }
+        ),
+    ),
+]
+
+
+@pytest.mark.parametrize('tc', test_cases)
+@pytest.mark.only()
```
honzajavorek, line 101:
```suggestion
```
Leftover?

tests/unit/scrapy/utils/test_apply_apify_settings.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+@dataclass(frozen=True)
+class TestCase:
+    settings: Settings
+    proxy_config: dict | None
+    expected_output: Settings
+
+
+test_cases = [
```
honzajavorek, line 18: Too DRY for my liking. As someone reading (not writing) the tests I'd probably prefer to have three separate `test_` functions with descriptive names, testing only the values relevant for that particular test case. In my opinion here the use of `parametrize` obfuscates the aim of each test. The fact that we needed a dataclass and a whole standalone list for the test cases IMHO signals that we're pushing something too hard here.

honzajavorek (changes requested):
The implementation looks good, awesome! I've found a few places in the test where I commented, but they're nothing serious except for the `@pytest.mark.only()` leftover, which was the decisive factor in why I'm requesting changes.

tests/unit/scrapy/utils/test_apply_apify_settings.py
vdusek: ouuuu, thanks

tests/unit/scrapy/utils/test_apply_apify_settings.py
vdusek: yeah... thanks

tests/unit/scrapy/utils/test_apply_apify_settings.py
vdusek: Do you think really this approach is so bad?

IMO it seems to me both approaches have pros and cons.

What you're suggesting will lead to a much longer test file and repeating code (also quite long function names, hah).

My approach doesn't have these negatives, however, could be less readable at first glance. I agree with that.

honzajavorek: I don't think it's _so_ bad and it wouldn't be a reason to reject the PR (if I was in charge). I think DRY doesn't suit situations where you write the code once and read it 50 times, which tests usually are. I just don't see a big benefit of not repeating code here, and it comes with the readability penalty, which IMHO, in tests, has a larger impact. Function names could be:

- `test_apply_apify_settings_no_proxy_config`
- `test_apply_apify_settings_proxy_config`
- `test_apply_apify_settings_custom_settings_and_proxy_config`

Doesn't look so terrible to me. But of course, YMMV.

tests/unit/scrapy/utils/test_apply_apify_settings.py
fnesveda: I made https://github.com/fnesveda/flake8-no-pytest-mark-only exactly for this reason but then we migrated to ruff 😅

tests/unit/scrapy/utils/test_apply_apify_settings.py
honzajavorek: I tried to refactor it as an exercise and ended up with something like this:

```python
def test_apply_apify_settings_takes_default_settings_from_scrapy():
    settings = apply_apify_settings()

    assert settings['DOWNLOADER_MIDDLEWARES'].get('apify.scrapy.middlewares.ApifyRetryMiddleware') == 1000


def test_apply_apify_settings_overrides_scheduler():
    settings = apply_apify_settings(settings=Settings())

    assert settings['SCHEDULER'] == 'apify.scrapy.scheduler.ApifyScheduler'


def test_apply_apify_settings_adds_dataset_pipeline():
    settings = apply_apify_settings(settings=Settings())

    assert settings['ITEM_PIPELINES'] == {'apify.scrapy.pipelines.ActorDatasetPushPipeline': 1000}


def test_apply_apify_settings_adds_proxy_middleware():
    settings = apply_apify_settings(settings=Settings())

    assert settings['DOWNLOADER_MIDDLEWARES'].get('apify.scrapy.middlewares.ApifyHttpProxyMiddleware') == 950


def test_apply_apify_settings_adds_retry_middleware():
    settings = apply_apify_settings(settings=Settings())

    assert settings['DOWNLOADER_MIDDLEWARES'].get('apify.scrapy.middlewares.ApifyRetryMiddleware') == 1000


@pytest.mark.parametrize(
    'proxy_config, expected',
    [
        (None, None),
        ({'useApifyProxy': True, 'apifyProxyGroups': []}, {'useApifyProxy': True, 'apifyProxyGroups': []}),
    ]
)
def test_apply_apify_settings_sets_proxy(proxy_config, expected):
    settings = apply_apify_settings(settings=Settings(), proxy_config=proxy_config)

    assert settings['APIFY_PROXY_SETTINGS'] == expected
```

It's written from the top of my head, I didn't execute the code, so it can contain mistakes. I just wanted to show the approach I had in mind. I iterated towards having a separate test case for each of the changes the function does. The tests are easy to read and simply "explain" each unit tested.

I can imagine one larger "integration" test which takes whole settings and checks what has been applied, if that's necessary.

I omitted the "custom" settings completely, because I realized what we're doing here is putting custom settings to Scrapy's `Settings()`, but that only tests whether Scrapy can apply those. Each of our tests actually uses the part where we pass custom settings in an argument, and none of them is testing settings set to `None`, in which case we load default settings from Scrapy. That part is hard to test. I can imagine approaches to do that, but I don't think they're worth it, I'd leave it as it is.

vdusek: Yeah... agreed and thank you :slightly_smiling_face:. I created a separate test function for every test case. Seems more readable and definitely will be easier to find out where the problem is when one of the test cases fails.

janbuchar: I understand the views of both of you and agree that in this case, it's probably better to just write everything out. However, naming test functions for example, is a pain point for many people (including me) - see for example https://github.com/darrenburns/ward, which tries to solve that. But I'm by no means suggesting we adopt that, just wanted to point out that you can typically think of more useful test cases than fitting names.

tests/unit/scrapy/utils/test_apply_apify_settings.py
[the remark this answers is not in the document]
vdusek: Yeah, it would be great to create some Scrapy Actor for integration testing (the issue is already created: https://github.com/apify/apify-sdk-python/issues/148).

vdusek:
Thank you both guys @honzajavorek, @janbuchar for your opinions. Merging it.

tests/unit/scrapy/utils/test_apply_apify_settings.py
honzajavorek: Yup, writing (py)test function names is a bit annoying and while I wasn't really amused by all of the [mocha mess](https://mastodonczech.cz/@iamkale@infosec.exchange/111801113019589560) back in the early node.js days when coming from Python, I think the decision to describe test cases in plain text was a good decision.

Copilot to the rescue 😄 Today we can just write the test first and let AI to suggest the function name.

PR #180, opened 2024-01-31
Change 'input' as var name

HillaShx:
Hello team!
using `input` as a var name in Python is discouraged because of the built-in function by the same name. I noticed in other parts of the docs you used `actor_input` so I propose to change it here as well.

Hope this helps!


janbuchar (approved):
Hello, and thanks for the suggestion!

Issue #189, opened 2024-03-05
Twisted/CRITICAL: builtins.TypeError: can't pickle Selector objects (Scrapy)


honzajavorek:
My spider https://github.com/juniorguru/plucker/blob/26d1758e310b8b2451541516cf4447e4a5e4a11a/juniorguru_plucker/jobs_jobscz/spider.py runs just fine with Scrapy, but fails with critical errors when teaming up with Apify.

<details>
<summary>See exception details 💌</summary>

```
[twisted] CRITICAL Unhandled error in Deferred:

Traceback (most recent call last):
  File "/Users/honza/.local/share/mise/installs/python/3.11/lib/python3.11/asyncio/events.py", line 84, in _run
    self._context.run(self._callback, *self._args)
  File "/Users/honza/Projects/juniorguru-plucker/.venv/lib/python3.11/site-packages/twisted/internet/asyncioreactor.py", line 271, in _onTimer
    self.runUntilCurrent()
  File "/Users/honza/Projects/juniorguru-plucker/.venv/lib/python3.11/site-packages/twisted/internet/base.py", line 994, in runUntilCurrent
    call.func(*call.args, **call.kw)
  File "/Users/honza/Projects/juniorguru-plucker/.venv/lib/python3.11/site-packages/twisted/internet/task.py", line 680, in _tick
    taskObj._oneWorkUnit()
--- <exception caught here> ---
  File "/Users/honza/Projects/juniorguru-plucker/.venv/lib/python3.11/site-packages/twisted/internet/task.py", line 526, in _oneWorkUnit
    result = next(self._iterator)
  File "/Users/honza/Projects/juniorguru-plucker/.venv/lib/python3.11/site-packages/scrapy/utils/defer.py", line 102, in <genexpr>
    work = (callable(elem, *args, **named) for elem in iterable)
  File "/Users/honza/Projects/juniorguru-plucker/.venv/lib/python3.11/site-packages/scrapy/core/scraper.py", line 298, in _process_spidermw_output
    self.crawler.engine.crawl(request=output)
  File "/Users/honza/Projects/juniorguru-plucker/.venv/lib/python3.11/site-packages/scrapy/core/engine.py", line 290, in crawl
    self._schedule_request(request, self.spider)
  File "/Users/honza/Projects/juniorguru-plucker/.venv/lib/python3.11/site-packages/scrapy/core/engine.py", line 297, in _schedule_request
[...]

PR #191, opened 2024-03-06
Improve error handling for `to_apify_request` serialization failures

vdusek:
The PR addresses the problem identified in [Issue #189](https://github.com/apify/apify-sdk-python/issues/189), where `to_apify_request` could fail silently during serialization of the Scrapy request. This PR has introduced enhanced error handling to ensure users are informed if serialization of a request fails, preventing the request from being not enqueued without notification.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,9 @@
 
 ## [1.6.1](../../releases/tag/v1.6.1) - Unreleased
 
-...
+### Fixed
+
+- Improve error handling for `to_apify_request` serialization failures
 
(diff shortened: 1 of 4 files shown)


fnesveda (changes requested):
Looks good, just one note:

Maybe we should log it as errors rather than warnings, since it actually is an issue that the developers should solve. But I agree that it's better to not fail the whole Actor in this case.

vdusek:
> Maybe we should log it as errors rather than warnings, since it actually is an issue that the developers should solve. But I agree that it's better to not fail the whole Actor in this case.

Ok, `enqueue_request` logs an error message when the conversion fails.

fnesveda (approved):

Issue #190, opened 2024-03-05
Scrapy integration silently does just one POST request


honzajavorek:
We've successfully solved https://github.com/apify/apify-sdk-python/issues/185, that's awesome! 🎉 It seems to process redirects correctly now. However, I still struggle to get past a single `POST` request done.

Running code of [my spider](https://github.com/juniorguru/plucker/blob/fe89340aa21f0169ee6bddd74bb26a502c6e5bcc/juniorguru_plucker/jobs_jobscz/spider.py) through Scrapy's `crawl` command gives me:

- downloader/request_count: 1919
- downloader/request_method_count/GET: 1723
- **downloader/request_method_count/POST: 196**
- downloader/response_count: 1919
- downloader/response_status_count/200: 1130
- downloader/response_status_count/302: 789
- dupefilter/filtered: 223
- item_scraped_count: 741

I tried twice without cache and got the same numbers. Running the very same code through Apify integration gives me:

- downloader/request_count: 1721
- downloader/request_method_count/GET: 1720
- **downloader/request_method_count/POST: 1**
- downloader/response_count: 1721
- downloader/response_status_count/200: 934
- downloader/response_status_count/302: 787
- item_scraped_count: 546

I don't understand why the number of GET requests differs by two, but let's say the difference in POSTs is a bigger concern for now. Looking to the log with `debug` level turned on, I noticed this one thing repeats:

```
[apify] DEBUG [nfTk1APL]: rq.add_request.result={
  'wasAlreadyPresent': True,
  'wasAlreadyHandled': False,
  'requestId': 'uPDbxqGboeKtU6J',
  'uniqueKey': 'https://api.example.com/api/graphql/widget'}...
```

The `DEBUG [...]` part changes, but `uniqueKey` doesn't change and `wasAlreadyPresent` is `True`, suspiciously. Is it possible that Apify's request queue dedupes the requests only based on the URL? Because the POSTs all have the same URL, just different payload. Which should be very common - by definition of what POST is, or even in practical terms with all the GraphQL APIs around.

honzajavorek:
I was able to confirm my hypothesis with the following minimal example:

```python
import json
import json
from typing import Generator, cast

from scrapy import Request, Spider as BaseSpider
from scrapy.http import TextResponse


class Spider(BaseSpider):
    name = "minimal-example"

    def start_requests(self) -> Generator[Request, None, None]:
        for number in range(3):
            yield Request(
                "https://httpbin.org/post",
[...]

Issue #141, opened 2023-11-29
Check Request class in Crawlee and replicate `uniqueKey` generation logic


vdusek:
- Check `Request` class in Crawlee TS and replicate `uniqueKey` generation logic.
  - https://github.com/apify/crawlee/blob/v3.8.1/packages/core/src/request.ts#L381-L395
- The issue is based on this comment [src/apify/storages/request_queue.py#L148](https://github.com/apify/apify-sdk-python/blob/v1.3.0/src/apify/storages/request_queue.py#L148).

vdusek:
Resolved in https://github.com/apify/apify-sdk-python/pull/193

PR #193, opened 2024-03-11
Improve unique key generation logic

vdusek:
### Request Queue

- Updates `RequestQueue.add_request` method by improving its unique key generation logic.
  - The logic is exposed via new optional parameters of `add_request`.
- I also extended the docstring to make clear what it does (mainly regarding the deduplication).
- The PR introduces 3 new "helper" functions in the `apify/_utils.py`:
    - `get_short_base64_hash` based on the [_hashPayload](https://github.com/apify/crawlee/blob/v3.8.1/packages/core/src/request.ts#L397:L405).
    - `normalize_url` based on the [normalizeUrl](https://github.com/apify/apify-shared-js/blob/%40apify/input_schema%403.5.15/packages/utilities/src/utilities.client.ts#L128:L158).
    - `compute_unique_key` based on the [_computeUniqueKey](https://github.com/apify/crawlee/blob/v3.8.1/packages/core/src/request.ts#L381-L395).

### Scrapy integration

- Use this new Request Queue functionality in the Scrapy integration.
- Also add support for `scrapy.Request.dont_filter` field in the `to_apify_request`.

## Issues

- Closing both of these issues:
  - https://github.com/apify/apify-sdk-python/issues/141
    - Technical debt in Request Queue. 
  - https://github.com/apify/apify-sdk-python/issues/190
    - Bug in the Scrapy integration.

### Unit tests

- The new code is covered by unit tests.

### Manual testing / execution

The `YieldPostSpider` tests the case of the POST requests to the same URL. 
  - It's copied from [#issuecomment-1978609687](https://github.com/apify/apify-sdk-python/issues/190#issuecomment-1978609687) - thank you @honzajavorek!

```python
# spiders/yield_post.py
import json
import json
from typing import Generator, cast

from scrapy import Request, Spider as BaseSpider
from scrapy.http import TextResponse


class YieldPostSpider(BaseSpider):
    name = 'yield-post'

    def start_requests(self) -> Generator[Request, None, None]:
        for number in range(3):
            yield Request(
                'https://httpbin.org/post',
                method='POST',
                body=json.dumps(dict(code=f'CODE{number:0>4}', rate=number)),
                headers={'Content-Type': 'application/json'},
            )

    def parse(self, response: TextResponse) -> Generator[dict, None, None]:
        data = json.loads(cast(dict, response.json())['data'])
        yield data
```

The `DontFilterSpider` tests the case of the Scrapy request with `dont_filter` option. 
  - It's copied from [issuecomment-1978953372](https://github.com/apify/apify-sdk-python/issues/190#issuecomment-1978953372) - thank you @honzajavorek!

```python
from typing import Generator

from scrapy import Request, Spider as BaseSpider
from scrapy.http import TextResponse


class DontFilterSpider(BaseSpider):
    name = 'dont-filter'

    def start_requests(self) -> Generator[Request, None, None]:
        for _ in range(3):
            yield Request('https://httpbin.org/get', method='GET', dont_filter=True)

    def parse(self, response: TextResponse) -> Generator[dict, None, None]:
        yield {'something': True}
```

And `src/main.py` to execute these spiders with Apify:
```python

from scrapy.crawler import CrawlerProcess

from apify import Actor
from apify.scrapy.utils import apply_apify_settings

from .spiders.yield_post import YieldPostSpider as Spider
# from .spiders.dont_filter import DontFilterSpider as Spider


async def main() -> None:
    """Apify Actor main coroutine for executing the Scrapy spider."""
    async with Actor:
        Actor.log.info('Actor is being executed...')

        # Apply Apify settings, it will override the Scrapy project settings
        settings = apply_apify_settings()

        # Execute the spider using Scrapy CrawlerProcess
        process = CrawlerProcess(settings, install_root_handler=False)
        process.crawl(Spider)
        process.start()
```

Execute it using Scrapy:
```
scrapy crawl 'dont-filter' -o dont_filter_output.json
```

```
scrapy crawl 'yield-post' -o yield_post_output.json
```

And using Apify (need to change the Spider in the `main.py` manually):

```
apify run --purge
```

And it produces the same output :tada:.

### DIFF
--- a/src/apify/_utils.py
+++ b/src/apify/_utils.py
@@ -415,14 +415,14 @@ ListOrDictOrAny = TypeVar('ListOrDictOrAny', list, dict, Any)
 
 
 def compute_short_hash(data: bytes, *, length: int = 8) -> str:
-    """Computes a hexadecimal SHA-256 hash of the provided data and returns a substring of it.
+    """Computes a hexadecimal SHA-256 hash of the provided data and returns a substring (prefix) of it.
 
     Args:
         data: The binary data to be hashed.
         length: The length of the hash to be returned.
 
     Returns:
-        An 8-character hexadecimal hash.
+        A substring (prefix) of the hexadecimal hash of the data.
     """
     hash_object = sha256(data)
     return hash_object.hexdigest()[:length]
--- a/src/apify/_utils.py
+++ b/src/apify/_utils.py
@@ -295,8 +295,8 @@ def maybe_parse_body(body: bytes, content_type: str) -> Any:
 
 def unique_key_to_request_id(unique_key: str) -> str:
     """Generate request ID based on unique key in a deterministic way."""
-    id_ = re.sub(r'(\+|\/|=)', '', b64encode(sha256(unique_key.encode('utf-8')).digest()).decode('utf-8'))
-    return id_[:REQUEST_ID_LENGTH] if len(id_) > REQUEST_ID_LENGTH else id_
+    request_id = re.sub(r'(\+|\/|=)', '', b64encode(sha256(unique_key.encode('utf-8')).digest()).decode('utf-8'))
+    return request_id[:REQUEST_ID_LENGTH] if len(request_id) > REQUEST_ID_LENGTH else request_id
 
 
 async def force_rename(src_dir: str, dst_dir: str) -> None:
--- a/src/apify/_utils.py
+++ b/src/apify/_utils.py
@@ -295,8 +295,8 @@ def maybe_parse_body(body: bytes, content_type: str) -> Any:
 
 def unique_key_to_request_id(unique_key: str) -> str:
     """Generate request ID based on unique key in a deterministic way."""
-    id = re.sub(r'(\+|\/|=)', '', b64encode(sha256(unique_key.encode('utf-8')).digest()).decode('utf-8'))  # noqa: A001
-    return id[:REQUEST_ID_LENGTH] if len(id) > REQUEST_ID_LENGTH else id
+    id_ = re.sub(r'(\+|\/|=)', '', b64encode(sha256(unique_key.encode('utf-8')).digest()).decode('utf-8'))
+    return id_[:REQUEST_ID_LENGTH] if len(id_) > REQUEST_ID_LENGTH else id_
 
 
 async def force_rename(src_dir: str, dst_dir: str) -> None:
--- a/src/apify/_utils.py
+++ b/src/apify/_utils.py
@@ -414,21 +414,18 @@ PARSE_DATE_FIELDS_KEY_SUFFIX = 'At'
 ListOrDictOrAny = TypeVar('ListOrDictOrAny', list, dict, Any)
 
 
-def get_short_base64_hash(data: bytes) -> str:
-    """Generates an 8-character, base64-encoded SHA-256 hash of the provided data.
+def compute_short_hash(data: bytes, *, length: int = 8) -> str:
+    """Computes a hexadecimal SHA-256 hash of the provided data and returns a substring of it.
 
     Args:
         data: The binary data to be hashed.
+        length: The length of the hash to be returned.
 
     Returns:
-        An 8-character, base64-encoded hash.
+        An 8-character hexadecimal hash.
     """
-    # Create a SHA-256 hash of the payload.
     hash_object = sha256(data)
-    # Convert the hash to a base64-encoded string.
-    base64_encoded_hash = b64encode(hash_object.digest()).decode()
-    # Remove any '+' or '/' or '=' characters and return the first 8 characters.
-    return base64_encoded_hash.replace('+', '').replace('/', '').replace('=', '')[:8]
+    return hash_object.hexdigest()[:length]
 
 
 def normalize_url(url: str, *, keep_url_fragment: bool = False) -> str:
@@ -512,7 +509,7 @@ def compute_unique_key(
 
     # Compute and return the extended unique key if required.
     if use_extended_unique_key:
-        payload_hash = get_short_base64_hash(payload) if payload else ''
+        payload_hash = compute_short_hash(payload) if payload else ''
         return f'{normalized_method}({payload_hash}):{normalized_url}'
 
     # Log information if there is a non-GET request with a payload.
--- a/src/apify/_utils.py
+++ b/src/apify/_utils.py
@@ -1,11 +1,9 @@
 from __future__ import annotations
 
 import asyncio
-import base64
 import builtins
 import contextlib
 import functools
-import hashlib
 import inspect
 import json
 import mimetypes
@@ -13,10 +11,13 @@ import os
 import re
 import sys
 import time
+from base64 import b64encode
 from collections import OrderedDict
 from collections.abc import MutableMapping
 from datetime import datetime, timezone
+from hashlib import sha256
 from importlib import metadata
+from logging import getLogger
 from typing import (
     Any,
     Callable,
@@ -30,6 +31,7 @@ from typing import (
     overload,
 )
 from typing import OrderedDict as OrderedDictType
+from urllib.parse import parse_qsl, urlencode, urlparse
 
 import aioshutil
 import psutil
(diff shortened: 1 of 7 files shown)


vdusek:
@fnesveda Just FYI; not requesting a review from you, taking into account the current situation, so I added Jirka as a reviewer instead.

src/apify/_utils.py
```diff
@@ -292,9 +295,8 @@ def maybe_parse_body(body: bytes, content_type: str) -> Any:
 
 def unique_key_to_request_id(unique_key: str) -> str:
     """Generate request ID based on unique key in a deterministic way."""
-    id = re.sub(r'(\+|\/|=)', '', base64.b64encode(hashlib.sha256(unique_key.encode('utf-8')).digest()).decode('utf-8'))  # noqa: A001
-
-    return id[:REQUEST_ID_LENGTH] if len(id) > REQUEST_ID_LENGTH else id
+    id_ = re.sub(r'(\+|\/|=)', '', b64encode(sha256(unique_key.encode('utf-8')).digest()).decode('utf-8'))
```
jirimoravcik, line 298: `request_id` rather than `id_`?

src/apify/_utils.py
```diff
@@ -410,3 +412,118 @@ def validate_single(field_value: Any, expected_type: type, required: bool, name:
...
+        data: The binary data to be hashed.
+
+    Returns:
+        An 8-character, base64-encoded hash.
+    """
+    # Create a SHA-256 hash of the payload.
+    hash_object = sha256(data)
+    # Convert the hash to a base64-encoded string.
+    base64_encoded_hash = b64encode(hash_object.digest()).decode()
+    # Remove any '+' or '/' or '=' characters and return the first 8 characters.
```
jirimoravcik, line 430: I'm just wondering why do you need first 8 characters of a base64 encoded sha256 hash without `+`, `/`, and `=`? Seems rather complicated

src/apify/_utils.py
```diff
@@ -410,3 +412,118 @@ def validate_single(field_value: Any, expected_type: type, required: bool, name:
...
+    Returns:
+        A string containing the normalized URL. If normalization fails, the original URL is returned. It logs
+            a warning message in this case.
+    """
+    try:
+        # Parse the URL
+        parsed_url = urlparse(url.strip())
+        search_params = dict(parse_qsl(parsed_url.query))  # Convert query to a dict
+
+        # Remove any 'utm_' parameters
```
jirimoravcik, line 456: Shouldn't this be more generic, `utm_` isn't the only tracking parameter I'd say?

jirimoravcik (approved):

src/apify/_utils.py
vdusek: In this case, I just followed the implementation in the Crawlee - [_hashPayload](https://github.com/apify/crawlee/blob/v3.8.1/packages/core/src/request.ts#L397:L405). We had a short Slack discussion with the team keeping it 1:1 with Crawlee.

```
$ g blame -L 397,+8 packages/core/src/request.ts
3a59bfc0b (Martin Adámek 2022-07-13 20:00:07 +0200 397)     protected _hashPayload(payload: BinaryLike): string {
3a59bfc0b (Martin Adámek 2022-07-13 20:00:07 +0200 398)         return crypto
3a59bfc0b (Martin Adámek 2022-07-13 20:00:07 +0200 399)             .createHash('sha256')
3a59bfc0b (Martin Adámek 2022-07-13 20:00:07 +0200 400)             .update(payload)
3a59bfc0b (Martin Adámek 2022-07-13 20:00:07 +0200 401)             .digest('base64')
3a59bfc0b (Martin Adámek 2022-07-13 20:00:07 +0200 402)             .replace(/[+/=]/g, '')
3a59bfc0b (Martin Adámek 2022-07-13 20:00:07 +0200 403)             .substring(0, 8);
3a59bfc0b (Martin Adámek 2022-07-13 20:00:07 +0200 404)     }
```

I believe the intention behind this was only to keep alphanumeric characters in the hash. @B4nan Maybe you could clarify it?

jirimoravcik: Hmm, ok, fair enough.
For the alphanumeric representation, hash can be converted to a string representation of a number with base 16, e.g.:

```
const crypto = require('crypto');

const hash = crypto.createHash('sha256');

hash.update('123');
console.log(hash.digest(456).toString('hex'));
```

Gives:
`a665a45920422f9d417e4867efdc4fb8a04a1f3fff1fa07e998e86f7f7a27ae3`, you don't need to do base64 and replace of `+/=`.

But yeah, let's keep the parity

src/apify/_utils.py
vdusek: Basically, it's the same as in the https://github.com/apify/apify-sdk-python/pull/193#discussion_r1521200953.

@B4nan Should we add the removal of more tracking parameters (if there are any, TBH I don't know)? Or rather keep the parity.

src/apify/_utils.py
vdusek: IMO this is internal, so we can go this way in this case (if the intention was really to keep only alphanumeric chars).

src/apify/_utils.py
janbuchar: I'm convinced that this is way out of scope of the PR.

B4nan: Same as with the other problem, I'd say we want consistency now, if we should change something, it should be changed in the JS version too (which we surely can do as well).

@jirimoravcik do you have some suggestions on what else to ignore? I don't think we can find a "more generic solution", but we can blacklist more common parameters (at the same time we need to be careful to not blacklist something that might be used for other purposes).

jirimoravcik: Yeah, I just wanted to point it out for discussion. E.g. for tracking, there's https://easylist.to/easylist/easyprivacy.txt

src/apify/_utils.py
```diff
@@ -410,3 +412,118 @@ def validate_single(field_value: Any, expected_type: type, required: bool, name:
...
+            .lower()
+        )
+
+        # Retain the URL fragment if required
+        if not keep_url_fragment:
+            new_url = new_url.split('#')[0]
+
+    except Exception as exc:
+        logger.warning(f'Failed to normalize URL: {exc}')
+        return url
```
janbuchar, line 481: I'd rather make the caller responsible for handling the exception - functions like `normalize_url` probably shouldn't write logs (side effects).

tests/unit/test_utils.py
```diff
@@ -266,3 +269,100 @@ def test__budget_ow() -> None:
...
+        ('http://example.com', 'GET', b'data', False, False, 'http://example.com'),
+        ('http://example.com', 'GET', b'data', False, True, 'GET(Om6weQ85):http://example.com'),
+        ('http://example.com', 'POST', b'data', False, True, 'POST(Om6weQ85):http://example.com'),
+        ('http://example.com#fragment', 'GET', None, True, False, 'http://example.com#fragment'),
+        ('http://example.com#fragment', 'GET', None, False, False, 'http://example.com'),
+        ('http://example.com', 'DELETE', b'test', False, True, 'DELETE(n4bQgYhM):http://example.com'),
+        ('https://example.com?utm_content=test', 'GET', None, False, False, 'https://example.com'),
+        ('https://example.com?utm_content=test', 'GET', None, True, False, 'https://example.com'),
+    ],
+    ids=[
```
janbuchar, line 339: Neat, TIL

tests/unit/test_utils.py
```diff
@@ -266,3 +269,100 @@ def test__budget_ow() -> None:
...
+        'delete_with_payload_extended',
+        'get_remove_utm',
+        'get_keep_utm_fragment',
+    ],
+)
+def test_compute_unique_key(
+    url: str,
+    method: str,
+    payload: bytes | None,
+    *,
```
janbuchar, line 356: Is this necessary?

janbuchar (approved):
LGTM

tests/unit/test_utils.py
vdusek: it is because of https://docs.astral.sh/ruff/rules/boolean-type-hint-positional-argument/

src/apify/_utils.py
vdusek: implemented

PR #206, opened 2024-05-23
fix: Set a timeout for Actor cleanup

### LINKED ISSUE #200: Correctly handle actor exit
Originally reported in https://apify.slack.com/archives/CGZSN9DQC/p1715957828029029.

The user spawned an asyncio task in the actor. When the actor finished (reached the end of the `Actor` context manager block), it didn't clean up the task and the container was left there hanging.

We should either cancel all tasks when the actor is exitting or just trigger a `sys.exit` after a certain timeout.

[Relevant code](https://github.com/apify/apify-sdk-python/blob/master/src/apify/actor.py) and [JS counterpart](https://github.com/apify/apify-sdk-js/blob/master/packages/apify/src/actor.ts)

janbuchar:
- closes #200

I could not find the actual reason why the linked run got stuck, but this should make the Actor cleanup more robust as a whole.

### DIFF
--- a/src/apify/actor.py
+++ b/src/apify/actor.py
@@ -336,21 +336,23 @@ class Actor(metaclass=_ActorContextManager):
 
         self.log.info('Exiting actor', extra={'exit_code': exit_code})
 
-        await self._cancel_event_emitting_intervals()
+        async def finalize() -> None:
+            await self._cancel_event_emitting_intervals()
 
-        # Send final persist state event
-        if not self._was_final_persist_state_emitted:
-            self._event_manager.emit(ActorEventTypes.PERSIST_STATE, {'isMigrating': False})
-            self._was_final_persist_state_emitted = True
+            # Send final persist state event
+            if not self._was_final_persist_state_emitted:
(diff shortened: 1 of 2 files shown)


src/apify/actor.py
```diff
@@ -336,21 +336,23 @@ async def _exit_internal(
...
 
-        # Sleep for a bit so that the listeners have a chance to trigger
-        await asyncio.sleep(0.1)
+            # Sleep for a bit so that the listeners have a chance to trigger
+            await asyncio.sleep(0.1)
 
-        await self._event_manager.close(event_listeners_timeout_secs=event_listeners_timeout_secs)
+            await self._event_manager.close(event_listeners_timeout_secs=event_listeners_timeout_secs)
 
+        await asyncio.wait_for(finalize(), 30)
```
B4nan, line 355: lets make this timeout configurable (via `exit` options)

B4nan (approved):
image

vdusek (approved):
LGTM, thanks

vdusek (changes requested):
oh, just please add it manually to the changelog

vdusek (approved):
:rocket:

B4nan:
please release this to stable afterwards

PR #224, opened 2024-07-11
chore: update dev deps & fix linter

vdusek:


src/apify/proxy_configuration.py
```diff
@@ -325,7 +325,7 @@ async def _check_access(self: ProxyConfiguration) -> None:
         proxy_status_url = f'{self._actor_config.proxy_status_url}/?format=json'
 
         status = None
-        async with httpx.AsyncClient(proxies=await self.new_url()) as client:
+        async with httpx.AsyncClient(proxies=await self.new_url(), timeout=10) as client:
```
barjin, line 328: Is this intentional? Can we parametrize somehow? Or is this a good enough default?

barjin: Ah alright, I see that the JS version also doesn't allow to parametrize this - but uses 4 seconds.

https://github.com/apify/apify-sdk-js/blob/e49c868aaf6e1642ba40e9b10d2a532dc21a8a84/packages/apify/src/proxy_configuration.ts#L19

No biggie though, if this works here, I'm more than fine with it.

barjin (approved):
lgtm, the timeout is no huge blocker IMO :)

src/apify/proxy_configuration.py
vdusek: Yeah, I think it's not necessary to parameterize it, good default should be enough.

PR #210, opened 2024-06-04
refactor!: Preparation for v2 release

janbuchar:
- **Update pyproject.toml**
- **Use storage code from crawlee, add ApifyStorageClient**

- [x] make sure we use `APIFY_IS_AT_HOME` instead of token for all platform vs. local decisions
- [x] port over `warnedAboutMissingInitCall` from JS SDK (this is not exactly ported over, but we expect this won't happen frequently, so we just throw an exception)

### Issues

- Closes: #135
- Closes: #137
- Closes: #138
- Closes: #147
- Closes: #149
- Closes: #237

PR #249, opened 2024-08-28
fix: Make apify.log public again

janbuchar:
The module was made private in #210, but it turns out it is actually mentioned in the documentation, so we're keeping it public to make less noise.


vdusek (approved):
LGTM

PR #254, opened 2024-08-28
chore: Remove unnecessary deps

janbuchar:


vdusek (approved):
Nice, and we can remove probably `colorama` as well

PR #261, opened 2024-09-04
fix: Hotfix for batch_add_requests batch size limit

janbuchar:


src/apify/apify_storage_client/_request_queue_client.py
```diff
@@ -157,8 +158,11 @@ async def batch_add_requests(
         *,
         forefront: bool = False,
     ) -> BatchRequestsOperationResponse:
-        return BatchRequestsOperationResponse.model_validate(
-            await self._client.batch_add_requests(
+        processed = []
+        unprocessed = []
+
+        for chunk in chunked(requests, 25):
```
vdusek, line 164: I would prefer a constant or a comment explaining why 25, but it's fine since it is just tmp change, thanks

vdusek (approved):

janbuchar:
This will need an integration test... Plus the API rejects duplicate uniqueKeys in the same batch.

PR #284, opened 2024-10-02
fix: adjust tests of scrapy user data

janbuchar:
Needs https://github.com/apify/crawlee-python/pull/562


vdusek (approved):

tests/unit/scrapy/requests/test_to_apify_request.py
```diff
@@ -25,7 +25,6 @@ def test__to_apify_request__simple(spider: Spider) -> None:
     assert apify_request.url == 'https://example.com'
 
     user_data = apify_request.user_data
-    assert isinstance(user_data, dict)
```
vdusek, line 28: I don't get it why are we removing this? thanks

janbuchar: `UserData` won't pass the check

PR #286, opened 2024-10-07
chore: Add deprecation notice to actor standby port variable

janbuchar:
This does nothing on runtime, it's just housekeeping.


vdusek (approved):
cool

PR #287, opened 2024-10-08
feat: add standby URL, change default standby port

jirimoravcik:
This PR introduces a new environment variable with standby URL, it also changes the default port for the standby port env var.

### DIFF
--- a/src/apify/_configuration.py
+++ b/src/apify/_configuration.py
@@ -265,7 +265,7 @@ class Configuration(CrawleeConfiguration):
         str,
         Field(
             alias='actor_standby_url',
-            description='URL for accessing the HTTP server in Actor Standby mode',
+            description='URL for accessing web servers of Actor runs in Standby mode',
         ),
     ]
 
@@ -304,7 +304,7 @@ class Configuration(CrawleeConfiguration):
                 'actor_web_server_url',
                 'apify_container_url',
             ),
-            description='Unique public URL for accessing the Actor run web server from the outside world',
+            description='Unique public URL for accessing a specific Actor run web server from the outside world.',
         ),
     ] = 'http://localhost:4321'
 
--- a/src/apify/_configuration.py
+++ b/src/apify/_configuration.py
@@ -259,7 +259,15 @@ class Configuration(CrawleeConfiguration):
             description='TCP port for the Actor to start an HTTP server to receive messages in the Actor Standby mode',
         ),
         deprecated('Use `web_server_port` instead'),
-    ] = 4322
+    ] = 4321
+
+    standby_url: Annotated[
+        str,
+        Field(
+            alias='actor_standby_url',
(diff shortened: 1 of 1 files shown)


janbuchar:
> > Cool. What's the relationship between this and `web_server_url` or `container_url`?
> 
> `web_server_url` == `container_url`, these are unique for each run (they look something like `https://someidhere.runs.apify.net/`). Standby URL is the nice generic, e.g. `https://super-scraper-api.apify.actor`, which does automatic load balancing and scaling.

Understood, thanks. Could you mention the different in the field description please? And maybe update https://docs.apify.com/platform/actors/development/programming-interface/environment-variables as well?

jirimoravcik:
> > > Cool. What's the relationship between this and `web_server_url` or `container_url`?
> > 
> > 
> > `web_server_url` == `container_url`, these are unique for each run (they look something like `https://someidhere.runs.apify.net/`). Standby URL is the nice generic, e.g. `https://super-scraper-api.apify.actor`, which does automatic load balancing and scaling.
> 
> Understood, thanks. Could you mention the different in the field description please? And maybe update [docs.apify.com/platform/actors/development/programming-interface/environment-variables](https://docs.apify.com/platform/actors/development/programming-interface/environment-variables) as well?

I tried to improve the descriptions here.

I'm already preparing a PR for docs, but I want to mention the configs in SDKs, so that's why I'm doing it first.

janbuchar (approved):

src/apify/_configuration.py
```diff
@@ -296,7 +304,7 @@ class Configuration(CrawleeConfiguration):
                 'actor_web_server_url',
                 'apify_container_url',
             ),
-            description='Unique public URL for accessing the Actor run web server from the outside world',
+            description='Unique public URL for accessing a specific Actor run web server from the outside world.',
```
fnesveda, line 307: All the other descriptions don't have a period at the end:
```suggestion
            description='Unique public URL for accessing a specific Actor run web server from the outside world',
```

fnesveda (approved):
Cool, just one small note

src/apify/_configuration.py
```diff
@@ -259,7 +259,15 @@ class Configuration(CrawleeConfiguration):
             description='TCP port for the Actor to start an HTTP server to receive messages in the Actor Standby mode',
         ),
         deprecated('Use `web_server_port` instead'),
-    ] = 4322
+    ] = 4321
+
+    standby_url: Annotated[
+        str,
```
vdusek, line 265: In Crawlee, we use a custom URL validator function:

```python
    url: Annotated[str, BeforeValidator(validate_http_url), Field(...)]
```

https://github.com/apify/crawlee-python/blob/v0.3.8/src/crawlee/_request.py#L112

@jirimoravcik Could we use the same approach here as well?

jirimoravcik: sure

PR #289, opened 2024-10-10
fix: use HttpHeaders type in Scrapy integration

vdusek:


src/apify/scrapy/requests.py
```diff
@@ -16,6 +16,7 @@
     ) from exc
 
 from crawlee import Request as CrawleeRequest
+from crawlee._types import HttpHeaders
```
janbuchar, line 19: Just to make sure, this doesn't raise our minimum required crawlee version, right?

.pre-commit-config.yaml
```diff
@@ -13,12 +13,6 @@ repos:
         language: system
         pass_filenames: false
 
-      - id: unit-tests
-        name: Run unit tests
-        entry: make unit-tests
-        language: system
-        pass_filenames: false
-
```
janbuchar, line 21: Yes please!!

janbuchar (approved):

src/apify/_configuration.py
```diff
@@ -262,12 +263,13 @@ class Configuration(CrawleeConfiguration):
...
     standby_url: Annotated[
-        str,
+        str | None,
+        BeforeValidator(validate_http_url),
         Field(
             alias='actor_standby_url',
             description='URL for accessing web servers of Actor runs in Standby mode',
         ),
-    ]
+    ] = None
```
janbuchar, line 272: It might be worth it to use some nonsense URL here instead, just for the ease of use.

src/apify/scrapy/requests.py
vdusek: Actually, it does, because in Scrapy integration we use conversion `HttpHeaders(...)` for headers. Thanks.

PR #311, opened 2024-10-29
ci: Make integrations tests more stable

### LINKED ISSUE #301: Make integrations tests more stable
Our integration tests are highly unstable, often requiring multiple attempts to pass. Below are the two main issues we are encountering...

### Issues during Actor creation

The first problem occurs when creating an Actor in `conftest/make_actor` ([conftest.py#L247](https://github.com/apify/apify-sdk-python/blob/v2.0.1b9/tests/integration/conftest.py#L247)):

```python
    ...
    build = await actor_client.build(version_number='0.0', wait_for_finish=300)
    assert build['status'] == ActorJobStatus.SUCCEEDED
```

The test frequently fails with the following assertion error:

```
>       assert build['status'] == ActorJobStatus.SUCCEEDED
E       AssertionError: assert 'READY' == <ActorJobStat...: 'SUCCEEDED'>
E         
E         - SUCCEEDED
E         + READY
```

### Issues during Actor execution

Another common issue arises when checking the Actor's run status in the integration tests:

```python
    ...
    run_result = await actor.call()
    assert run_result is not None
    assert run_result['status'] == 'SUCCEEDED'
```

Here, the test often fails with a similar mismatch in the expected status:

```
>       assert build['status'] == ActorJobStatus.SUCCEEDED
E       AssertionError: assert 'READY' == <ActorJobStat...: 'SUCCEEDED'>
E         
E         - SUCCEEDED
E         + READY
```

vdusek:
Both the build Actor step and the run Actor step.

Closes #301

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -242,9 +242,12 @@ async def make_actor(
         actor_client = apify_client_async.actor(created_actor['id'])
 
         print(f'Building Actor {actor_name}...')
-        build = await actor_client.build(version_number='0.0', wait_for_finish=300)
+        build_result = await actor_client.build(version_number='0.0', wait_for_finish=60)
+        build_client = apify_client_async.build(build_result['id'])
+        build_client_result = await build_client.wait_for_finish(wait_secs=600)
 
-        assert build['status'] == ActorJobStatus.SUCCEEDED
+        assert build_client_result is not None
+        assert build_client_result['status'] == ActorJobStatus.SUCCEEDED
 
         # We only mark the client for cleanup if the build succeeded,
         # so that if something goes wrong here,
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -244,7 +244,7 @@ async def make_actor(
         print(f'Building Actor {actor_name}...')
         build_result = await actor_client.build(version_number='0.0')
         build_client = apify_client_async.build(build_result['id'])
-        build_client_result = await build_client.wait_for_finish(wait_secs=300)
+        build_client_result = await build_client.wait_for_finish(wait_secs=600)
 
         assert build_client_result is not None
         assert build_client_result['status'] == ActorJobStatus.SUCCEEDED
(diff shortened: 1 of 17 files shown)


tests/integration/conftest.py
```diff
@@ -56,6 +56,53 @@ def apify_client_async() -> ApifyClientAsync:
...
 
 
+class RunActorFunction(Protocol):
+    """A type for the `run_actor` fixture."""
+
+    def __call__(
+        self,
+        actor: ActorClientAsync,
+        *,
+        run_input: Any = None,
```
janbuchar, line 66: Isn't `run_input` a dict?

tests/integration/conftest.py
```diff
@@ -56,6 +56,53 @@ def apify_client_async() -> ApifyClientAsync:
...
+
+        assert isinstance(call_result, dict), 'The result of ActorClientAsync.call() is not a dictionary.'
+        assert 'id' in call_result, 'The result of ActorClientAsync.call() does not contain an ID.'
+
+        run_client = apify_client_async.run(call_result['id'])
+        run_result = await run_client.wait_for_finish(wait_secs=600)
+
+        assert isinstance(run_result, dict), 'The result of RunClientAsync.wait_for_finish() is not a dictionary.'
+        assert 'status' in run_result, 'The result of RunClientAsync.wait_for_finish() does not contain a status.'
+        return run_result
```
janbuchar, line 101: It is quite an intrusive change, but couldn't we do `from apify._models import ActorRun` and `return ActorRun.model_validate(run_result)` here?

janbuchar (commented):
I like this, just one idea. Feel free to dismiss it if you like :slightly_smiling_face:

tests/integration/conftest.py
vdusek: Hmm, let's try it

tests/integration/conftest.py
vdusek: It really is `Any` - https://github.com/apify/apify-client-python/blob/v1.8.1/src/apify_client/clients/resource_clients/actor.py#L591

janbuchar: Huh, weird, but not important.

janbuchar (approved):
LGTM

Issue #310, opened 2024-10-29
Handle request list and proxy configuration inputs in a user-friendly way


janbuchar:
- https://docs.apify.com/platform/actors/development/actor-definition/input-schema/specification/v1#array - request list editor
- https://docs.apify.com/platform/actors/development/actor-definition/input-schema/specification/v1#object - proxy editor

An `actor.json` like this:

```json
{
    "actorSpecification": 1,
    "name": "example-actor",
    "version": "0.0",
    "buildTag": "latest",
    "environmentVariables": {},
    "input": {
        "properties": {
            "startUrls": {
                "title": "Start URLs",
                "type": "array",
                "description": "URLs to start with",
                "prefill": [
                    {
                        "url": "https://apify.com"
                    }
                ],
                "editor": "requestListSources"
            },
            "proxyConfiguration": {
                "title": "Proxy configuration",
                "type": "object",
                "description": "Select proxies to be used by your crawler.",
                "prefill": {
                    "useApifyProxy": true
                },
                "editor": "proxy"
            }
        }
    }
}

```

will have an input editor like this:

![image](https://github.com/user-attachments/assets/d9daff56-f8f2-4b80-948e-3beeb821374c)

and calling `Actor.get_input()` in the actor will result into something like this:

```py
{
    "proxyConfiguration": {
        "useApifyProxy": True,
        "apifyProxyGroups": ["BUYPROXIES63811", "SHADER"],
    },
    "startUrls": [
        {"url": "https://google.com", "method": "GET"},
        {"url": "https://salas.hu", "method": "GET"},
        {"url": "https://trivago.hu", "method": "GET"},
        {"requestsFromUrl": "https://lorem.se/ipsum.txt", "method": "GET"},
    ],
}
```

It would be awesome if we could implement some helpers to simplify converting these structures into `crawlee.storage.RequestList` or `apify.ProxyConfiguration`. 

Since it is pretty convenient to have the input structure described with a Pydantic model (so you can do `Input.model_validate(await Actor.get_input())`, I would love to explore the possibility of adding some Pydantic models that you could use in your `Input` models and then pass them to `ProxyConfiguration` or `RequestList` constructors (or some factory methods).

Pijukatel:
Pydantic seems natural for handling inputs, on the other hand it seems like it was already intentionally removed: https://github.com/apify/apify-sdk-python/pull/74

I do not have enough historical context. Is the reasoning behind not using Pydantic here still valid?

[...]

PR #326, opened 2024-11-18
feat: Handle request list user input

Pijukatel:
Add helper function to handle request list inputs.

Closes: #310


tests/unit/actor/test_actor_create_request_list.py
```diff
@@ -0,0 +1,204 @@
...
+    'https://www.something.net',
+    'http://nowww.cz',
+    'https://with-hypen.com',
+    'http://number1.com',
+    'http://www.number.123',
+    'http://many.dots.com',
+    'http://a.com',
+    'http://www.something.com/somethignelse'
+    'http://www.something.com/somethignelse.txt',
+    #  "http://non-english-chars-á.com"  # re module not suitable, regex can do this with \p{L}. Do we want this?
```
Pijukatel, line 178: JS version uses \p{L} and thus it supports this use case. Should we port it to Python as well? It would require using custom **regex** package instead of Python standard **re** package.

JS implementation:
https://github.com/apify/crawlee/blob/master/packages/utils/src/internals/general.ts#L9

Python regex package that supports \p{L} syntax:
https://github.com/mrabarnett/mrab-regex

src/apify/_actor.py
```diff
@@ -974,6 +977,22 @@ async def create_proxy_configuration(
 
         return proxy_configuration
 
+    @staticmethod
```
Pijukatel, line 980: _ActorType class seems very large. Maybe we should compose it from smaller components instead of having big class with so many methods?

tests/unit/actor/test_actor_create_request_list.py
```diff
@@ -0,0 +1,204 @@
...
+
+import re
+import typing
+from unittest import mock
+from unittest.mock import call
+
+import pytest
+
+from crawlee._request import UserData  #  TODO: Make public in Crawlee?
+from crawlee._types import HttpHeaders, HttpMethod  #  TODO: Make public in Crawlee?
```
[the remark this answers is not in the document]
janbuchar: What do you mean, make them public? If you do for example this:

```py
class A(BaseModel):
  method: HttpMethod
```

and `A` gets exported... that's fine. If you just need to export it so that the end user can do `from apify import HttpMethod`, then I'm not sure if it's a good idea.

src/apify/_actor.py
janbuchar: You're right, feel free to open an issue for that. We better do a similar stuff in the JS SDK as well too.

tests/unit/actor/test_actor_create_request_list.py
janbuchar: 1. if the package is not huge in terms of disk space, there's no problem in adding it :shrug: 
2. **however**, if `\p{L}` is the only thing that's missing, I believe that `\w` does the same in Python, doesn't it?

src/apify/_actor.py
B4nan: I wouldn't be so sure about this. `Actor` class is the main touchpoint, it consolidates all the public API, and moving the implementation somewhere else would only complicate development for us probably.

Not sure if python has traits like PHP does, JS doesn't have a thing like that, and even if it would have them, I don't think it would help much.

tests/unit/actor/test_actor_create_request_list.py
Pijukatel: **\p{L}** should match just letters, not numbers or underscore and it should match non-English letters like á, í, é, ...as well
I think that **\w** is basically matching **[a-zA-Z0-9_]** , so  does not include letters like á, í, é, ...

B4nan: > however, if \p{L} is the only thing that's missing, I believe that \w does the same in Python, doesn't it?

`\w` is not the same as `\p{L}`, but I wouldn't care about this if its not natively supported by python.

src/apify/_actor.py
janbuchar: Python has multiple inheritance, so we could have something like traits so that there is some internal division of the public methods.

Thanks for saying that we shouldn't break the public API @B4nan, I assumed the comment was about internal changes.

tests/unit/actor/test_actor_create_request_list.py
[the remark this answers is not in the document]
Pijukatel: I mean that in apify-sdk-python I use internal crawlee types. So either I expose those types in crawlee or I will keep using internal types of another module, which is also not very good.

So I would like the user to be able to do 
from crawlee import HttpMethod
instead of 
from crawlee._types import HttpMethod

src/apify/_actor.py
B4nan: I am still not convinced it would help. I did something like this in one project years ago, just to reduce the size of a class, I introduced unnecessary abstraction, and it is a pain in the ass to maintain it.

Let's write down some concrete ideas and decide on that.

janbuchar: And please let's do so in a different issue and not derail this PR :slightly_smiling_face:

tests/unit/actor/test_actor_create_request_list.py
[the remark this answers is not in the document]
janbuchar: Keep reaching inside crawlee. It's better to make an exception for ourselves instead of committing to maintaining BC with stuff that we only need to make for SDK by making it public.

tests/unit/actor/test_actor_create_request_list.py
janbuchar: Quoting https://docs.python.org/3/howto/regex.html#matching-characters,

> Let’s take an example: \w matches any alphanumeric character. If the regex pattern is expressed in bytes, this is equivalent to the class [a-zA-Z0-9_]. If the regex pattern is a string, \w will match all the characters marked as letters in the Unicode database provided by the [unicodedata](https://docs.python.org/3/library/unicodedata.html#module-unicodedata) module. You can use the more restricted definition of \w in a string pattern by supplying the [re.ASCII](https://docs.python.org/3/library/re.html#re.ASCII) flag when compiling the regular expression.

So... good enough?

B4nan: So is that particular test with `http://non-english-chars-á.com` commented out for no reason and it will work fine? I guess not.

Either way, even if it's not working, I would say it is good enough to ignore this difference for now and only deal with that once someone actually tries to work with such a domain and asks us to fix it.

pyproject.toml
```diff
@@ -141,6 +141,9 @@ indent-style = "space"
 docstring-quotes = "double"
 inline-quotes = "single"
 
+[tool.ruff.lint.flake8-type-checking]
+runtime-evaluated-base-classes = ["pydantic.BaseModel"]
+
```
vdusek, line 146: Am I correct in understanding that with this setting, we should be able to use type fields for Pydantic models within `TYPE_CHECKING`? Or what is this good for? If we are gonna use it, please update it in the Crawlee & Client as well.

src/apify/storages/_known_actor_input_keys.py
```diff
@@ -0,0 +1,29 @@
...
+    def __str__(self) -> str:
+        return self._name
+
+    def __repr__(self) -> str:
+        return str(self)
+
+
+class _StartUrls(_KnownInputKey):
+    url = 'url'
+    requestsFromUrl = 'requestsFromUrl'  # noqa: N815  # Intentional to respect actual naming of input keys.
```
vdusek, line 16: couldn't we use pydantic & alias field instead of "noqa: N815"?

src/apify/storages/_actor_inputs.py
```diff
@@ -0,0 +1,126 @@
...
+    http_client is client that will be used to send get request to url defined in requestsFromUrl
+
+    Example:
+        actor_start_urls_input = [
+            # Gather urls from response body.
+            {'requestsFromUrl': 'https://crawlee.dev/file.txt', 'method': 'GET'},
+            # Directly include this url.
+            {'url': 'https://crawlee.dev', 'method': 'GET'}
+        ]
+    """
```
janbuchar, line 53: you could greatly simplify this with

```py
from pydantic import TypeAdapter
from typing import Union

adapter = TypeAdapter(list[Union[_RequestsFromUrlInput, _SimpleUrlInput]])
adapter.validate_python(actor_start_urls_input)
```

pyproject.toml
Pijukatel: Without this, ruff is complaining that annotations used in Pydantic classes are not conditionally imported with **if TYPE_CHECKING**. This tells ruff to not complain about it as those annotations are needed on runtime and not only during static code analysis.

src/apify/_platform_event_manager.py
```diff
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import asyncio
-from datetime import datetime  # noqa: TCH003
+from datetime import datetime
```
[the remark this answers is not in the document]
Pijukatel: https://docs.astral.sh/ruff/settings/#lint_flake8-type-checking_runtime-evaluated-base-classes

pyproject.toml
Pijukatel: I will do small PRs for other repos to include it there as well.

Pijukatel:
Sorry for the force-push. Due to my previous mainly Gerrit-based experience I was expecting this **Rebase** button in Github to do something more clean in the log...

src/apify/storages/_actor_inputs.py
Pijukatel: Done

src/apify/storages/_actor_inputs.py
```diff
@@ -0,0 +1,126 @@
...
+
+class _RequestsFromUrlInput(_RequestDetails):
+    requests_from_url: str = Field(alias=ActorInputKeys.startUrls.requestsFromUrl)
+
+
+class _SimpleUrlInput(_RequestDetails):
+    url: str
+
+
+async def create_request_list(
```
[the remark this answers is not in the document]
Pijukatel: Done

src/apify/storages/_known_actor_input_keys.py
Pijukatel: I got rid of the whole file. It was causing more harm than benefit

pyproject.toml
```diff
@@ -141,6 +141,9 @@ indent-style = "space"
 docstring-quotes = "double"
 inline-quotes = "single"
 
+[tool.ruff.lint.flake8-type-checking]
+runtime-evaluated-base-classes = ["pydantic.BaseModel", "crawlee.configuration.Configuration"]
```
Pijukatel, line 145: Even though crawlee.configuration.Configuration inherits from pydantic_settings.BaseSettings and that one from pydantic.BaseModel, ruff has some issues in following inheritance hierarchy too far. So until that changes, some models will have to be explicitly mentioned even though they inherit from pydantic.BaseModel.
See closed issue where this is described as known limitation to certain degree:
https://github.com/astral-sh/ruff/issues/8725

janbuchar (changes requested):
This is looking pretty good! Just a bunch of nits to iron out...

src/apify/storages/request_list.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+class _SimpleUrlInput(_RequestDetails):
+    url: str
+
+
+url_input_adapter = TypeAdapter(list[Union[_RequestsFromUrlInput, _SimpleUrlInput]])
+
+
+class RequestList(CrawleeRequestList):
```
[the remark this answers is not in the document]
Pijukatel: Not yet available in version used in sdk. Added it as comment for now.

src/apify/storages/request_list.py
```diff
@@ -0,0 +1,137 @@
...
+    """Extends crawlee RequestList."""
+
+    @classmethod
+    async def open(
+        cls,
+        name: str | None = None,
+        actor_start_urls_input: list[dict[str, Any]] | None = None,
```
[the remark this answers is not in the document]
Pijukatel: Done

src/apify/storages/request_list.py
```diff
@@ -0,0 +1,137 @@
...
+            url=request_input.url,
+            payload=request_input.payload.encode('utf-8'),
+            headers=request_input.headers,
+            user_data=request_input.user_data,
+        )
+        for request_input in simple_url_inputs
+    ]
+
+
+async def _create_requests_from_url(
```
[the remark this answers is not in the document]
Pijukatel: Done

tests/unit/actor/test_request_list.py
```diff
@@ -0,0 +1,231 @@
...
+            'payload': 'some_payload',
+            'userData': {'another_key': 'another_value'},
+        },
+    ]
+
+    mocked_read_outputs = ('' for url in actor_start_urls_input)
+    http_client = HttpxHttpClient()
+    with mock.patch.object(
+        http_client, 'send_request', return_value=_create_dummy_response(mocked_read_outputs)
+    ) as mocked_send_request:
```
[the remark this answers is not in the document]
Pijukatel: Done

src/apify/storages/request_list.py
```diff
@@ -0,0 +1,137 @@
...
+
+class _RequestDetails(BaseModel):
+    method: HttpMethod = 'GET'
+    payload: str = ''
+    headers: dict[str, str] = Field(default_factory=dict)
+    user_data: dict[str, str] = Field(default_factory=dict, alias='userData')
+
+
+class _RequestsFromUrlInput(_RequestDetails):
+    requests_from_url: str = Field(alias='requestsFromUrl')
```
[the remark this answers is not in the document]
janbuchar: I get it. It shouldn't even be necessary to have both `default_factory` and an actual default, but `mypy` seems to disagree.

src/apify/storages/_request_list.py
```diff
@@ -0,0 +1,140 @@
...
+    async def open(
+        name: str | None = None,
+        request_list_sources_input: list[dict[str, Any]] | None = None,
+        http_client: BaseHttpClient | None = None,
+    ) -> RequestList:
+        """Creates RequestList from Actor input requestListSources.
+
+        name is name of the returned RequestList
+        request_list_sources_input  can contain list dicts with either url or requestsFromUrl key
+        http_client is client that will be used to send get request to url defined in requestsFromUrl
```
janbuchar, line 56: Use the same docblock style as we do in the rest of the code please.

src/apify/storages/_request_list.py
```diff
@@ -0,0 +1,140 @@
...
+
+        Example request_list_sources_input:
+            [
```
janbuchar, line 59: We wrap examples in markdown code blocks - https://github.com/apify/crawlee-python/blob/master/src/crawlee/http_crawler/_http_crawler.py#L30-L52

src/apify/storages/_request_list.py
```diff
@@ -0,0 +1,140 @@
...
+        return await RequestList._create_request_list(name, request_list_sources_input, http_client)
+
+    @staticmethod
+    async def _create_request_list(
+        name: str | None, request_list_sources_input: list[dict[str, Any]], http_client: BaseHttpClient | None
+    ) -> RequestList:
+        if not http_client:
+            http_client = HttpxHttpClient()
+
+        ulr_inputs = url_input_adapter.validate_python(request_list_sources_input)  # instance of list[Union[...]]
```
janbuchar, line 76: tpyo, plus the comment was just meant for you, it's not that useful in the code.

tests/unit/actor/test_request_list.py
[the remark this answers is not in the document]
janbuchar: If you want to go with mocking, then please mock `BaseHttpClient` instead of a concrete implementation. I have a feeling that using `respx` could result in a more readable code - did you try it?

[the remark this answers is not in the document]
Pijukatel: Yes, respx is actually quite convenient to use. Thanks for the tip.

src/apify/storages/_request_list.py
Pijukatel: Done

src/apify/storages/request_list.py
```diff
@@ -0,0 +1,137 @@
...
+def _create_requests_from_input(simple_url_inputs: list[_SimpleUrlInput]) -> list[Request]:
+    return [
+        Request.from_url(
+            method=request_input.method,
+            url=request_input.url,
+            payload=request_input.payload.encode('utf-8'),
+            headers=request_input.headers,
+            user_data=request_input.user_data,
+        )
+        for request_input in simple_url_inputs
```
barjin: Btw thinking about this class reminded me of this: https://github.com/apify/crawlee/pull/2466

Basically, in JS Crawlee, we used to instantiate all the requests at the beginning (like here), but that ate up much more memory than just keeping the POJO objects (I guess `dict`s in Python) and instantiating the `Request` class only once we really need it. Maybe Python handles this better (has slimmer class instances).

It was at that time we learned that users love to make >1.000.000 requests long RLs 😅

Pijukatel: That is a valid point, but I think it will be done later in https://github.com/apify/crawlee-python/issues/99 as the current class is openly not as mature as the JS version:

https://github.com/apify/crawlee-python/blob/master/src/crawlee/storages/_request_list.py#L23

src/apify/storages/_request_list.py
```diff
@@ -0,0 +1,148 @@
...
+    async def _create_request_list(
+        name: str | None, request_list_sources_input: list[dict[str, Any]], http_client: BaseHttpClient | None
+    ) -> RequestList:
+        if not http_client:
+            http_client = HttpxHttpClient()
+
+        ulr_inputs = url_input_adapter.validate_python(request_list_sources_input)
+
+        simple_url_inputs = [url_input for url_input in ulr_inputs if type(url_input) is _SimpleUrlInput]
+        remote_url_inputs = [url_input for url_input in ulr_inputs if type(url_input) is _RequestsFromUrlInput]
```
janbuchar, line 87:
```suggestion
        url_inputs = url_input_adapter.validate_python(request_list_sources_input)

        simple_url_inputs = [url_input for url_input in url_inputs if isinstance(url_input, _SimpleUrlInput)]
        remote_url_inputs = [url_input for url_input in url_inputs if isinstance(url_input, _RequestsFromUrlInput)]
```

src/apify/storages/_request_list.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+class _SimpleUrlInput(_RequestDetails):
+    url: str
+
+
+url_input_adapter = TypeAdapter(list[Union[_RequestsFromUrlInput, _SimpleUrlInput]])
+
+
+# @docs_group('Classes')  # Not yet available in crawlee
```
janbuchar, line 39: PR here https://github.com/apify/apify-sdk-python/pull/328

src/apify/storages/request_list.py
janbuchar: Agreed, let's make sure the SDK supports everything the platform can offer first, then we can optimize.

janbuchar (changes requested):
One last suggestion :slightly_smiling_face:

src/apify/storages/_request_list.py
```diff
@@ -0,0 +1,150 @@
...
+
+
+class _SimpleUrlInput(_RequestDetails):
+    url: str
+
+
+url_input_adapter = TypeAdapter(list[Union[_RequestsFromUrlInput, _SimpleUrlInput]])
+
+
+@docs_group('Classes')  # Not yet available in crawlee
```
janbuchar, line 41:
```suggestion
@docs_group('Classes')
```

janbuchar (changes requested):

janbuchar (approved):

PR #328, opened 2024-11-20
docs: reexport storages documentation from `crawlee-python`

barjin:
Updates the `@apify/docusaurus-plugin-typedoc-api` plugin, which enables us to use the unified Python documentation pipeline (instead of handling it in each Python project separately).

Decorates the symbols to be documented with `@docs_group` and `@docs_name` decorators (just like in [crawlee-python](https://github.com/apify/crawlee-python/pull/655)). Those decorators could be moved up into some shared package (e.g. `apify-shared`, which already has the `@ignore_docs` decorator).

Uses the new `reexports` feature of the plugin to load external API docs (from `crawlee.dev/python`) and re-render them in the current documentation instance.


src/apify/_actor.py
```diff
@@ -39,6 +39,8 @@
 MainReturnType = TypeVar('MainReturnType')
 
 
+@docs_name('Actor')
+@docs_group('Classes')
 class _ActorType:
```
vdusek, line 44: Thinking whether rather not to use just a one decorator with keyword arguments?

```suggestion
@docs(name='Actor', group='Classes')
class _ActorType:
```

barjin: Sure, that's absolutely doable 👍🏽  It would require some more patching of the docusaurus plugin and I'd rather get these changes out now (because of the [broken links](https://github.com/apify/apify-docs/actions/runs/11924202390) this PR fixes). 

We can definitely put this on the todo list for the next iteration of the Python docs (together with putting this decorator in `apify-shared`... or somewhere, as I'm mentioning in the PR description). 

If you feel like we should have it now, I can do it too, but... I'd rather pay attention to other things, this little detour took way too much of my time yesterday 😅

vdusek: Sure, it can wait, thanks :slightly_smiling_face:

vdusek (approved):
lgtm otherwise

barjin:
Is it alright to merge with the failing 3.13 Integration test? Is it the flaky one you guys were talking about, or is there something I should do about it?

vdusek:
> Is it alright to merge with the failing 3.13 Integration test? Is it the flaky one you guys were talking about, or is there something I should do about it?

It passes now, let's merge it.

PR #347, opened 2024-12-03
fix: add upper bound of HTTPX version

vdusek:
- HTTPX 0.27 supports both the `proxies` and `proxy` args, while version 0.28 supports only the `proxy`. 
- I am adopting the updated arg name to align with the newer version. 
- However, we cannot fully upgrade to 0.28 as there are more issues.


janbuchar (approved):
Thanks for the quick fix! I suppose the same is happening in Crawlee?

PR #361, opened 2024-12-18
fix: allow calling `Actor.reboot()` from migrating handler, align reboot behavior with JS SDK

fnesveda:
This fixes several issues with the `Actor.reboot()` behavior:
- `Actor.reboot()` waits for all event handlers to finish, but if itself it was called in an event handler, it would be waiting for itself, getting into a deadlock
- `Actor.reboot()` in the JS SDK triggers event handlers for the `migrating` and `persistState` events, but in the Python SDK it was triggering only the `persistState` handlers

This aligns the behavior to work like the JS SDK, and prevents reboot getting into an infinite loop by allowing it to be called only once.

Related PR in JS SDK: https://github.com/apify/apify-sdk-js/pull/345


src/apify/_actor.py
```diff
@@ -839,12 +841,29 @@ async def reboot(
...
-        self._event_manager.emit(event=Event.PERSIST_STATE, event_data=EventPersistStateData(is_migrating=True))
+        # Call all the listeners for the PERSIST_STATE and MIGRATING events, and wait for them to finish.
+        # We can't just emit the events and wait for all listeners to finish,
+        # because this method might be called from an event listener itself, and we would deadlock.
+        persist_state_listeners = chain.from_iterable(
+            (self._event_manager._listeners_to_wrappers[Event.PERSIST_STATE] or {}).values()  # noqa: SLF001
+        )
+        migrating_listeners = chain.from_iterable(
+            (self._event_manager._listeners_to_wrappers[Event.MIGRATING] or {}).values()  # noqa: SLF001
+        )
```
janbuchar, line 861: I might be missing something, but the `chain.from_iterable` doesn't seem to have a purpose here.

src/apify/_actor.py
```diff
@@ -839,12 +841,29 @@ async def reboot(
...
 
-        await self._event_manager.__aexit__(None, None, None)
+        await asyncio.gather(
+            *[listener(EventPersistStateData(is_migrating=True)) for listener in persist_state_listeners],
+            *[listener(EventMigratingData()) for listener in migrating_listeners],
```
janbuchar, line 865: I might be bringing up something we've already discussed, but what if we only emitted `EventPersistData(is_migrating=True)` and skipped the "migrating" event? If it's just for parity with the JS SDK, maybe we could just... not port it here?

fnesveda: I think @B4nan said that in JS crawlee, crawlers are listening for the `migrating` event and pausing once they receive it. I suppose this behavior should be consistent between JS and Python.

janbuchar: Right, it makes sense to make sure that those get called as well. Could you drop a comment about that in the code?

fnesveda: Sure, will do

src/apify/_actor.py
janbuchar: Huh, thanks. SDK already depends on Crawlee and that depends on `more_itertools` which contain `flatten` - could you add that dependency to SDK as well and use that instead? I know it serves no practical purpose other than readability.

fnesveda: Sure, will do.

janbuchar (approved):
LGTM

Issue #250, opened 2024-08-28
Move code samples to separate files and revise the whole docs 


vdusek:
- Move doc code samples to separate files to be able to perform static analysis. 
  - In the same way as we did in Crawlee in https://github.com/apify/crawlee-python/pull/457.
- Revise the whole documentation, update the text and fix the code samples if necessary.

PR #378, opened 2025-01-22
docs: Move code samples to files and other updates

vdusek:
### Description

- Relocated all code samples to dedicated files for being able to perform static analysis.
- Updated home page example and moved it to a dedicated file.
- Updated `pyproject.toml` configurations to include only directories with source files (`website/` is added there, same as in Crawlee).
- Revised most of the documentation content.
- Updated legacy examples (new imports, types, ...).
- Added a new Crawlee guide.
- Fixed the Crawlee Playwright example in the README.
- The text in the Concepts section was remained as it is, only some minor updates there. Code samples were moved to separate files and corrected if necessary as was said before.
- Ensured all examples in the documentation are fully executable, meeting the prerequisites for the upcoming "Run on Apify" feature.
- The docs content files were renamed to undescore-based, in the same way as we have in Crawlee.

### Issues

- Closes: #250

### Testing

- All code samples pass checks with Ruff and Mypy.

### Checklist

- [x] CI passed


docs/01_overview/code/actor_structure/py.typed
Pijukatel, line 1: What is the point of adding `py.typed` files to the example folders?

vdusek: Because we do `from .main import main` here, Mypy considers it as a Python package and is complaining otherwise.

> Cannot find implementation or library stub for module named ".main"

Pijukatel (approved):

janbuchar (approved):
I did not read this in depth, but it looks OK on a high level.

PR #381, opened 2025-01-23
docs: bump `docusaurus-plugin-typedoc-api`

barjin:
Adds the correct inheritance/overrides rendering, multiple function signatures (below), correct optional / keyword-only pills for function arguments.

Reexports https://crawlee.dev/python/api/class/Request .

https://github.com/user-attachments/assets/9f3c9afd-0609-4f71-b2d3-f1698afc19dd


barjin:
The Request reexport might be slightly broken (inherited symbols not getting rendered) until we rebuild `crawlee-python` docs. The docs build there is now manual so we don't have unreleased features in the docs, right?

B4nan:
I think it is triggered by release workflow, but also for beta releases?

B4nan (approved):

vdusek:
> The docs build there is now manual

Currently it is only manual.

image

website/docusaurus.config.js
```diff
@@ -117,6 +117,10 @@ module.exports = {
                         url: 'https://crawlee.dev/python/api/class/RequestQueue',
                         group: 'Classes',
                     },
+                    {
+                        url: 'https://crawlee.dev/python/api/class/Request',
+                        group: 'Classes',
+                    },
```
vdusek, line 123: I already added this.

https://github.com/apify/apify-sdk-python/blob/master/website/docusaurus.config.js#L120:L124

barjin: Cool! As I expected, it caused some rendering issues with the reexported `Request` docs (missing inherited symbols). To fix this, we need to deploy Crawlee docs with the latest version of the plugin (and then rebuild SDK docs, preferably with the changes from this PR).

obrazek

janbuchar:
> > The docs build there is now manual
> 
> Currently it is only manual.
> 

We should align that with crawlee-python and trigger it on pre-releases

vdusek (approved):

PR #385, opened 2025-01-29
fix: Fix RQ usage in Scrapy scheduler

vdusek:
Relates: https://github.com/apify/actor-templates/issues/303

### DIFF
--- a/src/apify/scrapy/middlewares/apify_proxy.py
+++ b/src/apify/scrapy/middlewares/apify_proxy.py
@@ -78,11 +78,6 @@ class ApifyHttpProxyMiddleware:
         Raises:
             ValueError: If username and password are not provided in the proxy URL.
         """
-        # Do not use proxy for robots.txt, as it causes 403 Forbidden.
-        if request.url.endswith('/robots.txt'):
-            request.meta.pop('proxy', None)
-            return
-
         Actor.log.debug(f'ApifyHttpProxyMiddleware.process_request: request={request}, spider={spider}')
         url = await self._get_new_proxy_url()
 
--- a/src/apify/scrapy/middlewares/apify_proxy.py
+++ b/src/apify/scrapy/middlewares/apify_proxy.py
@@ -62,7 +62,7 @@ class ApifyHttpProxyMiddleware:
         if use_apify_proxy is not True:
             Actor.log.warning(
                 'ApifyHttpProxyMiddleware is not going to be used. Actor input field '
-                '"proxyConfiguration.useApifyProxy" is probably set to False.'
+                '"proxyConfiguration.useApifyProxy" is set to False.'
             )
             raise NotConfigured
 
--- a/src/apify/scrapy/middlewares/apify_proxy.py
+++ b/src/apify/scrapy/middlewares/apify_proxy.py
@@ -78,6 +78,11 @@ class ApifyHttpProxyMiddleware:
         Raises:
             ValueError: If username and password are not provided in the proxy URL.
         """
+        # Do not use proxy for robots.txt, as it causes 403 Forbidden.
+        if request.url.endswith('/robots.txt'):
+            request.meta.pop('proxy', None)
+            return
+
         Actor.log.debug(f'ApifyHttpProxyMiddleware.process_request: request={request}, spider={spider}')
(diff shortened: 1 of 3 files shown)


src/apify/scrapy/middlewares/apify_proxy.py
```diff
@@ -78,6 +78,11 @@ async def process_request(self, request: Request, spider: Spider) -> None:
         Raises:
             ValueError: If username and password are not provided in the proxy URL.
         """
+        # Do not use proxy for robots.txt, as it causes 403 Forbidden.
```
janbuchar, line 81: Like... universally, everywhere? I don't mind it, it just seems weird.

src/apify/scrapy/scheduler.py
```diff
@@ -52,8 +54,15 @@ def open(self, spider: Spider) -> None:  # this has to be named "open"
...
 
         async def open_queue() -> RequestQueue:
-            custom_loop_apify_client = ApifyStorageClient(configuration=Configuration.get_global_configuration())
-            return await RequestQueue.open(storage_client=custom_loop_apify_client)
+            config = Configuration.get_global_configuration()
+
+            # Use the ApifyStorageClient if the Actor is running on the Apify platform,
+            # otherwise use the MemoryStorageClient.
+            storage_client = (
+                ApifyStorageClient.from_config(config) if config.is_at_home else MemoryStorageClient.from_config(config)
```
janbuchar, line 62: This is supposed to happen in `Actor.init`, right? Why duplicate it here?

src/apify/scrapy/middlewares/apify_proxy.py
vdusek: Maybe it is a problem of Apify proxies, I don't know, but it results in the following:

```
[scrapy.downloadermiddlewares.robotstxt] ERROR Error downloading <GET https://console.apify.com/robots.txt>: Could not open CONNECT tunnel with proxy proxy.apify.com:8000 [{'status': 403, 'reason': b'Forbidden'}] ({"spider": "<TitleSpider 'title_spider' at 0x7f2bc3aee660>"})
      Traceback (most recent call last):
        File "/home/vdusek/Projects/apify-sdk-python/.venv/lib/python3.13/site-packages/twisted/internet/defer.py", line 2013, in _inlineCallbacks
          result = context.run(
              cast(Failure, result).throwExceptionIntoGenerator, gen
          )
        File "/home/vdusek/Projects/apify-sdk-python/.venv/lib/python3.13/site-packages/twisted/python/failure.py", line 467, in throwExceptionIntoGenerator
          return g.throw(self.value.with_traceback(self.tb))
                 ~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "/home/vdusek/Projects/apify-sdk-python/.venv/lib/python3.13/site-packages/scrapy/core/downloader/middleware.py", line 68, in process_request
          return (yield download_func(request, spider))
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      scrapy.core.downloader.handlers.http11.TunnelError: Could not open CONNECT tunnel with proxy proxy.apify.com:8000 [{'status': 403, 'reason': b'Forbidden'}]
```

src/apify/scrapy/scheduler.py
vdusek: Because of nested event loop, otherwise, it will result in:
```
RuntimeError: <asyncio.locks.Event object at 0x7c2d640c8fc0 [unset]> is bound to a different event loop
```
when using Apify client.

src/apify/scrapy/middlewares/apify_proxy.py
janbuchar: Humph. But the connect call should happen way before the path part of the URL matters, right?

vdusek: Yep, that's strange. I'm not sure why we can't connect when it comes to robots.txt, while other URLs works. I've reverted the changes and kept only the storage client fix.

janbuchar (approved):

honzajavorek:
Thanks!

PR #389, opened 2025-02-05
feat: allow custom exit function through dependency injection

honzajavorek:
In some contexts it is not desirable to call `sys.exit()` directly from the Actor object. This is already apparent from the surrounding code, which has to deal with such situations: IPython, pytest, or nested loop. I found yet another such situation: Twisted reactor.

I think this leads to a refactoring where the `fail()` and `exit()` functions are not responsible for ending the program, as [explicit is better than implicit](https://en.wikipedia.org/wiki/Zen_of_Python). However, such change would be a larger task I'm not up to, and it would possibly be a breaking change. I'll let the maintainers to discuss and decide the best solution.

My change is minimal and allows an escape route well within the existing API. If we use dependency injection and move `sys.exit` from being hardcoded to being a parameter of both `fail()` and `exit()`, the interface stays the same for existing code, but I'd get the ability to call these methods in contexts, where I don't want them to terminate the program:

```python
from twisted.internet import asyncioreactor, defer
from scrapy.utils.project import get_project_settings
from scrapy.crawler import CrawlerRunner
from scrapy.utils.defer import deferred_from_coro
from twisted.internet.task import react

asyncioreactor.install()

settings = get_project_settings()
runner = CrawlerRunner(settings)

@defer.inlineCallbacks
def crawl(reactor):
    yield deferred_from_coro(Actor.init())
    yield runner.crawl(MySpider)
    yield deferred_from_coro(Actor.exit())
    print("Done!")

react(crawl, [])
```

Currently, the above code won't print `Done!`, because the program ends and there is no straightforward way to prevent it. I invented a workaround like this:

```python
builtins.__IPYTHON__ = True  # deception, Actor.exit() won't call sys.exit()
yield deferred_from_coro(Actor.exit())
```

If this PR is accepted and a new version of SDK released, I could do this:

```python
yield deferred_from_coro(Actor.exit(sys_exit=lambda _: None))
```

In tests, for example, it would be possible to pass a [mock](https://docs.python.org/3/library/unittest.mock.html) and check that the `sys_exit` function got called with a correct exit code, etc.

### DIFF
--- a/src/apify/_actor.py
+++ b/src/apify/_actor.py
@@ -231,7 +231,7 @@ class _ActorType:
         event_listeners_timeout: timedelta | None = EVENT_LISTENERS_TIMEOUT,
         status_message: str | None = None,
         cleanup_timeout: timedelta = timedelta(seconds=30),
-        sys_exit: Callable = sys.exit,
+        sys_exit: Callable[[int], None] = sys.exit,
     ) -> None:
         """Exit the Actor instance.
 
--- a/src/apify/_actor.py
+++ b/src/apify/_actor.py
@@ -231,6 +231,7 @@ class _ActorType:
         event_listeners_timeout: timedelta | None = EVENT_LISTENERS_TIMEOUT,
         status_message: str | None = None,
         cleanup_timeout: timedelta = timedelta(seconds=30),
+        sys_exit: Callable = sys.exit,
     ) -> None:
         """Exit the Actor instance.
 
@@ -238,10 +239,11 @@ class _ActorType:
         sends a final `PERSIST_STATE` event, waits for all the event listeners to finish, and stops the event manager.
 
         Args:
-            exit_code: The exit code with which the Actor should fail (defaults to `0`).
+            exit_code: The exit code with which the program should end (defaults to `0`).
             event_listeners_timeout: How long should the Actor wait for Actor event listeners to finish before exiting.
             status_message: The final status message that the Actor should display.
             cleanup_timeout: How long we should wait for event listeners.
+            sys_exit: A function to use to terminate the program (defaults to `sys.exit`)
         """
         self._raise_if_not_initialized()
 
@@ -267,13 +269,13 @@ class _ActorType:
         self._is_initialized = False
 
         if is_running_in_ipython():
-            self.log.debug(f'Not calling sys.exit({exit_code}) because Actor is running in IPython')
+            self.log.debug(f'Not calling sys_exit({exit_code}) because Actor is running in IPython')
         elif os.getenv('PYTEST_CURRENT_TEST', default=False):  # noqa: PLW1508
-            self.log.debug(f'Not calling sys.exit({exit_code}) because Actor is running in an unit test')
+            self.log.debug(f'Not calling sys_exit({exit_code}) because Actor is running in an unit test')
         elif hasattr(asyncio, '_nest_patched'):
-            self.log.debug(f'Not calling sys.exit({exit_code}) because Actor is running in a nested event loop')
+            self.log.debug(f'Not calling sys_exit({exit_code}) because Actor is running in a nested event loop')
         else:
-            sys.exit(exit_code)
+            sys_exit(exit_code)
 
     async def fail(
         self,
@@ -281,15 +283,17 @@ class _ActorType:
         exit_code: int = 1,
         exception: BaseException | None = None,
         status_message: str | None = None,
+        sys_exit: Callable = sys.exit,
     ) -> None:
         """Fail the Actor instance.
 
         This performs all the same steps as Actor.exit(), but it additionally sets the exit code to `1` (by default).
 
         Args:
-            exit_code: The exit code with which the Actor should fail (defaults to `1`).
+            exit_code: The exit code with which the program should fail (defaults to `1`).
             exception: The exception with which the Actor failed.
             status_message: The final status message that the Actor should display.
+            sys_exit: A function to use to terminate the program (defaults to `sys.exit`)
         """
         self._raise_if_not_initialized()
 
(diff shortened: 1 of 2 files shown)


src/apify/_actor.py
```diff
@@ -231,17 +231,19 @@ async def exit(
         event_listeners_timeout: timedelta | None = EVENT_LISTENERS_TIMEOUT,
         status_message: str | None = None,
         cleanup_timeout: timedelta = timedelta(seconds=30),
+        sys_exit: Callable = sys.exit,
```
janbuchar, line 234: The type should specify the parameter and return value types as well
```suggestion
        sys_exit: Callable[[int], None] = sys.exit,
```

janbuchar, line 234: Can this be passed in when the context manager is used (`async with Actor`)? Otherwise, I'd be more inclined to just not calling `sys.exit` from `Actor.exit`. From my point of view, the proposed solution just lets you do `Actor.exit(sys_exit=something)` instead of `Actor.exit(); something()`.

src/apify/_actor.py
```diff
@@ -267,29 +269,31 @@ async def finalize() -> None:
...
+            self.log.debug(f'Not calling sys_exit({exit_code}) because Actor is running in IPython')
         elif os.getenv('PYTEST_CURRENT_TEST', default=False):  # noqa: PLW1508
-            self.log.debug(f'Not calling sys.exit({exit_code}) because Actor is running in an unit test')
+            self.log.debug(f'Not calling sys_exit({exit_code}) because Actor is running in an unit test')
         elif hasattr(asyncio, '_nest_patched'):
-            self.log.debug(f'Not calling sys.exit({exit_code}) because Actor is running in a nested event loop')
+            self.log.debug(f'Not calling sys_exit({exit_code}) because Actor is running in a nested event loop')
         else:
-            sys.exit(exit_code)
+            sys_exit(exit_code)
```
janbuchar, line 278: I believe this logic could be simplified a fair bit. 

- we don't need to check if we're running under pytest - we can instead pass in a mock `sys_exit`
- didn't @vdusek manage to remove the nesting?
- we probably still have to check for ipython :confused:

vdusek (approved):
I don't see a problem in this, but let's wait for @janbuchar opinion as well.

janbuchar (commented):
Thanks for this initiative! In all honesty, I'm not sure if we should call `sys.exit` at all, but this seems like a nice, non-breaking change. But there's stuff to discuss first.

src/apify/_actor.py
vdusek: Regarding the `nest-asyncio` ...

We managed to resolve the problem of executing the Actor & Spider together. Here is more context and the core result structure - https://github.com/scrapy/scrapy/discussions/6639#discussioncomment-12019270.

However, there's still one remaining challenge - executing asyncio coroutines from the synchronous Scheduler (and HTTP cache storage) code. This is in-progress. 

Once that's addressed, I'll open a new PR removing `nest-asyncio` completely.

honzajavorek:
Just for a context, I agree the library probably shouldn't call `sys.exit()` at all.

But I'm just a passing person who happened to need this in order to contribute to solving the nested loops 😄 I'm already deep in the mud over there. I can't work on two mud pools in parallel. Hence I've made this minimal change, with pincers and gloves, and I send it to you as a gift 💝 

If I was a maintainer of the code, I'd take a bulldozer and yeah, sure, after a week of work and a few breaking changes, there would be no `sys.exit()` left and no conditions checking on pytest etc. 😅 🚜

janbuchar:
> But I'm just a passing person who happened to need this in order to contribute to solving the nested loops 😄 I'm already deep in the mud over there. I can't work on two mud pools in parallel. Hence I've made this minimal change, with pincers and gloves, and I send it to you as a gift 💝

And we will forever be grateful for that. If you don't have the resources to work on this further, we can totally take over and polish it, but we value any input that you can provide.

> If I was a maintainer of the code, I'd take a bulldozer and yeah, sure, after a week of work and a few breaking changes, there would be no `sys.exit()` left and no conditions checking on pytest etc. 😅 🚜

I am a maintainer and I'm still not sure we should remove it completely, so let's try and come up with a nice incremental upgrade for now.

honzajavorek:
Mr. @vdusek, I see [what you did there](https://github.com/apify/apify-sdk-python/compare/master...fixing-scrapy#diff-b5eb8a1a9d4ae05ce1f06ff99335f1aeaa03d121890539fc4d22855401b871b9R38)! 😂

Screenshot 2025-02-05 at 16 34 53

If you're okay to institutionalize my use case like this, I'm happy to close this 😄 

----

Regarding about what's right and wrong... my hunch would be that the methods do too many things. They finalize and close the actor's resources, and they take care of ending the program at the same time. In some contexts, it's handy that I don't have to deal with the exit code myself, but in other contexts, it's a problem that I cannot opt out. So I think those should be separate steps:

- Actor.finalize() without exit_code, but with all the rest
- Actor.exit() which sets the exit code and calls sys.exit
- Actor.__init__() takes a parameter whether it should end the program or not, saves it as a property
- Actor.__aexit__() reads the property and either ends the program or not
- I'm not sure whether Actor.fail() has to exist at all?

Hmm, and do we need ending the program at all? If Python program raises, it ends with non-zero exit code by default. Do we need to handle the program termination, or do we leave it up to the user of the SDK? I don't see a good reason to handle it at all, tbh. I struggle to see the added value, but I don't have a scraping platform and can't read user support and don't know about all the use cases. To me, ending the program as it is now feels like "we'll do this as a convenience to the user", but given the number of ifs for the exceptions and the fact that raising an exception results in the same thing, I feel like in the end it's more of an inconvenience, which needs workarounds.

vdusek (commented):
Taking into account the JS version [actor.ts#L1978-L1981](https://github.com/apify/apify-sdk-js/blob/apify%403.3.0/packages/apify/src/actor.ts#L1978-L1981), I would say we should align the Py-SDK behavior with it. Which means add a flag for turning off the calling of `sys.exit`, rather than adding an custom exit function.

honzajavorek:
Closed by https://github.com/apify/apify-sdk-python/pull/424

PR #393, opened 2025-02-06
feat: Support pay-per-event via `Actor.charge`

### LINKED ISSUE #374: Implement `Actor.charge`
See https://github.com/apify/apify-sdk-js/pull/346 for inspiration

janbuchar:
- closes #374
- documentation will be added after https://github.com/apify/apify-sdk-js/pull/356 is finished
- the functionality should be used mainly via `Actor.charge`
- directly accessing `ChargingManager` is needed only for advanced use cases
- see [integration tests](https://github.com/apify/apify-sdk-python/pull/393/files#diff-c209584a944d571ec9bd7d62563827b91c80eed00e4c91170ea9e500d39c0477) for an illustration of how this works


vdusek (commented):
Could you please explain it somewhere in the concepts docs?

Starting with: https://docs.apify.com/sdk/python/docs/concepts/actor-lifecycle

janbuchar:
> Could you please explain it somewhere in the concepts docs?
> 
> Starting with: https://docs.apify.com/sdk/python/docs/concepts/actor-lifecycle

There are docs being written in https://github.com/apify/apify-sdk-js/pull/356 - would you like to check it out so that I can mirror it here afterwards?

Are you sure that the Actor lifecycle page is a good place for this? If monetization belongs there, why not dataset handling, for example?

vdusek:
> There are docs being written in https://github.com/apify/apify-sdk-js/pull/356 - would you like to check it out so that I can mirror it here afterwards?

Cool, I'll check it out, thanks.

> Are you sure that the Actor lifecycle page is a good place for this? If monetization belongs there, why not dataset handling, for example?

Sorry for the confusion. I just linked the first page from the concepts. Unfortunately, the https://docs.apify.com/sdk/python/docs/concepts/ leads to 404. Choose the right location by yourself.

src/apify/_charging.py
```diff
@@ -0,0 +1,241 @@
...
+                )
+
+        if not self._is_at_home and self._pricing_model == 'PAY_PER_EVENT':
+            if self._purge_charging_log_dataset:
+                dataset = await Dataset.open(name=self.LOCAL_CHARGING_LOG_DATASET_NAME)
+                await dataset.drop()
+
+            self._charging_log_dataset = await Dataset.open(name=self.LOCAL_CHARGING_LOG_DATASET_NAME)
+
+    async def charge(self, event_name: str, count: int = 1) -> ChargeResult:
```
Pijukatel, line 88: This is quite a long method, docstring with some description would be nice.

src/apify/_charging.py
```diff
@@ -0,0 +1,241 @@
...
+        if self._charging_state is None:
```
Pijukatel, line 89: This guard is repeated in several methods. Maybe it could be turned into decorator, something like
`@ensure_initialized`. I think that would improve readability as one could focus on the unique part of the method.

(See similar usecase [@ensure_context](https://github.com/apify/crawlee-python/blob/master/src/crawlee/browsers/_browser_pool.py#L228 )

src/apify/_charging.py
```diff
@@ -0,0 +1,241 @@
...
+            raise RuntimeError('Charging manager is not initialized')
+
+        def calculate_chargeable() -> dict[str, int | None]:
+            return {
+                event_name: self.calculate_max_event_charge_count_within_limit(event_name)
+                for event_name in self._pricing_info
+            }
+
+        if self._pricing_model != 'PAY_PER_EVENT':
```
Pijukatel, line 98: What is the point of this branch?

If the actor does not support 'PAY_PER_EVENT', shouldn't this raise an exception instead? In which scenario we want to just log it?

tests/integration/test_actor_charge.py
```diff
@@ -0,0 +1,95 @@
...
+
+    await actor_client.update(
+        pricing_infos=[
+            {
+                'pricingModel': 'PAY_PER_EVENT',
+                'pricingPerEvent': {
+                    'actorChargeEvents': {
+                        'foobar': {
+                            'eventTitle': 'Foo bar',
+                            'eventPriceUsd': 0.1,
```
Pijukatel, line 43: I saw just indirect use of `eventPriceUsd` in the tests through the charge limit. Should we maybe check some total charged sum. For example in test_actor_charge_basic it should be 0.4 Usd?

Seems to me that the total sum is kind of the most relevant output.

tests/integration/test_actor_charge.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+async def test_actor_charge_basic(
+    ppe_actor: ActorClientAsync,
+    run_actor: RunActorFunction,
+    apify_client_async: ApifyClientAsync,
+) -> None:
+    run = await run_actor(ppe_actor)
+
+    # Wait until the platform gets its act together and refetch
```
Pijukatel, line 73: Is there any guarantee that the actor will finish within 6 seconds?
Maybe it would be better with more event based approach instead of flat sleep? For example some periodic polling of actor.run until run.status signals that we can do the asserts. All wrapped in some deadline timeout.

src/apify/_charging.py
janbuchar: You might want to change the pricing model of your Actor at some point. If you do that, you need to support both of them in your code, at least for a limited transition period. Crashing here would make achieving that harder.

tests/integration/test_actor_charge.py
janbuchar: According to platform people, we can be 99% sure after 6 seconds. I guess we could instead retry the check every .5 seconds for a longer time, but this seems realiable enough... so far.

tests/integration/test_actor_charge.py
janbuchar: Unfortunately, this information cannot be found in the run response - this is probably the closest we can get.

src/apify/_charging.py
```diff
@@ -0,0 +1,277 @@
...
+@dataclass(frozen=True)
+class ChargeResult:
+    event_charge_limit_reached: bool
+    """If true, no more events of this type can be charged within the limit"""
+
+    charged_count: int
+    """Total amount of charged events - may be lower than the requested amount"""
+
+    chargeable_within_limit: dict[str, int | None]
+    """How many events of each known type can still be charged within the limit"""
```
vdusek, line 249: Missing class docstring and dots at the end of the sentences.

vdusek (commented):
a few comments... (more will come tomorrow)

src/apify/_actor.py
```diff
@@ -445,19 +452,45 @@ async def open_request_queue(
...
+    @overload
+    async def push_data(self, data: dict | list[dict]) -> None: ...
+    @overload
+    async def push_data(self, data: dict | list[dict], event_name: str) -> ChargeResult: ...
+    async def push_data(self, data: dict | list[dict], event_name: str | None = None) -> ChargeResult | None:
         """Store an object or a list of objects to the default dataset of the current Actor run.
 
         Args:
             data: The data to push to the default dataset.
+            event_name: If provided, the method will attempt to charge for the event for each pushed item.
```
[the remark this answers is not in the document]
janbuchar: Brother, I'm but the last link in a long chain of subtasks that make up the PPE initiative. At least this is consistent with the Actor whitepaper and with the API conventions.

tests/integration/test_actor_charge.py
Pijukatel: 99% still means, that the test is 1% flaky. That is reliable in the scope of few test runs, but it gradually compounds in CI with other flaky tests to more frequent CI failures. I think your alternative implementation suggestion is worth it.

src/apify/_charging.py
Pijukatel: I am having hard time imagining this scenario. What is exactly the transition period in this case? Don't you change the actor pricing model the at the same time when you change the code? The actors that were started  before transition (and still running) are using old pricing model and old code and actors starting after the change are using new pricing model and new code.

src/apify/_actor.py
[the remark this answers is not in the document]
vdusek: Oh, of course... I was thinking more about clarifying this in the docstrings so users don't confuse this event with "system" events (and the event manager).  

Also, maybe we could consider renaming the system events to `system_events` in the future? (open issue for that in this case)

tests/integration/test_actor_charge.py
```diff
@@ -0,0 +1,95 @@
...
+    ppe_actor: ActorClientAsync,
+    run_actor: RunActorFunction,
+    apify_client_async: ApifyClientAsync,
+) -> None:
+    run = await run_actor(ppe_actor, max_total_charge_usd=Decimal('0.2'))
+
+    # Wait until the platform gets its act together and refetch
+    await asyncio.sleep(6)
+    updated_run = await apify_client_async.run(run.id).get()
+    run = ActorRun.model_validate(updated_run)
```
vdusek, line 92: same as here https://github.com/apify/apify-sdk-python/pull/393/files/46bef0de4258c42d2f66be3c9d23121ffeb3d85c#r1952294158

src/apify/_charging.py
janbuchar: But using `@ensure_context` would not tell the type checker that `self._charging_state is not None`, right? So I would have to keep the checks anyway, or add some casts. I guess I could make a specialized decorator that would ensure those types, which could be fun and useful in the future, but I doubt that this is the right time and place.

Or I could add a private `get_charging_state_or_crash` method :shrug:

src/apify/_charging.py
janbuchar: When you want to change the pricing model of a public Actor, the change won't happen immediately - there's a two week (I believe) transition period that is out of control of the Creator.

So if you don't want to sit by the console until the transition period ends and then push the updated code, you need to make sure that your code supports both models.

tests/integration/test_actor_charge.py
janbuchar: @Pijukatel I rewrote the checks, please resolve this if it's OK.

src/apify/_charging.py
```diff
@@ -0,0 +1,277 @@
...
+        self._client = client
+        self._charging_log_dataset: Dataset | None = None
+
+        self._charging_state: dict[str, ChargingStateItem] | None = None
+        self._pricing_info: dict[str, PricingInfoItem] = {}
+
+        self._not_ppe_warning_printed = False
+
+    async def init(self) -> None:
+        """Initialize the charging manager - this is called by the `Actor` class and shouldn't be invoked manually."""
```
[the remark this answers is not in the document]
janbuchar: I have no problem making it a context manager, but that hardly resolves the issue - anyone can attempt `async with ChargingManager` in the same way they would try to do `ChargingManager.init()`.

The whole reason for making `ChargingManager` public is that it is the return type of `Actor.get_charging_manager`. Perhaps we could make a `Protocol` and use that for the return type annotation?

src/apify/_actor.py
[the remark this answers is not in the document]
janbuchar: Okay, I tried to clarify the docstring, let me know what you think

tests/integration/test_actor_charge.py
```diff
@@ -63,20 +65,33 @@ async def ppe_actor(
     return apify_client_async.actor(ppe_actor_build)
 
 
+def retry_counter(retry_count: int) -> Iterable[tuple[bool, int]]:
+    for retry in range(retry_count - 1):
+        yield False, retry
```
Pijukatel, line 70: Is there any point in returning the `retry`? I saw it is always saved into `_`

tests/integration/test_actor_charge.py
```diff
@@ -63,20 +65,33 @@ async def ppe_actor(
     return apify_client_async.actor(ppe_actor_build)
 
 
+def retry_counter(retry_count: int) -> Iterable[tuple[bool, int]]:
```
Pijukatel, line 68: Tiny detail, but seems to me, it is offseted by one. For example:  
`retry_counter(retry_count=10)` will do 1 x initial attempt and 9 x retry. 
So I guess the argument should be called something like `total_attempts` or -1 offset removed
`retry_counter(total_attempts=10)`

src/apify/_actor.py
```diff
@@ -513,6 +547,19 @@ async def set_value(
...
+        self._raise_if_not_initialized()
+        return self._charging_manager
+
+    async def charge(self, event_name: str, count: int = 1) -> ChargeResult:
+        """Charge for a specified number of events - sub-operations of the Actor.
+
+        This is relevant only for the pay-per-event pricing model.
+        """
+        self._raise_if_not_initialized()
+        return await self._charging_manager.charge(event_name, count)
```
vdusek, line 561: missing arg docstring

src/apify/_charging.py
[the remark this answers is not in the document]
vdusek: > The whole reason for making ChargingManager public is that it is the return type of Actor.get_charging_manager. Perhaps we could make a Protocol and use that for the return type annotation?

I would give it a try.

src/apify/_actor.py
janbuchar: True, but what useful information can I provide there?

src/apify/_charging.py
```diff
@@ -0,0 +1,286 @@
...
+                dataset = await Dataset.open(name=self.LOCAL_CHARGING_LOG_DATASET_NAME)
+                await dataset.drop()
+
+            self._charging_log_dataset = await Dataset.open(name=self.LOCAL_CHARGING_LOG_DATASET_NAME)
+
+    async def charge(self, event_name: str, count: int = 1) -> ChargeResult:
+        """Charge for a specified number of events - sub-operations of the Actor.
+
+        This is relevant only for the pay-per-event pricing model.
+        """
```
vdusek, line 100: missing arg docstring, especially if we have 2 events with different meanings now :smile:

vdusek (commented):
I would really add a clarifying note to the new events docstring, such as: "Note: Do not confuse these events with system events used by event manager blah blah". Because I am sure someone will get confused in the future since 2 different things are named the same.

src/apify/_actor.py
vdusek: 1) it will be rendered as an empty description field otherwise
2) https://github.com/apify/apify-sdk-python/pull/393#pullrequestreview-2626102302

src/apify/_actor.py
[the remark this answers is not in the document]
janbuchar: Yes. In the `push_data` overload, it makes sense to make the parameter name more specific - it is not obvious why it would accept an event name.

src/apify/_charging.py
Pijukatel: You could also turn it into property that never returns `None` and do the `None` check there. The downside is that private property is somewhat strange.

vdusek (approved):
LGTM

jirispilka:
Do you think we can close it today? 🙏🏻 I'd like to have it ready for the Agents Hackathon.

src/apify/_charging.py
janbuchar: Good point, will adjust.

Pijukatel (approved):

src/apify/_models.py
```diff
@@ -115,3 +120,55 @@ class ActorRun(BaseModel):
...
+    pricing_per_event: Annotated[PricingPerEvent, Field(alias='pricingPerEvent')]
+    minimal_max_total_charge_usd: Annotated[Decimal | None, Field(alias='minimalMaxTotalChargeUsd')] = None
+
+
+PricingModel: TypeAlias = Literal[
+    'FREE',
+    'FLAT_PRICE_PER_MONTH',
+    'PRICE_PER_DATASET_ITEM',
+    'PAY_PER_EVENT',
+]
```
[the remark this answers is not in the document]
barjin: I see, so you cannot use `PricingModel.FREE` as a runtime value, I got confused by the assignment (`sth = Literal[]`) syntax. I'll see what I can do regarding the expansion of the type alias to its "value" in the call site doc.

PR #394, opened 2025-02-07
chore(ci): unblock stalling documentation build

barjin:
Bumps the available heap size for the doc-building Docusaurus process. We already do this e.g. [in Crawlee JS](https://github.com/apify/crawlee/blob/7f09d56ed5776d92a11feee3a5e61255301d0099/website/package.json#L7).


barjin:
Good catch! The `max-old-space-size` flag apparently doesn't work at all. Weirdly, I didn't manage to reproduce this locally. Let's see how the latest Docusaurus performs with `experimental_faster`

website/src/pages/index.module.css
```diff
@@ -176,10 +176,6 @@ html[data-theme='dark'] .heroBanner h2 {
     }
 }
 
-.tagline span {
-    color: var(--ifm-color-primary);
-}
```
barjin, line 181: The new `rspack` bundler (or adjacent tooling) seems to wrongly resolve `!important` CSS rules (the tagline had blue words instead of transparent). Removing this rule seems to fix the issue (and not break anything else.)

B4nan:
Its already set up on the main docs repo, we need to keep the swc loader disabled since that doesn't work well (or should I say at all) with styled components. 

https://github.com/apify/apify-docs/blob/master/docusaurus.config.js#L22

Although it might be fine in this repo, the place where we use it is in the landing pages like http://docs.apify.com/api

barjin:
Huh, it seems that `rspack` doesn't unblock the stalling CI builds either 🙃 I'll try reproducing it locally a bit more and see what I can come up with.

barjin:
Closing as solved by https://github.com/apify/apify-docs/pull/1452

B4nan:
I would still merge this one with the docusaurus update, or is there any reason not to?

barjin:
Oh yeah, that makes sense. Let me regenerate the lockfile anew (there seem to be quite a lot of unrelated changes as a result of the debugging) and we can switch to `3.7.0`.

PR #398, opened 2025-02-11
docs: Set line length to docs related code to 90

Pijukatel:
Description

Set line length to docs related code to 90 to have each code example fully visible without the need to use slider.
Update existing examples to be compliant.


vdusek:
we can do this now

vdusek (approved):
LGTM

PR #400, opened 2025-02-13
chore: utilize `ensure_context` decorator for Actor

vdusek:
Utilize `ensure_context` decorator from Crawlee rather than calling `raise_if_...` method in the Actor class.

### DIFF
--- a/src/apify/_actor.py
+++ b/src/apify/_actor.py
@@ -14,6 +14,7 @@ from apify_client import ApifyClientAsync
 from apify_shared.consts import ActorEnvVars, ActorExitCodes, ApifyEnvVars
 from apify_shared.utils import ignore_docs, maybe_extract_enum_member_value
 from crawlee import service_locator
+from crawlee._utils.context import ensure_context
 from crawlee.events import (
     Event,
     EventAbortingData,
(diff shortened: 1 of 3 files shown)


janbuchar (commented):
I'm honestly not sure about this. It reduces some duplication, yes, but the error message will be pretty cryptic. It is fine for an internal helper in Crawlee, but here, it's very much user-facing. Maybe we could make a wrapper for the decorator?

janbuchar:
> Is it that bad? We can wrap it of course.
Well, leaking the existence of `_ActorType` is confusing by itself. Plus, the message should tell the user "Just use `Actor.init` or `async with Actor`" - many of them won't even know what a context manager is.

Pijukatel (commented):
I like this decorator, maybe we have to have custom error message in this case, but I think it makes the code more readable.

vdusek:
Yep, I will try to update the decorator to have a better error message there.

Pijukatel (approved):

janbuchar (changes requested):
Please wrap that error message :slightly_smiling_face:

vdusek:
Clousing, since we moved the flag(s) to the class level, and I'm afraid this means we can't utilize `ensure_context` now, as it relies on an instance-level attribute.

PR #402, opened 2025-02-13
fix: Make sure that Actor instances with non-default configurations are also accessible through the global Actor proxy after initialization

### LINKED ISSUE #397: Constructor `_ActorType.__init__` is being executed twice
## Description

The `_ActorType.__init__` constructor is executed twice when creating an instance of Actor.

## Reproduction

I added simple prints to the `_ActorType.__init__`, `_ActorType.__aenter__` and `_ActorType.init` methods and prepared a simple Actor:

```python
from apify import Actor

async def main() -> None:
    async with Actor() as actor:
        actor.log.info(f'Hello! Actor={Actor}, actor={actor}.')
```

Execution:

```
$ apify run
Run: /home/vdusek/Projects/apify-sdk-python/.venv/bin/python3 -m src
Hello from _ActorType.__init__
Hello from _ActorType.__init__
Hello from _ActorType.__aenter__
Hello from _ActorType.init
[apify] INFO  Initializing Actor...
[apify] INFO  System info ({"apify_sdk_version": "2.2.2", "apify_client_version": "1.8.1", "crawlee_version": "0.5.3", "python_version": "3.13.2", "os": "linux"})
[apify] INFO  Hello! Actor=<apify.Actor>, actor=<apify._actor._ActorType object at 0x7ff7f757fed0>.
[apify] INFO  Exiting Actor ({"exit_code": 0})
```

This does not happen when using the Actor class without an instance, as shown below:

```python
from apify import Actor

async def main() -> None:
    async with Actor:
        Actor.log.info(f'Hello! Actor={Actor}.')
```

Output

```
$ apify run
Run: /home/vdusek/Projects/apify-sdk-python/.venv/bin/python3 -m src
Hello from _ActorType.__init__
Hello from _ActorType.__aenter__
Hello from _ActorType.init
[apify] INFO  Initializing Actor...
[apify] INFO  System info ({"apify_sdk_version": "2.2.2", "apify_client_version": "1.8.1", "crawlee_version": "0.5.3", "python_version": "3.13.2", "os": "linux"})
[apify] INFO  Hello! Actor=<apify.Actor>.
[apify] INFO  Exiting Actor ({"exit_code": 0})
```

janbuchar:
- closes #397

### DIFF
--- a/src/apify/_actor.py
+++ b/src/apify/_actor.py
@@ -55,9 +55,6 @@ MainReturnType = TypeVar('MainReturnType')
 class _ActorType:
     """The class of `Actor`. Only make a new instance if you're absolutely sure you need to."""
 
-    _apify_client: ApifyClientAsync
-    _configuration: Configuration
-    _is_exiting = False
     _is_rebooting = False
 
     def __init__(
@@ -76,6 +73,8 @@ class _ActorType:
                 be created.
             configure_logging: Should the default logging configuration be configured?
         """
+        self._is_exiting = False
+
         self._configuration = configuration or Configuration.get_global_configuration()
         self._configure_logging = configure_logging
         self._apify_client = self.new_client()
--- a/src/apify/_actor.py
+++ b/src/apify/_actor.py
@@ -900,11 +900,11 @@ class _ActorType:
             self.log.error('Actor.reboot() is only supported when running on the Apify platform.')
             return
 
-        if self._is_rebooting:
+        if _ActorType._is_rebooting:
             self.log.debug('Actor is already rebooting, skipping the additional reboot call.')
             return
 
-        self._is_rebooting = True
+        _ActorType._is_rebooting = True
 
         if not custom_after_sleep:
             custom_after_sleep = self._configuration.metamorph_after_sleep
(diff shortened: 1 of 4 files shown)


src/apify/_actor.py
```diff
@@ -55,9 +55,6 @@
 class _ActorType:
     """The class of `Actor`. Only make a new instance if you're absolutely sure you need to."""
 
-    _apify_client: ApifyClientAsync
-    _configuration: Configuration
-    _is_exiting = False
     _is_rebooting = False
```
vdusek, line 58: Why `is_rebooting` remain the class attribute and `is_exiting` not?

janbuchar: Because when you call `reboot` on platform, it is an event that affects any possible `_ActorType` instances. `_is_exiting` can be local to the current instance, especially if we let users disable calling `sys.exit` from there.

vdusek: Thanks

vdusek (commented):
Ok, I tested it out, now the `Actor()` works for Scrapy Actors, which is great.

But the `Actor.__init__` is still being called twice, am I correct?

janbuchar:
> Ok, I tested it out, now the `Actor()` works for Scrapy Actors, which is great.

:tada: 

> But the `Actor.__init__` is still being called twice, am I correct?

That much is correct. Avoiding that would be tricky, though I'm sure it's doable. Yet, I was under the impression that it was not really a problem.

vdusek (approved):
Thanks, LGTM in that case

PR #406, opened 2025-02-18
chore: migrate to uv package manager

vdusek:
- Migrate from `poetry` to `uv`.
- Relates: https://github.com/apify/crawlee-python/issues/628


pyproject.toml
```diff
@@ -32,49 +37,57 @@ keywords = [
...
-pytest-only = "~2.1.0"
-pytest-timeout = "~2.3.0"
-pytest-xdist = "~3.6.0"
-respx = "~0.22.0"
-ruff = "~0.9.0"
-setuptools = "~75.8.0"     # setuptools are used by pytest but not explicitly required
-
-[tool.poetry.extras]
-scrapy = ["scrapy"]
+#######################
```
Pijukatel, line 81: These `XXX configuration` highlights are somewhat redundant as the `[tool.XXX]` already specifies section relevant for each tool.
(Same for others)

vdusek: yep, but the tools can have multiple such sections (see Ruff), however, if we don't have a consensus here, I'll remove it

Pijukatel: Yes, but those multiple sections can be still misplaced even with this extra highlighted text. It is visual help, but does not enforce anything. But I do not have strong opinion against it either. Approve in either case.

Mantisus: But adding tests, that's always a great idea)

Mantisus (approved):
LGTM

PR #413, opened 2025-02-24
fix: Allow None value in 'inputBodyLen' in ActorRunStats

janbuchar:


jirimoravcik (approved):
Thank you!

vdusek (approved):

janbuchar:
Relevant bug report https://apify.slack.com/archives/C0L33UM7Z/p1740388155419149

src/apify/_models.py
```diff
@@ -46,7 +46,7 @@ class ActorRunMeta(BaseModel):
 class ActorRunStats(BaseModel):
     __model_config__ = ConfigDict(populate_by_name=True)
 
-    input_body_len: Annotated[int, Field(alias='inputBodyLen')]
+    input_body_len: Annotated[int | None, Field(alias='inputBodyLen')] = None
```
Pijukatel, line 49: Seems it is not used, we might just delete it since we do not use `forbid` settings for extra arguments:
https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.extra

janbuchar: Not so fast cowboy, it's part of our public interface (e.g., `Actor.call` returns `ActorRun`) and library clients might use it (it is dubitable, but still)

PR #420, opened 2025-03-03
feat: update to Crawlee v0.6

vdusek:
I use "feat", I believe it is correct to release this as minor version (not patch).


Pijukatel (commented):
I guess you will need to either add

```
    ignored_env_vars = {
        ApifyEnvVars.XVFB,
        ApifyEnvVars.CHROME_EXECUTABLE_PATH
...
```

to `test_get_env_with_randomized_env_vars`

or remove those variables from apify-shared (probably not)

vdusek:
I removed them from the `apify-shared` and released a new version (it's used only here and in client).

Pijukatel (approved):

PR #427, opened 2025-03-10
chore: Enhanced missing local storage error

### LINKED ISSUE #412: `Actor.open_dataset` with `getData()` method produces an `RuntimeError`
This works:
```
await Actor.apify_client.dataset("dataset-id").list_items(limit=1)
```

This doesn't:
```
dataset = await Actor.open_dataset(id="dataset-id")
data = await dataset.get_data()
```

Fails with:
```
RuntimeError: Dataset with id "dataset-id" does not exist!
```

According to the [documentation](https://docs.apify.com/sdk/python/reference/class/Actor#open_dataset), opening the dataset via ID should be supported.

Pijukatel:
Enhanced missing local storage error

Closes: #412

### DIFF
--- a/src/apify/_actor.py
+++ b/src/apify/_actor.py
@@ -65,7 +65,9 @@ def _add_local_storage_error_hint(function: TFun) -> TFun:
             return await function(self=self, id=id, name=name, force_cloud=force_cloud)
         except Exception as e:
             if not force_cloud:
-                e.args = (f'{e.args[0]} (If you are trying to retrieve a remote storage, use `force_cloud=True`.)',)
+                e.args = (
+                    f'{e.args[0]} (If you are trying to retrieve a remote storage, use `force_cloud=True` argument.)',
+                )
             raise
 
     return cast(TFun, wrapper)
--- a/src/apify/_actor.py
+++ b/src/apify/_actor.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import asyncio
+import functools
 import os
 import sys
 from datetime import timedelta
@@ -49,6 +50,24 @@ if TYPE_CHECKING:
 
 
 MainReturnType = TypeVar('MainReturnType')
+TFun = TypeVar('TFun', bound=Callable[..., Any])
+
+
+def _add_local_storage_error_hint(function: TFun) -> TFun:
+    """This decorator adds a local storage error hint in situation where storage was not found locally."""
+
+    @functools.wraps(function)
+    async def wrapper(
+        self: _ActorType, *, id: str | None = None, name: str | None = None, force_cloud: bool = False
+    ) -> Any:
+        try:
+            return await function(self=self, id=id, name=name, force_cloud=force_cloud)
+        except Exception as e:
+            if not force_cloud:
+                e.args = (f'{e.args[0]} (If you are trying to retrieve a remote storage, use `force_cloud=True`.)',)
+            raise
+
+    return cast(TFun, wrapper)
 
 
 @docs_name('Actor')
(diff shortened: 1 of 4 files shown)


src/apify/_actor.py
```diff
@@ -50,6 +51,26 @@
...
+def _add_local_storage_error_hint(function: TFun) -> TFun:
+    """This decorator adds a local storage error hint in situation where storage was not found locally."""
+
+    @functools.wraps(function)
+    async def wrapper(
+        self: _ActorType, *, id: str | None = None, name: str | None = None, force_cloud: bool = False
+    ) -> Any:
+        try:
+            return await function(self=self, id=id, name=name, force_cloud=force_cloud)
+        except Exception as e:
```
vdusek, line 66: Do we really want to catch all exceptions?

src/apify/_actor.py
```diff
@@ -50,6 +51,26 @@
...
+            if not force_cloud:
+                e.args = (
+                    f'{e.args[0]} (If you are trying to retrieve a remote storage, use `force_cloud=True` argument.)',
+                )
```
vdusek, line 70: Couldn't `e.args` contain more than one item? If so, am I correct that we are discarding them?

src/apify/_actor.py
```diff
@@ -50,6 +51,26 @@
 
 
 MainReturnType = TypeVar('MainReturnType')
+TFun = TypeVar('TFun', bound=Callable[..., Any])
+
+
+def _add_local_storage_error_hint(function: TFun) -> TFun:
```
vdusek, line 57: Couldn't we move the decorator to a dedicated private module, perhaps [apify._utils](https://github.com/apify/apify-sdk-python/blob/master/src/apify/_utils.py)?

src/apify/_actor.py
```diff
@@ -50,6 +51,26 @@
...
+        try:
+            return await function(self=self, id=id, name=name, force_cloud=force_cloud)
+        except Exception as e:
+            if not force_cloud:
+                e.args = (
+                    f'{e.args[0]} (If you are trying to retrieve a remote storage, use `force_cloud=True` argument.)',
+                )
+            raise
+
+    return cast(TFun, wrapper)
```
vdusek, line 73: Do we really need the cast? Doesn't `functools.wraps` already take care of preserving the function's type?

vdusek (commented):
I have a few questions about the decorator.

I'm also wondering if we're not using a sledgehammer to crack a nut.

From the issue description:

> Yeah, it was reported by me.
>
> I did some tests and to me it seems to be working as expected.
>
> Cool 👍 As discussed in the Slack thread, there is no actual bug, it's just that mixing remote and local development makes things really confusing. So updating the documentation makes sense.
>
> On top of that, how about updating the error message? If:
> 
> I access storage by ID, and
> I'm in local environment, and
> the storage does not exist, then
> show an error mentioning that the issue might be that the storage is remote?

So this should be really just about updating the error message. I would rather do it directly rather than introduce this decorator-obscuration magic.

It seems like this should be only about updating the error message. Maybe I would rather do it directly.

Pijukatel:
> So this should be really just about updating the error message. I would rather do it directly rather than introduce this decorator-obscuration magic.
> 
> It seems like this should be only about updating the error message. Maybe I would rather do it directly.

I am not sure this is even necessary. I raised the question [in the slack](https://apify.slack.com/archives/C06ENBV1B9Q/p1740499187582409), but no one responded there, so this is just the same question more aloud :-)

This error originates from `crawlee` and that is not aware of the `force_cloud` argument which is added by sdk. So we can't have this extra message being added at the place where the exception is originally raised.

I am not saying that inside that decorator it is nice, but I think in general it is correct approach to detach such additional error message enhancement (If needed at all?!) to decorator so that it is not polluting the main function logic.

PR #447, opened 2025-03-26
fix: Fix calculation of CPU utilization from SystemInfo events

janbuchar:
Apify reports CPU utilization as the sum of utilization of all CPUs, but Crawlee expects a number between 0 and 1. Because of this, it is impossible for AutoscaledPool to scale beyond one CPU.

### DIFF
--- a/src/apify/_platform_event_manager.py
+++ b/src/apify/_platform_event_manager.py
@@ -48,11 +48,11 @@ class SystemInfoEventData(BaseModel):
     is_cpu_overloaded: Annotated[bool, Field(alias='isCpuOverloaded')]
     created_at: Annotated[datetime, Field(alias='createdAt')]
 
-    def to_crawlee_format(self) -> EventSystemInfoData:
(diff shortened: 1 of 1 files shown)


vdusek (approved):
LGTM

Pijukatel (commented):
Just something to think about: Is this scaling based on multiple cpus load working well in python context?

I think that there we are mostly running on one core all the time, not sure if there are some load relevant sub-processes. So scaling based on other cores being not utilized does not seem very compatible with current somewhat single core crawlee architecture.

janbuchar:
> Just something to think about: Is this scaling based on multiple cpus load working well in python context?

Well, about the same as in Javascript context really :shrug: 

> I think that there we are mostly running on one core all the time, not sure if there are some load relevant sub-processes. So scaling based on other cores being not utilized does not seem very compatible with current somewhat single process crawlee architecture.

Playwright is the most relevant one, I suppose. For HTTP-based crawlers, I agree that we probably won't be able to utilize more than one core. But also, most people probably won't try to run those on thicker units than your standard 4GB one.

Issue #469, opened 2025-05-10
Introduce new Apify storage client


vdusek:
Introduce a new Apify storage client based on the new interface in Crawlee.

PR #470, opened 2025-05-10
refactor!: Adapt to the Crawlee v1.0

vdusek:
### Description

- Integration of the Crawlee v1 changes, mostly new storages & storage clients (introduced in https://github.com/apify/crawlee-python/pull/1194).

### Issues

- Closes: #469
- Closes: #540 

### Testing

- The current test set covers the changes.


src/apify/apify_storage_client/_dataset_client.py
```diff
@@ -1,190 +1,183 @@
...
-                    view=view,
-                )
-            )
+        id: str | None,
+        name: str | None,
+        configuration: Configuration,
+    ) -> ApifyDatasetClient:
+        token = configuration.token
+        api_url = configuration.api_base_url
+
```
janbuchar, line 73: I believe this should use `configuration.default_dataset_id` if both `id` and `name` are `None`.

src/apify/apify_storage_client/__init__.py
janbuchar, line 1: What's the point of renaming this module to `apify.storage_clients`? I don't think there is a realistic chance that we'll add another storage client implementation to the SDK.

docs/03_concepts/code/03_dataset_exports.py
```diff
@@ -11,14 +11,14 @@ async def main() -> None:
         await dataset.export_to(
             content_type='csv',
             key='data.csv',
-            to_key_value_store_name='my-cool-key-value-store',
+            to_kvs_name='my-cool-key-value-store',
```
janbuchar, line 14: Is this BC break worth it?

src/apify/storage_clients/_apify/_dataset_client.py
```diff
@@ -0,0 +1,214 @@
...
+            created_at=created_at,
+            accessed_at=accessed_at,
+            modified_at=modified_at,
+            item_count=item_count,
+        )
+
+        self._api_client = api_client
+        """The Apify dataset client for API operations."""
+
+        self._lock = asyncio.Lock()
```
janbuchar: I've seen `asyncio.Lock` instantiation outside of an `async` function cause event loop mismatch problems. I'll try to find evidence later :grin:

src/apify/storage_clients/_apify/_dataset_client.py
```diff
@@ -0,0 +1,214 @@
...
+            fields=fields,
+            omit=omit,
+            unwind=unwind,
+            skip_empty=skip_empty,
+            skip_hidden=skip_hidden,
+            flatten=flatten,
+            view=view,
+        )
+        result = DatasetItemsListPage.model_validate(vars(response))
+        await self._update_metadata()
```
janbuchar: Doing this after every method call might be costly.

src/apify/_actor.py
```diff
@@ -682,7 +682,23 @@ def get_env(self) -> dict:
                 config[alias] = getattr(self._configuration, field_name)
 
         env_vars = {env_var.value.lower(): env_var.name.lower() for env_var in [*ActorEnvVars, *ApifyEnvVars]}
-        return {option_name: config[env_var] for env_var, option_name in env_vars.items() if env_var in config}
+        result = {option_name: config[env_var] for env_var, option_name in env_vars.items() if env_var in config}
+
+        # These environment variables are not part of the Configuration model,
```
janbuchar: Why though? It would be nice to add them to the `apify.Configuration` model.

tests/integration/conftest.py
janbuchar, line 1: Nice!

src/apify/_actor.py
vdusek: you're right

docs/03_concepts/code/03_dataset_exports.py
vdusek: let's evaluate all the potential BCs at the end

src/apify/apify_storage_client/__init__.py
vdusek: Just wanted to keep the same FS structure as in the Crawlee. Maybe it does not make sense. Should I revert it?

src/apify/storage_clients/_apify/_dataset_client.py
vdusek: That's correct, I believe we had some warning in Crawlee. But in case of storages, users should use `open` "constructor", which is asynchronous. So this should not be a problem.

src/apify/storage_clients/_apify/_dataset_client.py
vdusek: Well, there are two API calls instead of one. And they have to be called consecutively. Not ideal, but since it is `Dataset.get_data` (which retrieves results), it should not be a crawling performance bottleneck. What would be a better approach anyway? Probably the only way is to make it lazy and update metadata only when someone accesses it. That would require making the metadata getter async, which would result in another significant restructure in all storage clients (including Crawlee, of course). I wouldn't go for it.

src/apify/apify_storage_client/__init__.py
janbuchar: It would make sense to me to keep it like it was, sorry.

src/apify/storage_clients/_apify/_dataset_client.py
janbuchar: Can we accept the lock as a parameter to the constructor then? And create it in the `open` method.

vdusek: that would make sense, I'll update it in the Crawlee as well

src/apify/storage_clients/_apify/_dataset_client.py
```diff
@@ -0,0 +1,214 @@
...
+    @override
+    @classmethod
+    async def open(
+        cls,
+        *,
+        id: str | None,
+        name: str | None,
+        configuration: Configuration,
+    ) -> ApifyDatasetClient:
+        token = getattr(configuration, 'token', None)
```
Pijukatel: We do getattr even though we know we want to use only Configuration from Apify, but we type check it as Configuration from Crawlee. That looks like a workaround and we are loosing static type safety and need to do runtime checks.

What is the best approach? I guess that making `class DatasetClient(Generic[TConfiguration]):` and `ApifyDatasetClient(DatasetClient[ApifyConfiguration])` would solve this, but will that open whole range of new problems?

(Same for other clients)

janbuchar: Making stuff generic like this feels like an overkill, yeah. Can we just do an `isinstance(configuration, apify.Configuration)` every time we want to use something that's not in the crawlee config?

Or we could accept `apify.Configuration` (or pull it from service locator) in `ApifyStorageClient` and pass it down in `create_*_client` methods.

src/apify/apify_storage_client/__init__.py
vdusek: Resolving based on our Slack discussion.

src/apify/storage_clients/_apify/_dataset_client.py
janbuchar: But... the metadata were originally (before the storages refactor) retrieved via `get_info`, right?

src/apify/storage_clients/_apify/_dataset_client.py
janbuchar: Also, maybe we could remove `open` from the `DatasetClient`, `KeyValueStoreClient` and `RequestQueueClient` interfaces and leave defining that to the implementer.

src/apify/storage_clients/_apify/_dataset_client.py
vdusek: That's correct... And maybe it's a good point and we should do it - It won't help us in the current local storage clients, but it will help here and maybe in other potential cloud/DB-based storage clients, where you don't have to maintain the metadata manually.

janbuchar: Then I guess the question is if keeping the metadata up to date is worth the possible performance hit. Since in principle, nothing is keeping the storages from being accessed by multiple clients, the metadata may become stale at virtually any point. So on demand fetching actually makes a lot of sense to me. But I'm ready to hear other opinions.

docs/03_concepts/code/03_dataset_exports.py
janbuchar: Sure. I thought we are nearing that now :grin:

src/apify/storage_clients/_apify/_dataset_client.py
vdusek: I'm probably for it (it will be some work of course)... And maybe I would just name it `async def get_metadata()`.

@Pijukatel what do you think?

Pijukatel: If I understand correctly, the metadata would always be accessed through this async getter, and there would be no local copy of it and thus we will not need to update it all the time. It makes sense to me.

vdusek: async metadata getter implemented

src/apify/storage_clients/_apify/_dataset_client.py
vdusek: It should be much better now.

docs/03_concepts/code/03_dataset_exports.py
vdusek: Since we're just re-exporting the `storages` from `crawlee` here, there will be many more cases than this one. I'm not saying we have to rename this particular argument (and I will undo it if you insist—just I don't like those long identifiers, especially when we can use the KVS abbreviation).

src/apify/storage_clients/_apify/_dataset_client.py
```diff
@@ -0,0 +1,288 @@
...
+class ApifyDatasetClient(DatasetClient):
+    """An Apify platform implementation of the dataset client."""
+
+    _MAX_PAYLOAD_SIZE = ByteSize.from_mb(9)
+    """Maximum size for a single payload."""
+
+    _SAFETY_BUFFER_PERCENT = 0.01 / 100  # 0.01%
+    """Percentage buffer to reduce payload limit slightly for safety."""
+
+    _EFFECTIVE_LIMIT_SIZE = _MAX_PAYLOAD_SIZE - (_MAX_PAYLOAD_SIZE * _SAFETY_BUFFER_PERCENT)
```
Pijukatel: Maybe rename:
 _SAFETY_BUFFER_PERCENT to _SAFETY_BUFFER_COEFFICIENT as it is not really percent

_SAFETY_BUFFER_COEFFICIENT = 0.9999 # 0.01% reserve
_EFFECTIVE_LIMIT_SIZE = _MAX_PAYLOAD_SIZE * _SAFETY_BUFFER_COEFFICIENT

Pijukatel (approved):

vdusek:
Probably let's wait for @janbuchar, and also I'll add an upgrading guide for v2.

src/apify/storage_clients/_file_system/_key_value_store_client.py
```diff
@@ -0,0 +1,36 @@
...
+
+class ApifyFileSystemKeyValueStoreClient(FileSystemKeyValueStoreClient):
+    """Apify-specific implementation of the `FileSystemKeyValueStoreClient`.
+
+    The only difference is that it overrides the `purge` method to delete all files in the key-value store
+    directory, except for the metadata file and the `INPUT.json` file.
+    """
+
+    @override
+    async def purge(self) -> None:
```
janbuchar, line 19: Does this override mean that Crawlee does not treat `INPUT.json` differently from other files anymore and this logic is only added in the SDK? I'm a fan, but I wonder if the rest of the team is on board with this.

src/apify/storage_clients/_apify/_key_value_store_client.py
```diff
@@ -0,0 +1,238 @@
...
+            url_signing_secret_key = metadata.model_extra.get('urlSigningSecretKey')
+            if url_signing_secret_key is not None:
+                public_url = public_url.with_query(signature=create_hmac_signature(url_signing_secret_key, key))
+
+        return str(public_url)
+
+    async def _update_metadata(self) -> None:
+        """Update the key-value store metadata with current information."""
+        metadata = await self._api_client.get()
+        self._metadata = KeyValueStoreMetadata.model_validate(metadata)
```
janbuchar: Is this ever used? I recall discussing not storing the metadata, but I'm not sure what the conclusion was.

src/apify/storage_clients/_apify/_key_value_store_client.py
```diff
@@ -0,0 +1,238 @@
...
+        self._api_public_base_url = api_public_base_url
+        """The public base URL for accessing the key-value store records."""
+
+        self._lock = lock
+        """A lock to ensure that only one operation is performed at a time."""
+
+    @override
+    async def get_metadata(self) -> KeyValueStoreMetadata:
+        metadata = await self._api_client.get()
+        return KeyValueStoreMetadata.model_validate(metadata)
```
janbuchar: Could this use an extended model that would include `urlSigningSecretKey`?

janbuchar (commented):
I haven't made it through the request queue client, but here's a batch of comments.

src/apify/storage_clients/_apify/_key_value_store_client.py
vdusek: Thanks, good catch! This is really forgotten code from the previous state. There is the `get_metadata` method now.

src/apify/storage_clients/_file_system/_key_value_store_client.py
vdusek: Exactly - because Crawlee does not use `INPUT.json` for anything. We are in the same boat with @Pijukatel here.

@B4nan, would you be OK with this change? TLDR; so you don't have to study the code: As Honza wrote, Crawlee now treats `INPUT.json` just like any other file. Here in SDK, I've added a minimal override of `FileSystemStorageClient` that only customizes the `purge` method, preventing it from deleting `INPUT.json`. That's it.

src/apify/storage_clients/_apify/_key_value_store_client.py
vdusek: sure

src/apify/storage_clients/_file_system/_key_value_store_client.py
B4nan: are we talking about the input.json in the project root? since that is implemented on crawlee level (in JS) and i think it should stay this way. i am not talking about the purging logic, in JS, if there is an INPUT.json file in the root, `KVS.getInput()` will prefer it over anything in the storage folder (the local/memory implementation, with apify client it is ignored completely). not sure if we even do this in the python version (and I am open to discuss if we want to remove this from v4, but it feels like a good thing to me)

if this is just about purging, i am not entirely sure i understand what it means, we dont want to autopurge inputs on crawlee level, crawlee shouldnt behave differently because you init the SDK (not locally).

B4nan: this is the part where we handle input file in the project root in JS

https://github.com/apify/crawlee/blob/5c4726df96e358a9bbf44a0cd2760e4e269f0fae/packages/core/src/storages/key_value_store.ts#L697

janbuchar: From the "Crawlee is a web scraping framework" point of view, handling input feels a bit redundant here - if you intend to only run the crawler outside of Apify, you'll probably want to accept input some other way - CLI arguments, environment variables, some domain-speciffic config file... Providing a helper for loading a JSON file from two possible locations doesn't add much value IMO.

Asking Apifiers about this doesn't make much sense in this situation, because basically everything we write is intended to run on Apify :slightly_smiling_face:

Pijukatel (changes requested):
While testing the change on platform I used modified benchmark tests. They are specific, because they scrape locally hosted site and so the speed of scraping is unusually fast - no network related delays. This exposes  RequestQueue race condition that can happen when checking if the RQ is empty or not.
When running the test using this branch, I saw that the crawler premature finished even though RQ was still quite full. The hypothesis about premature exit was confirmed when running same test on branch modified with extra log and sleep.

Please see following tiny commit based on this branch and attached log:
https://github.com/apify/apify-sdk-python/commit/5b25728e9ca70f9aa2642194a5951c8db6b8ebfd


image

src/apify/storage_clients/_apify/_request_queue_client.py
```diff
@@ -0,0 +1,650 @@
...
+            requests: The requests to add.
+            forefront: Whether to add the requests to the beginning of the queue.
+
+        Returns:
+            Response containing information about the added requests.
+        """
+        # Prepare requests for API by converting to dictionaries
+        requests_dict = [request.model_dump(by_alias=True) for request in requests]
+
+        # Remove 'id' fields from requests as the API doesn't accept them
```
Pijukatel: I think this could be achieved during model dump by `request.model_dump(by_alias=True, exclude={"id"})`

src/apify/storage_clients/_apify/_request_queue_client.py
```diff
@@ -0,0 +1,650 @@
...
+        lock: asyncio.Lock,
+    ) -> None:
+        """Initialize a new instance.
+
+        Preferably use the `ApifyRequestQueueClient.open` class method to create a new instance.
+        """
+        self._api_client = api_client
+        """The Apify request queue client for API operations."""
+
+        self._api_public_base_url = api_public_base_url
```
Pijukatel: Is this used anywhere if the client is already created?

vdusek: It is not, thanks.

Pijukatel (approved):

docs/04_upgrading/upgrading_to_v3.md
```diff
@@ -0,0 +1,18 @@
+---
+id: upgrading-to-v2
+title: Upgrading to v2
```
Mantisus, line 3: to v3

Issue #472, opened 2025-05-16
Add conveninece `RemainingTime` option for `Actor.start` and `Actor.call`


Pijukatel:
Add convenient way to start another Actor with reasonable timeout to prevent lingering actor runs after main actor shuts down.

This is one part of larger feature request in JS version of the client: https://github.com/apify/apify-client-js/issues/632
(Will be first implemented in Python and then ported to JS version)

PR #473, opened 2025-05-16
feat: Add `RemainingTime` option for `timeout` argument of `Actor.call` and `Actor.start`

Pijukatel:
### Description

Added convenient way to start another Actor with reasonable timeout to prevent lingering actor runs after main actor shuts down.
Added test.

`RemainingTime` value of the timeout argument is not a default one and has to be explicitly passed by the user. Using this value will calculate remaining time of this actor run and pass it as a timeout to the actor that is being started.

### Issues

- Closes: #472


src/apify/_actor.py
```diff
@@ -732,18 +733,35 @@ async def start(
...
         )
 
         return ActorRun.model_validate(api_result)
 
+    def _get_remaining_time(self) -> timedelta | None:
+        """Get time remaining from the actor timeout. Returns `None` if not on an Apify platform."""
+        if self.is_at_home() and self.configuration.timeout_at:
+            return self.configuration.timeout_at - datetime.now(tz=timezone.utc)
+
+        self.log.warning('Using `RemainingTime` argument for timeout outside of the Apify platform. Returning `None`')
```
janbuchar, line 762: Could this be printed out if `self.is_at_home() is True`, but `self.configuration.timeout_at is None`? I assume that this is unlikely, but still.

src/apify/_actor.py
```diff
@@ -711,7 +711,8 @@ async def start(
             memory_mbytes: Memory limit for the run, in megabytes. By default, the run uses a memory limit specified
                 in the default run configuration for the Actor.
             timeout: Optional timeout for the run, in seconds. By default, the run uses timeout specified in
-                the default run configuration for the Actor.
+                the default run configuration for the Actor. Using `RemainingTime` will set timeout of the other actor
```
vdusek, line 714: Please, start using capitalized `Actor`.

src/apify/_actor.py
```diff
@@ -732,18 +733,35 @@ async def start(
...
             serialized_webhooks = None
 
+        if timeout == 'RemainingTime':
+            actor_start_timeout = self._get_remaining_time()
+        elif isinstance(timeout, str):
+            raise ValueError(
+                f'`timeout` can be `None`, `RemainingTime` literal or `timedelta` instance, but is {timeout=}'
+            )
+        else:
+            actor_start_timeout = timeout
```
vdusek, line 743: Maybe swap the 2nd and 3rd branch would be better?

```suggestion
        if timeout == 'RemainingTime':
            actor_start_timeout = self._get_remaining_time()
        elif timeout is None:
            actor_start_timeout = None
        elif isinstance(timeout, timedelta):
            actor_start_timeout = timeout
        else:
            raise ValueError(
                f'Invalid timeout {timeout!r}: expected `None`, `"RemainingTime"`, or a `timedelta`.'
            )
```

src/apify/_actor.py
```diff
@@ -826,12 +845,21 @@ async def call(
...
             serialized_webhooks = None
 
+        if timeout == 'RemainingTime':
+            actor_call_timeout = self._get_remaining_time()
+        elif isinstance(timeout, str):
+            raise ValueError(
+                f'`timeout` can be `None`, `RemainingTime` literal or `timedelta` instance, but is {timeout=}'
+            )
+        else:
+            actor_call_timeout = timeout
```
vdusek, line 855: Same as above

src/apify/_actor.py
Pijukatel: Fixed

src/apify/_actor.py
Pijukatel: Fixed

src/apify/_actor.py
Pijukatel: Expanded the warning

vdusek (approved):
LGTM

janbuchar (approved):

PR #482, opened 2025-06-05
feat(crypto): decrypt secret objects

MFori:
Add support for `isSecret` objects and arrays in input schema and the new form of encrypted value:

```
ENCRYPTED_JSON_VALUE:{FIELD_SCHEMA_HASH}:{ENCRYPTED_PASSWORD}:{ENCRYPTED_VALUE}
```

More context here https://github.com/apify/apify-shared-js/pull/515


vdusek (approved):
LGTM

PR #487, opened 2025-06-18
feat: Expose `logger` argument on `Actor.call` to control log redirection from started Actor run

Pijukatel:
### Description

Added `logger` argument is used to either disable log redirection, use the default redirection logger or to pass custom logger that will process the redirected logs from started Actor run.

### Issues

Related: https://github.com/apify/apify-client-js/issues/632

### DIFF
--- a/tests/unit/conftest.py
+++ b/tests/unit/conftest.py
@@ -38,7 +38,8 @@ def prepare_test_env(monkeypatch: pytest.MonkeyPatch, tmp_path: Path) -> Callabl
     """
 
     def _prepare_test_env() -> None:
-        delattr(apify._actor.Actor, '__wrapped__')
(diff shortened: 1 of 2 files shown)


tests/unit/conftest.py
```diff
@@ -38,7 +38,8 @@ def prepare_test_env(monkeypatch: pytest.MonkeyPatch, tmp_path: Path) -> Callabl
     """
 
     def _prepare_test_env() -> None:
-        delattr(apify._actor.Actor, '__wrapped__')
+        if hasattr(apify._actor.Actor, '__wrapped__'):
```
Pijukatel, line 41: This is very mysterious. Why did the previous implementation work on older Python versions and not in new ones (sometimes)?

vdusek (approved):
LGTM

PR #504, opened 2025-07-15
docs: Add log redirection guide

Pijukatel:
### Description

Add log redirection guide

### Issues

- Relates to: [#632](https://github.com/apify/apify-client-js/issues/632)


vdusek (changes requested):
Use `webp` for images please.

docs/03_concepts/10_logging.mdx
```diff
@@ -82,3 +85,28 @@ Result:
 </pre>
 
 <!-- markdownlint-enable no-inline-html -->
+
+## Redirect logs from other actor runs
+In some situations, one Actor is going to start one or more other Actors and wait for them to finish and produce some results. In such cases, you might want to redirect the logs and status messages of the started Actors runs back to the parent Actor run, so that you can see the progress of the started Actors' runs in the parent Actor's logs. This guide will show possibilities on how to do it.
```
vdusek:
```suggestion
## Redirect logs from other Actor runs

In some situations, one Actor is going to start one or more other Actors and wait for them to finish and produce some results. In such cases, you might want to redirect the logs and status messages of the started Actors runs back to the parent Actor run, so that you can see the progress of the started Actors' runs in the parent Actor's logs. This guide will show possibilities on how to do it.
```

docs/03_concepts/10_logging.mdx
```diff
@@ -82,3 +85,28 @@ Result:
...
+
+### Redirecting logs from `Actor.call`
+Typical use case for log redirection is to call another Actor using the [`Actor.call`](../../reference/class/Actor#call) method. This method has an optional `logger` argument, which is by default set to the `default` literal. This means that the logs of the called Actor will be automatically redirected to the parent Actor's logs with default formatting and filtering. If you set the `logger` argument to `None`, then no log redirection happens. The third option is to pass your own `Logger` instance with the possibility to define your own formatter, filter, and handler. Below you can see those three possible ways of log redirection when starting another Actor run through [`Actor.call`](../../reference/class/Actor#call).
```
vdusek: Do not use backtick sections in the headings - it doesn't render well.

```suggestion
### Redirecting logs from Actor.call

Typical use case for log redirection is to call another Actor using the [`Actor.call`](../../reference/class/Actor#call) method. This method has an optional `logger` argument, which is by default set to the `default` literal. This means that the logs of the called Actor will be automatically redirected to the parent Actor's logs with default formatting and filtering. If you set the `logger` argument to `None`, then no log redirection happens. The third option is to pass your own `Logger` instance with the possibility to define your own formatter, filter, and handler. Below you can see those three possible ways of log redirection when starting another Actor run through [`Actor.call`](../../reference/class/Actor#call).
```

docs/03_concepts/10_logging.mdx
```diff
@@ -82,3 +85,28 @@ Result:
...
+
+The log redirection can be deep, meaning that if the other actor also starts another actor and is redirecting logs from it, then in the top-level Actor, you can see it as well. See the following example screenshot of the Apify log console when one actor recursively starts itself (there are 2 levels of recursion in the example).
+
+![Console with redirected logs](/img/guides/redirected_logs_example.webp 'Example of console with redirected logs from recursively started actor.')
+
+### Redirecting logs from already running Actor run
+In some cases, you might want to connect to an already running Actor run and redirect its logs to your current Actor run. This can be done using the [ApifyClient](../../reference/class/Actor#apify_client) and getting the streamed log from a specific Actor run. You can then use it as a context manager, and the log redirection will be active in the context, or you can control the log redirection manually by explicitly calling `start` and `stop` methods.
```
vdusek:
```suggestion
### Redirecting logs from already running Actor run

In some cases, you might want to connect to an already running Actor run and redirect its logs to your current Actor run. This can be done using the [ApifyClient](../../reference/class/Actor#apify_client) and getting the streamed log from a specific Actor run. You can then use it as a context manager, and the log redirection will be active in the context, or you can control the log redirection manually by explicitly calling `start` and `stop` methods.
```

docs/03_concepts/10_logging.mdx
```diff
@@ -7,6 +7,9 @@ import CodeBlock from '@theme/CodeBlock';
 
 import LogConfigExample from '!!raw-loader!./code/10_log_config.py';
 import LoggerUsageExample from '!!raw-loader!./code/10_logger_usage.py';
+import RedirectLog from '!!raw-loader!./code/10_redirect_log.py';
+import RedirectLogExistingRun from '!!raw-loader!./code/10_redirect_log_existing_run.py';
+
 
 The Apify SDK is logging useful information through the [`logging`](https://docs.python.org/3/library/logging.html) module from Python's standard library, into the logger with the name `apify`.
```
vdusek:
```suggestion
import RedirectLog from '!!raw-loader!./code/10_redirect_log.py';
import RedirectLogExistingRun from '!!raw-loader!./code/10_redirect_log_existing_run.py';

The Apify SDK is logging useful information through the [`logging`](https://docs.python.org/3/library/logging.html) module from Python's standard library, into the logger with the name `apify`.
```

vdusek (commented):
a few comments

vdusek (approved):
LGTM

PR #518, opened 2025-07-25
chore: Add more RQ integration tests

vdusek:


tests/integration/test_request_queue.py
```diff
@@ -78,36 +85,1069 @@ async def test_add_non_unique_requests_in_batch(
...
+            fetched_urls = []
+            while next_request := await rq.fetch_next_request():
+                Actor.log.info(f'Fetched request: {next_request.url}')
+                fetched_urls.append(next_request.url)
+                await rq.mark_request_as_handled(next_request)
+
+            # Forefront requests should come first (in reverse order of addition)
+            assert fetched_urls[0] == 'https://example.com/priority2'
+            assert fetched_urls[1] == 'https://example.com/priority1'
+            assert 'https://example.com/1' in fetched_urls[2:]
```
Pijukatel: These last 3 requests should be ordered as well, so maybe just assert whole list at once?
`fetched_urls = ['https://example.com/priority2', ..., 'https://example.com/3']`

tests/integration/test_request_queue.py
```diff
@@ -78,36 +85,1069 @@ async def test_add_non_unique_requests_in_batch(
...
+            assert 'https://example.com/2' in fetched_urls[2:]
+            assert 'https://example.com/3' in fetched_urls[2:]
+            Actor.log.info(f'Forefront ordering verified: {fetched_urls}')
```
Pijukatel: This and similar end-of-test logs seem somehow redundant because of the asserts

tests/integration/test_request_queue.py
```diff
@@ -78,36 +85,1069 @@ async def test_add_non_unique_requests_in_batch(
...
+            # The reclaimed request should be fetched first again
+            next_request = await rq.fetch_next_request()
+            assert next_request is not None
+            assert next_request.url == first_request.url
+            Actor.log.info(f'Confirmed reclaimed request came first: {next_request.url}')
+
+            # Clean up
+            await rq.mark_request_as_handled(next_request)
+            remaining_count = 0
+            while True:
```
Pijukatel: In another test you use this:
```
while next_request := await rq.fetch_next_request():
```
Maybe use it here as well

tests/integration/test_request_queue.py
```diff
@@ -78,36 +85,1069 @@ async def test_add_non_unique_requests_in_batch(
...
+                remaining_request = await rq.fetch_next_request()
+                if remaining_request is None:
+                    break
+                remaining_count += 1
+                await rq.mark_request_as_handled(remaining_request)
+
+            Actor.log.info(f'Test completed - processed {remaining_count} additional requests')
```
Pijukatel, line 337: What is the point of the end of the test without any asserts? Is there any point in doing any Rq cleanup?

tests/integration/test_request_queue.py
```diff
@@ -78,36 +85,1069 @@ async def test_add_non_unique_requests_in_batch(
...
+            Actor.log.info(f'Total requests in queue after batch add: {total_count}')
+            assert total_count == 500
+
+            # Process all in chunks to test performance
+            processed_count = 0
+            chunk_number = 0
+            while not await rq.is_empty():
+                chunk_number += 1
+
+                # Process in chunks of 50
```
Pijukatel: What is the point of this `loop in loop `# Process in chunks?

tests/integration/test_request_queue.py
```diff
@@ -78,36 +85,1069 @@ async def test_add_non_unique_requests_in_batch(
...
+                await rq.add_request(f'https://example.com/concurrent/{i}')
+            Actor.log.info('Added 20 requests for concurrent processing')
+
+            total_count = await rq.get_total_count()
+            Actor.log.info(f'Total requests in queue: {total_count}')
+
+            # Simulate concurrent workers
+            async def worker(_worker_id: int) -> int:
+                processed = 0
+                while True:
```
Pijukatel: In another test you use this:
```
while next_request := await rq.fetch_next_request():
```
Maybe use it here as well

tests/integration/test_request_queue.py
```diff
@@ -78,36 +85,1069 @@ async def test_add_non_unique_requests_in_batch(
...
+
+            # Run multiple workers concurrently
+            workers = [worker(i) for i in range(3)]
+            results = await asyncio.gather(*workers)
+
+            total_processed = sum(results)
+            Actor.log.info(f'Total processed by workers: {total_processed}')
+            Actor.log.info(f'Individual worker results: {results}')
+
+            # Process any remaining reclaimed requests
```
Pijukatel, line 805: I am missing some assert in this  test

tests/integration/test_request_queue.py
```diff
@@ -78,36 +85,1069 @@ async def test_add_non_unique_requests_in_batch(
...
+                while True:
+                    request = await rq.fetch_next_request()
+                    if request is None:
+                        break
+
+                    # Simulate some work
+                    await asyncio.sleep(0.01)
+
+                    # Randomly reclaim some requests (simulate failures)
+                    if processed % 7 == 0 and processed > 0:  # Reclaim every 7th request
```
Pijukatel, line 775: Maybe just consume the whole queue and then `assert total_processed = sum(results) == 20`


`if loop_counter % 7 == 0 and processed > 0:`

tests/integration/test_request_queue.py
```diff
@@ -78,36 +85,1069 @@ async def test_add_non_unique_requests_in_batch(
...
+) -> None:
+    """Test edge cases in request deduplication."""
+
+    async def main() -> None:
+        async with Actor:
+            rq = await Actor.open_request_queue()
+            Actor.log.info('Request queue opened')
+
+            # Test URL normalization and deduplication
+            urls = [
```
Pijukatel: I suggest to keep information about dedulpication expectation in code and not comments and then use it for asserting exact expected `len(fetched_urls)`

Something like:
```
urls_and_deduplication_expectations = [...
('https://example.com/page/',True),
('https://example.com/page#fragment', False),
...
]
```

tests/integration/test_request_queue.py
```diff
@@ -78,36 +85,1069 @@ async def test_add_non_unique_requests_in_batch(
...
+async def test_request_queue_isolation(
+    make_actor: MakeActorFunction,
+    run_actor: RunActorFunction,
+) -> None:
+    """Test that different request queues are properly isolated."""
+
+    async def main() -> None:
+        async with Actor:
+            # Open multiple queues
+            rq1 = await Actor.open_request_queue(name='queue-1')
```
Pijukatel: These tests can run in parallel with themselves due to CI starting them on different PRs, so any named queue has to have a unique name or there is a risk of unexpected results in a rare case when the tests run at the same time. Maybe re-use the already unique name of the generated actor?

tests/integration/test_request_queue.py
```diff
@@ -78,36 +85,1069 @@ async def test_add_non_unique_requests_in_batch(
...
+            await rq1.drop()
+            await rq2.drop()
+            Actor.log.info('Dropped both queues')
+
+    actor = await make_actor(label='rq-isolation-test', main_func=main)
+    run_result = await run_actor(actor)
+    assert run_result.status == 'SUCCEEDED'
+
+
+async def test_request_unique_key_behavior(
```
Pijukatel: This looks similar like the test `test_duplicate_requests_by_unique_key` on line 173 , is there any significant difference?

tests/integration/test_request_queue.py
```diff
@@ -78,36 +85,1069 @@ async def test_add_non_unique_requests_in_batch(
...
+            unique_keys = {req.unique_key for req in fetched_requests}
+            Actor.log.info(f'Unique keys in fetched requests: {unique_keys}')
+            assert unique_keys == {'custom-key-1', 'custom-key-2'}
+
+    actor = await make_actor(label='rq-unique-key-test', main_func=main)
+    run_result = await run_actor(actor)
+    assert run_result.status == 'SUCCEEDED'
+
+
+async def test_queue_finished_state_accuracy(
```
Pijukatel: What about this scenario with race condition:

Empty RQ.
one coroutine awaits RQ.add_request
other coroutine awaits RQ.is_finished after calling RQ.add_request, but it gets response faster than the RQ.add_request and such response should say it is finished, but it should be also aware if it's own add_request call and wait for it's resolution

This scenario should be unit tested though, as this kind of delicate timing requires mocking

tests/integration/test_request_queue.py
vdusek: Sure, I'm just gonna do separate asserts because of a better error message

tests/integration/test_request_queue.py
vdusek: The issue is that it raises a bare `AssertionError` without any message when the condition fails. Since this is not a normal pytest test, Python doesn't display the actual left and right values. For example:

```
2025-07-26T10:12:43.393Z [apify] INFO  Initializing Actor...
2025-07-26T10:12:43.397Z [apify] INFO  System info ({"apify_sdk_version": "2.7.1", "apify_client_version": 
2025-07-26T10:12:43.650Z [apify] INFO  Request queue opened
2025-07-26T10:12:43.774Z [apify] INFO  Initial state - Total: 0, Handled: 0
2025-07-26T10:12:43.870Z [apify] INFO  Added 5 requests in batch
2025-07-26T10:12:43.910Z [apify] INFO  After adding - Total: 0, Handled: 0
2025-07-26T10:12:43.912Z [apify] ERROR Actor failed with an exception
2025-07-26T10:12:43.916Z       Traceback (most recent call last):
2025-07-26T10:12:43.918Z         File "/usr/src/app/src/main.py", line 27, in main
2025-07-26T10:12:43.920Z           assert total_after_add == 5
2025-07-26T10:12:43.922Z                  ^^^^^^^^^^^^^^^^^^^^
2025-07-26T10:12:43.924Z       AssertionError
```

But I'm gonna add assertion messages instead.

tests/integration/test_request_queue.py
vdusek: same issue as explained in https://github.com/apify/apify-sdk-python/pull/518#discussion_r2234819409, resolving and let's discuss it there

tests/integration/test_request_queue.py
vdusek: updated

tests/integration/test_request_queue.py
vdusek: I'm gonna leave just `test_request_unique_key_behavior` test.

tests/integration/test_request_queue.py
Pijukatel: I added it here. Maybe it can be closed if we are sure it is properly handled on the `RequestQueue` level
https://github.com/apify/apify-sdk-python/issues/519

PR #532, opened 2025-08-06
chore: Replace `respx` mocks with `pytest-httpserver` in tests

Mantisus:
- Replace `respx` mocks with `pytest-httpserver` in tests

### DIFF
--- a/tests/unit/actor/test_actor_create_proxy_configuration.py
+++ b/tests/unit/actor/test_actor_create_proxy_configuration.py
@@ -1,9 +1,8 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING
 from unittest.mock import Mock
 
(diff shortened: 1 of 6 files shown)


tests/unit/actor/test_actor_create_proxy_configuration.py
```diff
@@ -24,25 +28,43 @@ def patched_apify_client(apify_client_async_patcher: ApifyClientAsyncPatcher) ->
     return ApifyClientAsync()
 
 
+@pytest.fixture
+def patched_httpx_client(monkeypatch: pytest.MonkeyPatch) -> Iterator[None]:
```
Pijukatel: Seems like this identical fixture is defined twice, maybe move it to  test/unit/conftest.py?

Mantisus: Great point

Issue #514, opened 2025-07-24
Ensure that duplicate links are handled in a cost effective way when using Apify RequestQueue


Pijukatel:
In JS version fo RQ, there is [this](https://github.com/apify/crawlee/blob/master/packages/core/src/storages/request_provider.ts#L165-L179) and [this](https://github.com/apify/crawlee/blob/master/packages/core/src/storages/request_queue_v2.ts#L95-L108), which deal with duplicate links that would otherwise dramatically increase RequestQueue writes. Ensure that similar handling is done in Python version of the tooling.

For more context:
https://apify.slack.com/archives/C02JQSN79V4/p1753358299203499

Consider solving this in Crawlee level if possible

B4nan:
Why would we do that? It literally means ignoring the problem on crawlee level (where it is a problem as well, rather a perf issue than a cost issue like on the platform, but still an issue).

janbuchar:
We can either try to handle deduplication in a general way in `storages.RequestQueue`, or we can make it a requirement of the `storage_clients.RequestQueueClient` interface (either a soft requirement explained in docs or some method(s) that would guarantee this).

The current state in crawlee-python is the latter (explained in the docblock of `RequestQueueClient.add_batch_of_requests`). There are now two implementations of the interface that resolve this on their own:  `MemoryRequestQueueClient`, which has a unique_key => request dictionary and `FilesystemRequestQueueClient`, which handles this with a file system traversal. The first one does not need any additional optimization (it would needlessly consume memory), the second (file system) does not seem optimal (@vdusek am I reading the code right?).

For a future SQL-backed storage client, I believe that a database-level upsert will be more reliable and efficient than anything we can manage on the Crawlee side.

In the Apify platform client, handling this on the client side reduces usage costs.

It seems like a 2:2 tie to me 😁

vdusek:
I believe deduplication should be the responsibility of the storage client, and that's how it's currently implemented. Mainly because the optimal deduplication strategy is tightly coupled with the "storage backend" itself.

As Honza pointed out...
- For in-memory storage, we can use a dictionary keyed by `unique_key`, which should be easy and optimal.
- The filesystem storage client isn't currently optimal (we discussed this about a month ago during a team sync and agreed to optimize it after the v1.0 release).
- For the Apify platform client, we should implement something on the client side (and it can be even shared with the file-system client, for example, if it makes sense).
- SQL-based storage clients can handle deduplication at the database.
[...]

PR #534, opened 2025-08-07
feat: Add deduplication to `add_batch_of_requests`

Pijukatel:
### Description

- Ensure that already known requests are excluded from `api_client.batch_add_requests` calls to avoid expensive and pointless API calls.
- Add all new requests to the cache when calling `batch_add_requests`.
- Add test with real API usage measurement.

### Issues

- Closes: #514

### Testing

- Added new integration tests to verify reduced API usage.
- Comparing benchmark actor based on master vs this PR. Actor is a simple ParselCrawler that crawls the whole crawlee.dev, which contains many duplicate links, as the documentation is cross-linked thoroughly. Results:
  - Massive reduction of cost from request queue.
  - Significant overall speed up due to reduced API calls.

image


tests/integration/test_actor_request_queue.py
```diff
@@ -61,55 +81,171 @@ async def main() -> None:
...
+            # Get raw client, because stats are not exposed in `RequestQueue` class, but are available in raw client
+            rq_client = Actor.apify_client.request_queue(request_queue_id=rq.id)
+            _rq = await rq_client.get()
+            assert _rq
+            stats_before = _rq.get('stats', {})
+            Actor.log.info(stats_before)
+
+            # Add same request twice
+            await rq.add_request(request)
+            await rq.add_request(request)
```
janbuchar: Maybe you could make two distinct `Request` instances with the same `uniqueKey` here?

tests/integration/test_actor_request_queue.py
```diff
@@ -61,55 +81,171 @@ async def main() -> None:
...
+            assert _rq
+            stats_before = _rq.get('stats', {})
+            Actor.log.info(stats_before)
+
+            # Add same requests in 10 parallel workers
+            async def add_requests_worker() -> None:
+                await rq.add_requests(requests)
+
+            add_requests_workers = [asyncio.create_task(add_requests_worker()) for _ in range(10)]
+            await asyncio.gather(*add_requests_workers)
```
janbuchar, line 257: I guess you made sure that these do in fact run in parallel? To the naked eye, 100 requests doesn't seem like much, I'd expect that the event loop may run the tasks in sequence.

Maybe you could add the requests in each worker in smaller batches and add some random delays? Or just add a comment saying that you verified parallel execution empirically 😁

tests/integration/test_actor_request_queue.py
```diff
@@ -61,55 +81,171 @@ async def main() -> None:
...
+
+    with mock.patch(
+        'apify_client.clients.resource_clients.request_queue.RequestQueueClientAsync.batch_add_requests',
+        side_effect=return_unprocessed_requests,
+    ):
+        # Simulate failed API call for adding requests. Request was not processed and should not be cached.
+        await apify_named_rq.add_requests(['http://example.com/1'])
+
+    # This will succeed.
+    await apify_named_rq.add_requests(['http://example.com/1'])
```
janbuchar, line 311: Any chance we could verify that the request was actually not cached between the two `add_requests` calls?

src/apify/storage_clients/_apify/_request_queue_client.py
```diff
@@ -242,17 +242,66 @@ async def add_batch_of_requests(
...
-            request.model_dump(
-                by_alias=True,
-                exclude={'id'},  # Exclude ID fields from requests since the API doesn't accept them.
-            )
-            for request in requests
-        ]
+        # Do not try to add previously added requests to avoid pointless expensive calls to API
+
+        new_requests: list[Request] = []
+        already_present_requests: list[dict[str, str | bool]] = []
```
janbuchar: This type is technically correct, but not very helpful. Since we run it through validation afterwards, perhaps we could just do `dict[str, Any]`?

src/apify/storage_clients/_apify/_request_queue_client.py
```diff
@@ -242,17 +242,66 @@ async def add_batch_of_requests(
...
+
+        for request in requests:
+            if self._requests_cache.get(request.id):
+                # We are no sure if it was already handled at this point, and it is not worth calling API for it.
```
janbuchar: Did you mean "We are now sure that it was already handled..."? I'm not sure 😁

src/apify/storage_clients/_apify/_request_queue_client.py
```diff
@@ -242,17 +242,66 @@ async def add_batch_of_requests(
...
+            requests_dict = [
+                request.model_dump(
+                    by_alias=True,
+                    exclude={'id'},  # Exclude ID fields from requests since the API doesn't accept them.
+                )
+                for request in new_requests
+            ]
+
+            # Send requests to API.
+            response = await self._api_client.batch_add_requests(requests=requests_dict, forefront=forefront)
```
janbuchar: It's probably out of the scope of the PR, but it might be worth it to validate the response with a Pydantic model.

Pijukatel: That happens in the original code already few lines down:  `api_response = AddRequestsResponse.model_validate(response)`

src/apify/storage_clients/_apify/_request_queue_client.py
Pijukatel: Yes, that was not very clear. Updated

src/apify/storage_clients/_apify/_request_queue_client.py
Pijukatel: I would not mind, but Mypy does.

tests/integration/test_actor_request_queue.py
Pijukatel: This is checked implicitly in the last line where it is asserted that there was exactly 1 writeCount difference. The first call is "hardcoded" to fail, even on all retries, so it never even sends the API request and thus has no chance of increasing the writeCount.

The second call can make the write only if it is not cached, as cached requests do not make the call (tested in other tests). So this means the request was not cached in between.

I could assert the state of the cache in between those calls, but since it is kind of an implementation detail, I would prefer not to.

tests/integration/test_actor_request_queue.py
Pijukatel: I wrote the test for the implementation that did not take parallel execution into account, and it was failing consistently. So from that perspective, I consider the test sufficient.

Anyway, I added some chunking to make the test slightly more challenging. The parallel execution can be verified in the logs. For example, below. From the logs it can be seen that the `add_batch_of_requests` that was started first did not finish first - as it was "taken over" during it's await by another worker.

```
DEBUG Tried to add new requests: 10, succeeded to add new requests: 10, skipped already present requests: 10
DEBUG Tried to add new requests: 10, succeeded to add new requests: 10, skipped already present requests: 20
DEBUG Tried to add new requests: 10, succeeded to add new requests: 10, skipped already present requests: 90
DEBUG Tried to add new requests: 10, succeeded to add new requests: 10, skipped already present requests: 80
DEBUG Tried to add new requests: 10, succeeded to add new requests: 10, skipped already present requests: 0
DEBUG Tried to add new requests: 10, succeeded to add new requests: 10, skipped already present requests: 40
DEBUG Tried to add new requests: 10, succeeded to add new requests: 10, skipped already present requests: 50
DEBUG Tried to add new requests: 10, succeeded to add new requests: 10, skipped already present requests: 60
DEBUG Tried to add new requests: 10, succeeded to add new requests: 10, skipped already present requests: 30
DEBUG Tried to add new requests: 10, succeeded to add new requests: 10, skipped already present requests: 70
INFO  {'readCount': 0, 'writeCount': 100, 'deleteCount': 0, 'headItemReadCount': 0, 'storageBytes': 7400}
```

tests/integration/test_actor_request_queue.py
janbuchar: Fair enough, can you explain this in a comment then?

src/apify/storage_clients/_apify/_request_queue_client.py
janbuchar: Bah. Thanks mypy. I guess it's not worth losing sleep over it.

src/apify/storage_clients/_apify/_request_queue_client.py
janbuchar: I'm sorry, I meant validating the whole response object with the two lists, so that you wouldn't need to do `response['unprocessedRequests']`

tests/integration/test_actor_request_queue.py
Pijukatel: Yes, good point. I also added one more test to make it explicit that deduplication works based on `unique_key` only and unless we use `use_extended_unique_key` argument, some attributes of the request might be ignored. Another test makes this behavior clearly intentional to avoid some confusion in the future.

src/apify/storage_clients/_apify/_request_queue_client.py
Pijukatel: I see, added.

src/apify/storage_clients/_apify/_request_queue_client.py
Pijukatel: Solved by doing early response validation, as this type is now: `already_present_requests: list[ProcessedRequest] = []`

tests/integration/test_actor_request_queue.py
Pijukatel: Yes, added to the test description.

src/apify/storage_clients/_apify/_request_queue_client.py
```diff
@@ -242,20 +242,77 @@ async def add_batch_of_requests(
...
-            request.model_dump(
-                by_alias=True,
-                exclude={'id'},  # Exclude ID fields from requests since the API doesn't accept them.
+        # Do not try to add previously added requests to avoid pointless expensive calls to API
+
+        new_requests: list[Request] = []
+        already_present_requests: list[ProcessedRequest] = []
+
+        for request in requests:
+            if self._requests_cache.get(request.id):
```
janbuchar, line 251: Judging by https://github.com/apify/crawlee/issues/3120, a day may come when we try to limit the size of `_requests_cache` somehow. Perhaps we should think ahead and come up with a more space-efficient way of tracking already added requests?

EDIT: hollup a minute, do you use the ID here for deduplication instead of unique key?

Pijukatel: Since there is this deterministic transformation function `unique_key_to_request_id`, which respects Apify platform way of creating IDs, this seems ok. If someone starts creating Requests with a custom id, then deduplication will most likely stop working.

There are two issues I created based on the discussion about this:
- https://github.com/apify/apify-sdk-python/issues/550
- https://github.com/apify/crawlee-python/issues/1358

tests/integration/test_actor_request_queue.py
```diff
@@ -61,55 +81,239 @@ async def main() -> None:
...
-            )
-
-            await request_queue.mark_request_as_handled(request)
-            assert await request_queue.is_finished()
-        finally:
-            await request_queue.drop()
+    async def main() -> None:
+        import asyncio
+
+        from apify import Actor, Request
```
Mantisus, line 131: Why are we duplicating top-level imports here?

Pijukatel: It is a special setup of the integration tests in this repo. The `make_actor` function will create a new actor from the `main`, but anything outside of the `main` will not be available in the actor. So the tests have these seemingly duplicate imports to have those package in the scope of the created actor.

Mantisus: Thanks for the explanation.

PR #537, opened 2025-08-08
Suggest solving the forefront handling of reclaimed requests

Pijukatel:
Drop forefront info from local chache as it is unreliable and complicates the flow. Only the platform knows the real forefront, do not try to keep imperfect copy of it locally and rather design the system to work without being aware of the cached forefront.

### DIFF
--- a/src/apify/storage_clients/_apify/_models.py
+++ b/src/apify/storage_clients/_apify/_models.py
@@ -105,6 +105,3 @@ class CachedRequest(BaseModel):
 
     lock_expires_at: datetime | None = None
     """The expiration time of the lock on the request."""
-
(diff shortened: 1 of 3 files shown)


janbuchar:
> Only the platform knows the real forefront

Actually, the platform doesn't store that information either (it just uses it to put the request at the right end of the queue). And adding it to `userData` won't help, because `list(andLock)Head` doesn't return it (because it may be very large).

> do not try to keep imperfect copy of it locally and rather design the system to work without being aware of the cached forefront.

I'd actually prefer to just avoid solving the problem if we're using locking.

vdusek (approved):
Thanks :+1: .

PR #548, opened 2025-08-14
chore!: Update apify-client and apify-shared to v2.0

vdusek:


src/apify/_utils.py
```diff
@@ -81,3 +82,10 @@ def wrapper(func: Callable) -> Callable:
         return func
 
     return wrapper
+
+
+def maybe_extract_enum_member_value(maybe_enum_member: Any) -> Any:
```
Pijukatel, line 87: `maybe_extract_enum_member_value` is defined in client as well

Issue #536, opened 2025-08-07
Actor should use same `client_key` for all ApifyRequestQueue client calls


Pijukatel:
Apify platform should see all requests from same actor run with same `client_key`, unless explicitly done otherwise.
This should be migration-resistant key, meaning that it will be same key for one actor run even when actor is migrated, but it should be different for distinct aruns of same actor. Probable suitable candidate is RunID that is preserved during migration.

Locally, outside of Apify platform, just generate some random key.

The main benefit is that the flag `hadMultipleClients` of [request queues](https://docs.apify.com/api/v2/request-queue-get) will be more trustworthy.

It might be the root cause for 
https://github.com/apify/crawlee-python/issues/1131

Pijukatel:
This will also enable better estimation of RQ metadata. Knowing with better certainty if there is another client allows either to use the local estimation of metadata(if there is no other client) or use the metadata from platform if there are multiple clients.

PR #552, opened 2025-08-19
fix: Use same `client_key` for `Actor` created `request_queue` and improve its metadata estimation

Pijukatel:
### Description

- When creating `RequestQueue` from Actor on the platform or with `force_cloud=True` the `client_key` should be set to `run_id`. This ensures:
  - Each API call of the same `RequestQueue` instance that is using `ApifyRequestQueueClient` will be done with the same `client_key` and thus in metadata `had_multiple_clients=False`
  - Multiple instances of RequestQueue created by `Actor.open_request_queue()` on the platform share the same `client_key` and thus in metadata `had_multiple_clients=False`
  - On the platform, since the `client_key` is set to `run_id`, it remains the same for resurrected or migrated run, and thus in metadata `had_multiple_clients=False`
- Improved reliability of `had_multiple_clients` allows better estimation of `RequestQueue` metadata.
  - When `had_multiple_clients=False`, it is possible to trust local estimation of the metadata.
  - When `had_multiple_clients=True`, local estimation is no longer valid, but still can, in some cases, improve estimation by being ahead of the delayed API update of the metadata. Therefore API-based metadata are fused with local metadata estimation to produce as good estimation as we can.
- `ApifyRequestQueueClient` init changed to properly initialize from full metadata - to enable more reliable metadata after migration/resurrection or when using existing `RequestQueue`
- During `_list_head`, if there is a call to API, use the available `had_multiple_clients` to update local estimation of this value. This is a cheap way of knowing if there is another client or not without the need to make a new API call. `_list_head` is called frequently enough to make the local estimation of `had_multiple_clients` decently good.

### Issues

- Closes: #536

### Testing

- New tests added


tests/integration/conftest.py
```diff
@@ -96,6 +98,11 @@ def apify_token() -> str:
     return api_token
 
 
+@pytest.fixture(autouse=True)
+def set_token(apify_token: str, monkeypatch: pytest.MonkeyPatch) -> None:
+    monkeypatch.setenv(ApifyEnvVars.TOKEN, apify_token)
```
vdusek: Why do we need this now? How did it work before without it?

tests/integration/test_actor_request_queue.py
```diff
@@ -113,3 +102,163 @@ async def test_request_queue_is_finished(
...
+    await request_queue_force_cloud.fetch_next_request()
+
+    # Check that it is correctly in the RequestQueueClient metadata
+    assert (await request_queue_force_cloud.get_metadata()).had_multiple_clients is False
+
+    # Check that it is correctly in the API
+    api_client = apify_client_async.request_queue(request_queue_id=request_queue_force_cloud.id)
+    api_response = await api_client.get()
+    assert api_response
+    assert api_response['hadMultipleClients'] is False
```
vdusek: Should these be here? We have `test_actor_request_queue.py` and `test_request_queue.py` btw.

src/apify/storage_clients/_apify/_request_queue_client.py
```diff
@@ -70,40 +65,33 @@ def __init__(
...
     @override
     async def get_metadata(self) -> RequestQueueMetadata:
-        total_count = self._initial_total_count + self._assumed_total_count
-        handled_count = self._initial_handled_count + self._assumed_handled_count
-        pending_count = total_count - handled_count
-
+        """Get metadata about the request queue."""
+        response = await self._api_client.get()
+        if response is None:
+            raise ValueError('Failed to fetch request queue metadata from the API.')
```
vdusek: It should be better explained why we have `_get_metadata` and `get_metadata` now.

vdusek (changes requested):

src/apify/storage_clients/_apify/_request_queue_client.py
Pijukatel: Added explanatory docstring

tests/integration/test_actor_request_queue.py
Pijukatel: I have moved them. But frankly, I do not understand what the rule is. I see actor-based tests in both files.

tests/integration/conftest.py
Pijukatel: Updated

tests/integration/test_actor_request_queue.py
vdusek: Hmm, okay, maybe we should merge those two files into one.

src/apify/storage_clients/_apify/_request_queue_client.py
```diff
@@ -70,40 +65,38 @@ def __init__(
...
         self._fetch_lock = asyncio.Lock()
         """Fetch lock to minimize race conditions when communicating with API."""
 
+    async def _get_metadata(self) -> RequestQueueMetadata:
+        """Try to get cached metadata first. If multiple clients, fuse with global metadata.
+
+        This method is used internally to avoid unnecessary API call unless needed (multiple clients).
+        Local estimation of metadata is without delay, unlike metadata from API. In situation where there is only one
+        client, it is the better choice.
+        """
```
vdusek: I would distinguish those by names as well, my suggestion (`get_metadata_cached` / `get_metadata_estimate`):

```suggestion
    async def _get_metadata_cached(self) -> RequestQueueMetadata:
        """Try to get cached metadata first. If multiple clients, fuse with global metadata.

        This method is used internally to avoid unnecessary API call unless needed (multiple clients).
        Local estimation of metadata is without delay, unlike metadata from API. In situation where there is only one
        client, it is the better choice.
        """
```

src/apify/storage_clients/_apify/_request_queue_client.py
```diff
@@ -70,40 +65,38 @@ def __init__(
...
+        # Get local estimation (will not include changes done bo another client)
+        return self._metadata
+
     @override
     async def get_metadata(self) -> RequestQueueMetadata:
-        total_count = self._initial_total_count + self._assumed_total_count
-        handled_count = self._initial_handled_count + self._assumed_handled_count
-        pending_count = total_count - handled_count
-
+        """Get metadata about the request queue."""
```
vdusek:
```suggestion
        """Get metadata about the request queue.
        
        Returns metadata from the API, merged with local estimation, because in some cases, 
        the data from the API can be delayed.
        """
```

tests/integration/test_actor_request_queue.py
Pijukatel: Some test cleanup and reordering might be good, but lets not do it in this PR.

src/apify/storage_clients/_apify/_request_queue_client.py
Pijukatel: Updated

src/apify/storage_clients/_apify/_request_queue_client.py
Pijukatel: I like the `_get_metadata_estimate`. Updated

vdusek (approved):
Looks good, let's see how it behaves...

I'm just not sure about the `feat:` :slightly_smiling_face:, maybe rather a fix?

PR #569, opened 2025-08-26
docs: Add shields to README

janbuchar:


README.md
```diff
@@ -1,5 +1,20 @@
 # Apify SDK for Python
 
+<p align=center>
```
janbuchar, line 3: I guess this would look better aligned to the left?

vdusek: I don't know, in Crawlee it looks good that everything is aligned to the center:

image

I would do the same here.

janbuchar: That means centering the heading as well then

Pijukatel (approved):

Issue #513, opened 2025-07-24
Add specialized/parametrized RequestQueueClients


Pijukatel:
**Context**
Default `RequestQueue` offers many features apart from request managment, like the possibility of multiple producers/consumers working with it, real-time persistence, and integration with console. To support all of it, it comes with some costs - actual financial costs to use request queue and increased network traffic as the request queue is sending many API calls related to request handling and locking. Some use cases do not need those additional features of the request queue and thus are only affected by the costs with no benefit.

**Proposed solution**

- Add single consumer variant of the `RequestQueue` - this will drop the locking mechanism of the RequestQueue and reduce costs and network traffic, while preserving all other benefits of  `RequestQueue`

- Add new specialized `RequestQueue` client that will not be based on platform-backed `RequestQueue` to even further reduce the costs. Possibly just extend `MemoryRequestQueueClient` with persistence capabilities to KVS. Use case is single consumer/producer request queue with on-demand persistence that is massively cheaper compared to platform-backed `RequestQueue`. - For example [this POC ](https://github.com/apify/apify-sdk-python/blob/8a3d28d99dc83b8726632b6324d69085a4d863b7/src/apify/storage_clients/hybrid/_key_value_store_client.py)

- Update documentation and describe all the variants of `RequestQueue`, their benefits, drawbacks, and recommended use cases. Also mention crawlee-based `RequestQueue` clients that still have some niche use case even on Apify platform.

PR #573, opened 2025-08-27
feat: Add specialized ApifyRequestQueue clients

Pijukatel:
### Description

- `ApifyRequestQueueClient` can be created in two access modes - `single`, `shared`:
  - `shared` - current version that supports multiple producers/consumers and locking of requests. More Apify API calls, higher API usage -> more expensive, slower.
  - `single` - new constrained client for self-consumer and multiple constrained producers. (Detailed constraints in the docs). Fewer Apify API calls, lower API usage  -> cheaper, faster.
 
- Most of the `ApifyRequestQueueClient` tests were moved away from actor-based tests, so that they can be parametrized for both variants of the `ApifyRequestQueueClients` and to make local debugging easier.

#### Usage:
RequestQueue with `shared`:
`await RequestQueue.open(storage_client=ApifyStorageClient(request_queue_access="shared"))`
RequestQueue with default `single`:
`await RequestQueue.open(storage_client=ApifyStorageClient())`

#### Stats difference:
The full client is doing significantly more API calls and regarding the API usage it is doing 50% more RequestQueue writes and also more RequestQueue reads.

**Example rq related stats for crawler started with 1000 requests:**
`shared`:
API calls: 2123
API usage: {'readCount': 1000, 'writeCount': 3000, 'deleteCount': 0, 'headItemReadCount': 0, 'storageBytes': 104035}

`single`:
API calls: 1059
API usage: {'readCount': 3, 'writeCount': 2000, 'deleteCount': 0, 'headItemReadCount': 14, 'storageBytes': 103826}

### Issues

- Part of: #513


src/apify/storage_clients/_apify/_request_queue_client_simple.py
```diff
@@ -0,0 +1,403 @@
...
+
+class ApifyRequestQueueClientSimple(ApifyRequestQueueClient):
+    """An Apify platform implementation of the request queue client with limited capability.
+
+    This client is designed to use as little resources as possible, but has to be used in constrained context.
+    Constraints:
+    - Only one client is consuming the request queue at the time.
+    - Multiple producers can put requests to the queue, but their forefront requests are not guaranteed to be handled
+      so quickly as this client does not aggressively fetch the forefront and relies on local head estimation.
+    - Requests are only added to the queue, never deleted. (Marking as handled is ok.)
```
vdusek, line 32: > Requests are only added to the queue, never deleted. (Marking as handled is ok.)

? I don't get it.

src/apify/storage_clients/_apify/_request_queue_client_simple.py
```diff
@@ -0,0 +1,403 @@
...
+    - Other producers can add new requests, but not modify existing ones (otherwise caching can miss the updates)
```
vdusek, line 33: > Other producers can add new requests, but not modify existing ones (otherwise caching can miss the updates)

Modify existing ones? What do you mean by that?

docs/04_upgrading/upgrading_to_v3.md
```diff
@@ -52,3 +52,21 @@ async def main():
...
+```python
+from apify.storages import RequestQueue
+from apify.storage_clients import ApifyStorageClient
+
+async def main():
+    # Full client
+    rq_full = await RequestQueue.open(storage_client=ApifyStorageClient(simple_request_queue=False))
+    # Default optimized client
+    rq_simple = await RequestQueue.open(storage_client=ApifyStorageClient())
+```
```
vdusek, line 159: We should use an example with Actor here.

tests/integration/conftest.py
```diff
@@ -50,6 +51,9 @@ def prepare_test_env(monkeypatch: pytest.MonkeyPatch, tmp_path: Path) -> Callabl
     """
 
     def _prepare_test_env() -> None:
+        # Reset the Actor class state.
+        apify._actor.Actor.__wrapped__.__class__._is_any_instance_initialized = False  # type: ignore[attr-defined]
+        apify._actor.Actor.__wrapped__.__class__._is_rebooting = False  # type: ignore[attr-defined]
```
vdusek, line 56: Why do we need this now?

src/apify/storage_clients/_apify/_storage_client.py
```diff
@@ -18,11 +19,24 @@
...
 
+    def __init__(self, *, simple_request_queue: bool = True) -> None:
+        """Initialize the Apify storage client.
+
+        Args:
+            simple_request_queue: If True, the `create_rq_client` will always return `ApifyRequestQueueClientSimple`,
+                if false it will return `ApifyRequestQueueClientFull`. Simple client is suitable for single consumer
+                scenarios and makes less API calls. Full client is suitable for multiple consumers scenarios at the
+                cost of higher API usage
+        """
```
vdusek: As we discussed on Slack, I'm more for:

```python
def __init__(self, *, access: Literal['single', 'shared' = 'single') -> None:
```

also taking into account potential RQ v3...

src/apify/storage_clients/_apify/_storage_client.py
```diff
@@ -8,7 +8,8 @@
 
 from ._dataset_client import ApifyDatasetClient
 from ._key_value_store_client import ApifyKeyValueStoreClient
-from ._request_queue_client import ApifyRequestQueueClient
+from ._request_queue_client_full import ApifyRequestQueueClientFull
+from ._request_queue_client_simple import ApifyRequestQueueClientSimple
```
vdusek: And the naming of these internal classes...

If we go with the access literal parameter, maybe we can use:
- `ApifyRequestQueueSingleClient`
- `ApifyRequestQueueSharedClient`

tests/integration/conftest.py
```diff
@@ -103,14 +107,15 @@ def apify_client_async(apify_token: str) -> ApifyClientAsync:
     return ApifyClientAsync(apify_token, api_url=api_url)
 
 
-@pytest.fixture
-async def request_queue_force_cloud(apify_token: str, monkeypatch: pytest.MonkeyPatch) -> AsyncGenerator[RequestQueue]:
+@pytest.fixture(params=[False, True])
+async def default_request_queue_apify(
```
vdusek: why not just `request_queue_apify`?

tests/integration/test_request_queue.py
vdusek, line 1: Wait, you are removing the Actor tests (!?).

With the current framework, the integration tests are full end-to-end tests running on the platform as Actors.

What you're proposing changes that (!).

vdusek (dismissed):
high-level things

tests/integration/test_request_queue.py
Pijukatel: I think there are several levels of integration tests.
1. level - API
- These are, for example, RequestQueue specific tests that do not mock API and are doing real API calls.
2. level - Platform
- Actor tests that are integration of Apify platform and also API
3. level - Crawlee on Platform
- Crawler running in Actor. Test integration of Crawlee with Apify Platform and API

What I have done to most RQ specific tests is to move them from level 2 and 3 to level 1. I think we should  have more tests on lower levels and fewer tests on higher levels

src/apify/storage_clients/_apify/_request_queue_client_simple.py
Pijukatel: Well API has delete endpoint. We do not expose it in RQ, but if someone is calling it while we use this client to work on that RQ, then it will have unpredictable behavior.
https://docs.apify.com/api/v2/request-queue-request-delete

It is not a normal use case, but better to be explicit about it

src/apify/storage_clients/_apify/_request_queue_client_simple.py
Pijukatel: API has update endpoint. If someone (other producers) are updating existing requests and this client has already cached the requests locally, then the client will use the outdated request. 
https://docs.apify.com/api/v2/request-queue-request-put

It is not a normal use case, but better to be explicit about it

docs/04_upgrading/upgrading_to_v3.md
Pijukatel: I thought about that, but it is just wrapping it with Actor:  
I can add it, but it felt like pointless clutter

tests/integration/conftest.py
Pijukatel: We don't need it, but I saw a warning log in some tests and realized that we do not isolate the tests so well because `is_any_instance_initialized` and `_is_rebooting` were leaking from the previous tests

This warning could be observed in tests
`WARN  Repeated Actor initialization detected - this is non-standard usage, proceed with care`

tests/integration/conftest.py
Pijukatel: Ok

src/apify/storage_clients/_apify/_storage_client.py
Pijukatel: Done

src/apify/storage_clients/_apify/_storage_client.py
Pijukatel: Done

docs/04_upgrading/upgrading_to_v3.md
```diff
@@ -52,3 +52,21 @@ async def main():
...
+
+**Now (v3.0):**
+
+```python
+from apify.storages import RequestQueue
+from apify.storage_clients import ApifyStorageClient
+
+async def main():
+   # Full client that supports multiple consumers of the Apify Request Queue
+    rq_shared = await RequestQueue.open(storage_client=ApifyStorageClient(access="shared"))
```
janbuchar: This will set the storage client both for local runs and Apify platform runs. This is usually not the desired behavior. We should instead recommend a way of setting this up so that it uses a shared request queue on Apify and filesystem locally.

docs/04_upgrading/upgrading_to_v3.md
janbuchar: I prefer a bit of clutter if it means that we prevent some confusion. We don't want to make someone think that maybe there are cases where you don't need to init your Actor.

src/apify/storage_clients/_apify/_request_queue_client.py
```diff
@@ -62,10 +49,7 @@ def unique_key_to_request_id(unique_key: str, *, request_id_length: int = 15) ->
 
 
 class ApifyRequestQueueClient(RequestQueueClient):
-    """An Apify platform implementation of the request queue client."""
-
-    _DEFAULT_LOCK_TIME: Final[timedelta] = timedelta(minutes=3)
-    """The default lock time for requests in the queue."""
+    """Base class for Apify platform implementations of the request queue client."""
```
janbuchar, line 31: I'm pretty sure that we don't want to do it like this. Using inheritance to share code makes it very hard to understand the difference between the two child classes - I'm speaking from experience with the JS counterpart which does the same and it's confusing as hell.

I'd suggest either having a single class with several if-else blocks, or if that proves too spaghettific, some approach based on composition (strategy pattern?)

src/apify/storage_clients/_apify/_storage_client.py
```diff
@@ -18,11 +19,24 @@
...
     from crawlee.configuration import Configuration as CrawleeConfiguration
 
+    from ._request_queue_client import ApifyRequestQueueClient
+
 
 @docs_group('Storage clients')
 class ApifyStorageClient(StorageClient):
     """Apify storage client."""
 
+    def __init__(self, *, access: Literal['single', 'shared'] = 'single') -> None:
```
janbuchar: `access` sounds too broad. How about `request_queue_access`?

tests/integration/conftest.py
```diff
@@ -103,14 +107,15 @@ def apify_client_async(apify_token: str) -> ApifyClientAsync:
...
+async def request_queue_apify(
+    apify_token: str, monkeypatch: pytest.MonkeyPatch, request: pytest.FixtureRequest
+) -> AsyncGenerator[RequestQueue]:
     """Create an instance of the Apify request queue on the platform and drop it when the test is finished."""
-    request_queue_name = generate_unique_resource_name('request_queue')
     monkeypatch.setenv(ApifyEnvVars.TOKEN, apify_token)
 
     async with Actor:
-        rq = await Actor.open_request_queue(name=request_queue_name, force_cloud=True)
+        rq = await RequestQueue.open(storage_client=ApifyStorageClient(access=request.param))
```
janbuchar: Please put the `yield` inside a `finally` block so that we minimize the chance of leaving a dangling RQ.

tests/integration/conftest.py
```diff
@@ -103,14 +107,15 @@ def apify_client_async(apify_token: str) -> ApifyClientAsync:
...
 
 
-@pytest.fixture
-async def request_queue_force_cloud(apify_token: str, monkeypatch: pytest.MonkeyPatch) -> AsyncGenerator[RequestQueue]:
+@pytest.fixture(params=['single', 'shared'])
+async def request_queue_apify(
+    apify_token: str, monkeypatch: pytest.MonkeyPatch, request: pytest.FixtureRequest
+) -> AsyncGenerator[RequestQueue]:
     """Create an instance of the Apify request queue on the platform and drop it when the test is finished."""
-    request_queue_name = generate_unique_resource_name('request_queue')
```
janbuchar, line 109: Will this now work magically thanks to aliases? If not, there might be concurrency-related issues

tests/integration/test_request_queue.py
janbuchar: I don't mind the tests calling the RQ directly without wrapping it in a container, as long as the logic is preserved.

tests/integration/conftest.py
Pijukatel: Yes. Each test using this fixture will create a new, unnamed RQ on the platform (through the alias mechanism). The platform will give a unique ID to each such RQ, and so there should be no concurrency-related problems.

It requires this to be merged, though
https://github.com/apify/apify-sdk-python/pull/606

tests/integration/conftest.py
Pijukatel: Could you please describe the scenario it solves?

(Btw. Now that unnamed storages will be used in tests, the platform will not keep them forever even if they somehow leak the test.)

src/apify/storage_clients/_apify/_request_queue_client.py
Pijukatel: I think this is not the case of hard-to-understand inheritance. These two classes inherit from "intermediate base class" only the identical methods. The reader can then focus on the implementations specific to the specialized class. There is no overriding of the intermediate class, so there should be no confusion.

src/apify/storage_clients/_apify/_storage_client.py
Pijukatel: Ok, renamed.

docs/04_upgrading/upgrading_to_v3.md
Pijukatel: Extracted storage-related Actor methods to ApifyHybridStorageClient. Now there is full support for many rare edge cases + reasonable default.

tests/integration/test_apify_storages.py
```diff
@@ -32,3 +33,91 @@ async def test_alias_concurrent_creation_local(
...
+
+async def test_actor_full_explicit_storage_init(apify_token: str) -> None:
+    service_locator.set_configuration(Configuration(token=apify_token))
+    service_locator.set_storage_client(
+        ApifyHybridStorageClient(
+            local_storage_client=MemoryStorageClient(),
+            cloud_storage_client=ApifyStorageClient(access='shared'),
+        )
+    )
+    async with Actor():
```
janbuchar: I sure hope that the parentheses aren't required here :slightly_smiling_face:

src/apify/storage_clients/_hybrid_apify/_storage_client.py
```diff
@@ -0,0 +1,226 @@
...
+from apify.storages import Dataset, KeyValueStore
+
+if TYPE_CHECKING:
+    from collections.abc import Hashable
+
+    from crawlee.configuration import Configuration as CrawleeConfiguration
+
+
+@docs_group('Storage clients')
+class ApifyHybridStorageClient(StorageClient):
```
janbuchar: Not sure about the name - after all, this is the default behavior, and "hybrid" does not sound like something that should be a default. Let's try to think of something better. Maybe we could even toot our own horn with something like `SmartApifyStorageClient`.

src/apify/storage_clients/_hybrid_apify/_storage_client.py
```diff
@@ -0,0 +1,226 @@
...
+        """
+        return await RequestQueue.open(
+            id=id,
+            alias=alias,
+            name=name,
+            configuration=configuration,
+            storage_client=self._get_suitable_storage_client(force_cloud=force_cloud),
+        )
+
+    @cached_property
```
janbuchar: I'd skip the caching here. IMO the subtle bugs it could introduce are not worth the tiny performance gain

src/apify/_actor.py
```diff
@@ -234,45 +233,49 @@ def log(self) -> logging.Logger:
...
+    @cached_property
+    def _storage_client(self) -> ApifyHybridStorageClient:
+        """Storage client used by the actor.
 
+        Depending on the initialization of the service locator the client can be created in different ways.
+        """
         try:
-            # Set implicit default local storage client, unless local storage client was already set.
-            implicit_storage_client = ApifyFileSystemStorageClient()
+            # Notning was set by the user.
```
janbuchar:
```suggestion
            # Nothing was set by the user.
```

src/apify/_actor.py
```diff
@@ -234,45 +233,49 @@ def log(self) -> logging.Logger:
...
 
-        self._local_storage_client = service_locator.get_storage_client()
-        if type(self._local_storage_client) is FileSystemStorageClient:
+        # User set something in the service locator.
+        storage_client = service_locator.get_storage_client()
+        if isinstance(storage_client, ApifyHybridStorageClient):
+            # The client was manually set to the right type in the service locator. This is the explicit way.
+            return storage_client
+
+        if isinstance(storage_client, ApifyStorageClient):
```
janbuchar: So if I do `service_locator.set_cstorage_client(ApifyStorageClient())` before Actor init, this will force using the filesystem locally? That doesn't sound too desirable.

src/apify/storage_clients/_apify/_request_queue_client.py
janbuchar: > There is no overriding of the intermediate class, so there should be no confusion.

it isn't there now, who knows, what the future will bring... please consider the alternatives I suggested

src/apify/storage_clients/_hybrid_apify/_storage_client.py
```diff
@@ -0,0 +1,226 @@
...
+        id: str | None = None,
+        name: str | None = None,
+        alias: str | None = None,
+        configuration: CrawleeConfiguration | None = None,
+    ) -> RequestQueueClient:
+        return await self._get_suitable_storage_client().create_rq_client(
+            id=id, name=id, alias=alias, configuration=configuration
+        )
+
+    async def open_dataset(
```
janbuchar: These three methods are super internal, it'd be great to mark them as such. The current docblocks make them look like something intended for the users to call

src/apify/_actor.py
Pijukatel: Well, at this point, it is a guess what the user wants. Only a fully explicit setting of the `SmartApifyStorageClient` tells us what the user is really trying to do. We can guess in the other direction or throw an exception and allow only one of the two options: a fully implicit(default) or fully explicit client (SmartApifyStorageClient).

It is kind of an edge case, so I am fine with any of those. If you have a strong preference, I will do it that way.

src/apify/storage_clients/_hybrid_apify/_storage_client.py
Pijukatel: Well, I was not thinking about the performance. More like, this should never change during runtime. If it changes during runtime, then a lot of assumptions made in the code are no longer valid.

janbuchar: Agreed. But caching the value only further obscures the problem, if one appears

src/apify/_actor.py
janbuchar: In this particular case, I would throw instead of guessing, but include detailed instructions for setting a custom client.

src/apify/storage_clients/_hybrid_apify/_storage_client.py
Pijukatel: Ok, I think they are actually not needed in the end. The Actor can call `SmartApifyStorageClient.get_suitable_storage_client` directly

tests/integration/conftest.py
janbuchar: If the test throws an error, it will be propagated through that `yield`. For that reason, it's better to put cleanup in a `finally` block. I agree that in this case it's unlikely that it'd cause a serious issue, but there's no good reason not to add that block.

Pijukatel: I am still not sure I follow. The pytest fixture after the yield will run even if there was an exception in the test. It will not run if there was an exception in the fixture code itself, but if there was an exception during rq creation, then it was not created and there is nothing to clean. Or do I miss something?

tests/integration/test_apify_storages.py
Pijukatel: Not required. Removed

src/apify/storage_clients/_hybrid_apify/_storage_client.py
Pijukatel: Ok, renamed

src/apify/storage_clients/_hybrid_apify/_storage_client.py
Pijukatel: Ok, removed

src/apify/_actor.py
Pijukatel: Ok, throwing

src/apify/storage_clients/_apify/_request_queue_client.py
Pijukatel: Ok, using composition instead

src/apify/storage_clients/_apify/_request_queue_single_client.py
```diff
@@ -0,0 +1,396 @@
...
+            request: The request to mark as handled.
+
+        Returns:
+            Information about the queue operation. `None` if the given request was not in progress.
+        """
+        # Set the handled_at timestamp if not already set
+
+        if request.handled_at is None:
+            request.handled_at = datetime.now(tz=timezone.utc)
+            self.metadata.handled_request_count += 1
```
Mantisus, line 274: The value of `pending_request_count` in `metadata` is not updated anywhere. Only `total_request_count` and `handled_request_count` are updated. Is this correct?

Mantisus (commented):
I like it.

Just one question.

src/apify/storage_clients/_apify/_request_queue_single_client.py
Pijukatel: Thanks. I added tracking of pending requests in local metadata estimation.

PR #574, opened 2025-08-27
feat: Add stats to `ApifyRequestQueueClient`

Pijukatel:
### Description

- Add specialized `ApifyRequestQueueMetadata` that contains Apify-specific `RequestQueueStats`
- This enhances metadata of `ApifyRequestQueueClient`

### Issues

- Closes: https://github.com/apify/crawlee-python/issues/1344


tests/integration/test_request_queue.py
```diff
@@ -1278,3 +1279,26 @@ async def test_request_queue_not_had_multiple_clients(
...
+    for _ in range(read_request_count):
+        await request_queue_force_cloud.get_request(Request.from_url('http://example.com/1').unique_key)
+
+    # Wait for stats to become stable
+    await asyncio.sleep(10)
+
+    metadata = await request_queue_force_cloud.get_metadata()
+
+    assert hasattr(metadata, 'stats')
+    apify_metadata = cast('ApifyRequestQueueMetadata', metadata)
```
vdusek, line 1302: I don't see much value in this, since typing casts are required anyway, instead of simply allowing extra fields in the Pydantic model, or omitting this altogether and using the Apify client directly.

Pijukatel: This is more about showing that `force_cloud=True` created RQ will have the stats. If that is the case, then it is implicitly proven that even the client has them.

vdusek: Sorry, I meant this whole PR, as we discussed on Slack [here](https://apify.slack.com/archives/C06ENBV1B9Q/p1755854534018019).

Pijukatel: We already include some [Apify specific field for the Apify based kvs in specialized metadata](https://github.com/apify/apify-sdk-python/blob/master/src/apify/storage_clients/_apify/_models.py#L15), so nothing new in the code base.

vdusek: Okay

PR #576, opened 2025-09-02
refactor!: Test with crawlee branch storage-clients-and-configurations-2

Pijukatel:
This PR is for testing in CI.
It is SDK related changes for the Crawlee branch storage-clients-and-configurations-2

TODO: Create a proper way of patching global service_locator to return Apify Configuration and not Crawlee configuration. Currently working hack is implemented, not very nice.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,7 +36,7 @@ keywords = [
 dependencies = [
     "apify-client>=2.0.0,<3.0.0",
     "apify-shared>=2.0.0,<3.0.0",
-    "crawlee==1.0.0rc1",
+    "crawlee @ git+https://github.com/apify/crawlee-python.git@storage-clients-and-configurations",
     "cachetools>=5.5.0",
     "cryptography>=42.0.0",
     # TODO: ensure compatibility with the latest version of lazy-object-proxy
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,7 +36,7 @@ keywords = [
 dependencies = [
     "apify-client>=2.0.0,<3.0.0",
     "apify-shared>=2.0.0,<3.0.0",
-    "crawlee @ git+https://github.com/apify/crawlee-python.git@storage-clients-and-configurations",
+    "crawlee @ git+https://github.com/apify/crawlee-python.git@storage-clients-and-configurations-2",
     "cachetools>=5.5.0",
     "cryptography>=42.0.0",
     # TODO: ensure compatibility with the latest version of lazy-object-proxy
--- a/src/apify/_actor.py
+++ b/src/apify/_actor.py
@@ -128,10 +128,6 @@ class _ActorType:
         self._configure_logging = configure_logging
         self._apify_client = self.new_client()
 
-        # Create an instance of the cloud storage client, the local storage client is obtained
-        # from the service locator.
-        self._cloud_storage_client = ApifyStorageClient()
-
         # Set the event manager based on whether the Actor is running on the platform or locally.
         self._event_manager = (
             ApifyEventManager(
@@ -281,6 +277,10 @@ class _ActorType:
 
         if _ActorType._is_any_instance_initialized:
             self.log.warning('Repeated Actor initialization detected - this is non-standard usage, proceed with care')
+
+        # Create an instance of the cloud storage client, the local storage client is obtained
+        # from the service locator
+        self._cloud_storage_client = ApifyStorageClient(configuration=self.configuration)
 
         # Make sure that the currently initialized instance is also available through the global `Actor` proxy
         cast('Proxy', Actor).__wrapped__ = self
--- a/src/apify/_actor.py
+++ b/src/apify/_actor.py
@@ -429,7 +429,6 @@ class _ActorType:
         return await Dataset.open(
             id=id,
             name=name,
-            configuration=self._configuration,
             storage_client=storage_client,
         )
 
@@ -463,7 +462,6 @@ class _ActorType:
         return await KeyValueStore.open(
             id=id,
             name=name,
-            configuration=self._configuration,
             storage_client=storage_client,
         )
 
@@ -500,7 +498,6 @@ class _ActorType:
         return await RequestQueue.open(
             id=id,
             name=name,
-            configuration=self._configuration,
             storage_client=storage_client,
         )
 
--- a/src/apify/_actor.py
+++ b/src/apify/_actor.py
@@ -13,6 +13,7 @@ from pydantic import AliasChoices
 
 from apify_client import ApifyClientAsync
 from apify_shared.consts import ActorEnvVars, ActorExitCodes, ApifyEnvVars
+from crawlee.errors import ServiceConflictError
 from crawlee.events import (
     Event,
     EventAbortingData,
@@ -118,9 +119,12 @@ class _ActorType:
         self._exit_process = self._get_default_exit_process() if exit_process is None else exit_process
         self._is_exiting = False
 
-        if configuration:
-            service_locator.set_configuration(configuration)
-        self._configuration = service_locator.get_configuration()
+        # Actor state when this method is being executed is unpredictable.
+        # Actor can be initialized by lazy object proxy or by user directly, or by both.
+        # Until `init` method is run, this state of uncertainty remains. This is the reason why any setting done here in
+        # `__init__` method should not be considered final.
+
+        self._configuration = configuration
         self._configure_logging = configure_logging
         self._apify_client = self.new_client()
 
@@ -131,17 +135,17 @@ class _ActorType:
         # Set the event manager based on whether the Actor is running on the platform or locally.
         self._event_manager = (
             ApifyEventManager(
-                configuration=self._configuration,
-                persist_state_interval=self._configuration.persist_state_interval,
+                configuration=self.config,
+                persist_state_interval=self.config.persist_state_interval,
             )
             if self.is_at_home()
             else LocalEventManager(
-                system_info_interval=self._configuration.system_info_interval,
-                persist_state_interval=self._configuration.persist_state_interval,
+                system_info_interval=self.config.system_info_interval,
+                persist_state_interval=self.config.persist_state_interval,
             )
         )
 
-        self._charging_manager = ChargingManagerImplementation(self._configuration, self._apify_client)
+        self._charging_manager = ChargingManagerImplementation(self.config, self._apify_client)
 
         self._is_initialized = False
 
@@ -204,12 +208,18 @@ class _ActorType:
     @property
     def configuration(self) -> Configuration:
         """The Configuration instance the Actor instance uses."""
-        return self._configuration
+        return self.config
 
     @property
     def config(self) -> Configuration:
         """The Configuration instance the Actor instance uses."""
-        return self._configuration
+        if self._configuration:
+            return self._configuration
+        self.log.debug(
+            'Implicit configuration used.'
+            "It's recommended to explicitly set the configuration to avoid unexpected behavior."
+        )
+        return Configuration()
 
     @property
     def event_manager(self) -> EventManager:
@@ -251,6 +261,21 @@ class _ActorType:
         This method should be called immediately before performing any additional Actor actions, and it should be
         called only once.
         """
+        if self._configuration:
+            # Set explicitly the configuration in the service locator
+            service_locator.set_configuration(self.configuration)
+        else:
+            try:
+                # Set implicit default Apify configuration, unless configuration was already set.
+                service_locator.set_configuration(self.configuration)
+            except ServiceConflictError:
+                self.log.info(
+                    'Configuration in service locator was set explicitly before Actor. '
+                    'Using the existing configuration.'
+                )
+            # Use the configuration from the service locator
+            self._configuration = service_locator.get_configuration()
+
         if self._is_initialized:
             raise RuntimeError('The Actor was already initialized!')
 
@@ -268,7 +293,6 @@ class _ActorType:
             service_locator.set_storage_client(self._cloud_storage_client)
 
         service_locator.set_event_manager(self.event_manager)
-        service_locator.set_configuration(self.configuration)
 
         # The logging configuration has to be called after all service_locator set methods.
         if self._configure_logging:
@@ -386,8 +410,8 @@ class _ActorType:
                 (increases exponentially from this value).
             timeout: The socket timeout of the HTTP requests sent to the Apify API.
         """
-        token = token or self._configuration.token
-        api_url = api_url or self._configuration.api_base_url
+        token = token or self.config.token
+        api_url = api_url or self.config.api_base_url
         return ApifyClientAsync(
             token=token,
             api_url=api_url,
@@ -547,9 +571,9 @@ class _ActorType:
         """Get the Actor input value from the default key-value store associated with the current Actor run."""
         self._raise_if_not_initialized()
 
-        input_value = await self.get_value(self._configuration.input_key)
-        input_secrets_private_key = self._configuration.input_secrets_private_key_file
-        input_secrets_key_passphrase = self._configuration.input_secrets_private_key_passphrase
+        input_value = await self.get_value(self.config.input_key)
+        input_secrets_private_key = self.config.input_secrets_private_key_file
+        input_secrets_key_passphrase = self.config.input_secrets_private_key_passphrase
         if input_secrets_private_key and input_secrets_key_passphrase:
             private_key = load_private_key(
                 input_secrets_private_key,
@@ -686,7 +710,7 @@ class _ActorType:
 
     def is_at_home(self) -> bool:
         """Return `True` when the Actor is running on the Apify platform, and `False` otherwise (e.g. local run)."""
-        return self._configuration.is_at_home
+        return self.config.is_at_home
 
     def get_env(self) -> dict:
         """Return a dictionary with information parsed from all the `APIFY_XXX` environment variables.
@@ -712,7 +736,7 @@ class _ActorType:
                 aliases = [field_name]
 
             for alias in aliases:
-                config[alias] = getattr(self._configuration, field_name)
+                config[alias] = getattr(self.config, field_name)
 
         env_vars = {env_var.value.lower(): env_var.name.lower() for env_var in [*ActorEnvVars, *ApifyEnvVars]}
         return {option_name: config[env_var] for env_var, option_name in env_vars.items() if env_var in config}
@@ -1000,13 +1024,13 @@ class _ActorType:
             return
 
         if not custom_after_sleep:
-            custom_after_sleep = self._configuration.metamorph_after_sleep
+            custom_after_sleep = self.config.metamorph_after_sleep
 
         # If is_at_home() is True, config.actor_run_id is always set
-        if not self._configuration.actor_run_id:
+        if not self.config.actor_run_id:
             raise RuntimeError('actor_run_id cannot be None when running on the Apify platform.')
 
-        await self._apify_client.run(self._configuration.actor_run_id).metamorph(
+        await self._apify_client.run(self.config.actor_run_id).metamorph(
             target_actor_id=target_actor_id,
             run_input=run_input,
             target_actor_build=target_actor_build,
@@ -1043,7 +1067,7 @@ class _ActorType:
         _ActorType._is_rebooting = True
 
         if not custom_after_sleep:
-            custom_after_sleep = self._configuration.metamorph_after_sleep
+            custom_after_sleep = self.config.metamorph_after_sleep
 
         # Call all the listeners for the PERSIST_STATE and MIGRATING events, and wait for them to finish.
         # PERSIST_STATE listeners are called to allow the Actor to persist its state before the reboot.
@@ -1063,10 +1087,10 @@ class _ActorType:
             *[listener(EventMigratingData()) for listener in migrating_listeners],
         )
 
-        if not self._configuration.actor_run_id:
+        if not self.config.actor_run_id:
             raise RuntimeError('actor_run_id cannot be None when running on the Apify platform.')
 
-        await self._apify_client.run(self._configuration.actor_run_id).reboot()
+        await self._apify_client.run(self.config.actor_run_id).reboot()
 
         if custom_after_sleep:
             await asyncio.sleep(custom_after_sleep.total_seconds())
@@ -1105,11 +1129,11 @@ class _ActorType:
             return
 
         # If is_at_home() is True, config.actor_run_id is always set
-        if not self._configuration.actor_run_id:
+        if not self.config.actor_run_id:
             raise RuntimeError('actor_run_id cannot be None when running on the Apify platform.')
 
         await self._apify_client.webhooks().create(
-            actor_run_id=self._configuration.actor_run_id,
+            actor_run_id=self.config.actor_run_id,
             event_types=webhook.event_types,
             request_url=webhook.request_url,
             payload_template=webhook.payload_template,
@@ -1141,10 +1165,10 @@ class _ActorType:
             return None
 
         # If is_at_home() is True, config.actor_run_id is always set
-        if not self._configuration.actor_run_id:
+        if not self.config.actor_run_id:
             raise RuntimeError('actor_run_id cannot be None when running on the Apify platform.')
 
-        api_result = await self._apify_client.run(self._configuration.actor_run_id).update(
+        api_result = await self._apify_client.run(self.config.actor_run_id).update(
             status_message=status_message, is_status_message_terminal=is_terminal
         )
 
@@ -1199,7 +1223,7 @@ class _ActorType:
             country_code=country_code,
             proxy_urls=proxy_urls,
             new_url_function=new_url_function,
-            _actor_config=self._configuration,
+            _actor_config=self.config,
             _apify_client=self._apify_client,
         )
 
--- a/src/apify/_actor.py
+++ b/src/apify/_actor.py
@@ -13,7 +13,6 @@ from pydantic import AliasChoices
 
 from apify_client import ApifyClientAsync
 from apify_shared.consts import ActorEnvVars, ActorExitCodes, ApifyEnvVars
-from crawlee import service_locator
 from crawlee.events import (
     Event,
     EventAbortingData,
@@ -25,7 +24,7 @@ from crawlee.events import (
 )
 
 from apify._charging import ChargeResult, ChargingManager, ChargingManagerImplementation
-from apify._configuration import Configuration
+from apify._configuration import Configuration, service_locator
 from apify._consts import EVENT_LISTENERS_TIMEOUT
 from apify._crypto import decrypt_input_secrets, load_private_key
 from apify._models import ActorRun
@@ -119,7 +118,9 @@ class _ActorType:
         self._exit_process = self._get_default_exit_process() if exit_process is None else exit_process
         self._is_exiting = False
 
-        self._configuration = configuration or Configuration.get_global_configuration()
+        if configuration:
+            service_locator.set_configuration(configuration)
+        self._configuration = service_locator.get_configuration()
         self._configure_logging = configure_logging
         self._apify_client = self.new_client()
 
--- a/src/apify/_actor.py
+++ b/src/apify/_actor.py
@@ -13,6 +13,7 @@ from pydantic import AliasChoices
 
 from apify_client import ApifyClientAsync
 from apify_shared.consts import ActorEnvVars, ActorExitCodes, ApifyEnvVars
+from crawlee import service_locator
 from crawlee.errors import ServiceConflictError
 from crawlee.events import (
     Event,
@@ -25,7 +26,7 @@ from crawlee.events import (
 )
 
 from apify._charging import ChargeResult, ChargingManager, ChargingManagerImplementation
-from apify._configuration import Configuration, service_locator
+from apify._configuration import Configuration
 from apify._consts import EVENT_LISTENERS_TIMEOUT
 from apify._crypto import decrypt_input_secrets, load_private_key
 from apify._models import ActorRun
--- a/src/apify/_configuration.py
+++ b/src/apify/_configuration.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import traceback
 from datetime import datetime, timedelta
 from decimal import Decimal
 from logging import getLogger
@@ -417,6 +418,7 @@ class Configuration(CrawleeConfiguration):
         """
         if self.is_at_home and not self.disable_browser_sandbox:
             self.disable_browser_sandbox = True
+            logger.info('Stack trace:\n%s', ''.join(traceback.format_stack()))
             logger.warning('Actor is running on the Apify platform, `disable_browser_sandbox` was changed to True.')
         return self
 
--- a/src/apify/_configuration.py
+++ b/src/apify/_configuration.py
@@ -9,6 +9,8 @@ from typing import Annotated, Any
 from pydantic import AliasChoices, BeforeValidator, Field, model_validator
 from typing_extensions import Self, deprecated
 
+import crawlee
+from crawlee._service_locator import ServiceLocator
 from crawlee._utils.models import timedelta_ms
 from crawlee._utils.urls import validate_http_url
 from crawlee.configuration import Configuration as CrawleeConfiguration
@@ -429,8 +431,25 @@ class Configuration(CrawleeConfiguration):
         Mostly for the backwards compatibility. It is recommended to use the `service_locator.get_configuration()`
         instead.
         """
-        return cls()
+        return service_locator.get_configuration()
 
 
-# Monkey-patch the base class so that it works with the extended configuration
-CrawleeConfiguration.get_global_configuration = Configuration.get_global_configuration  # type: ignore[method-assign]
+class ApifyServiceLocator(ServiceLocator):
+    """Same as ServiceLocator from Crawlee, but it always returns Apify Configuration."""
+
+    def get_configuration(self) -> Configuration:
+        # ApifyServiceLocator can store any children of Crawlee Configuration, but in Apify context it is desired to
+        # return Apify Configuration.
+        stored_configuration = super().get_configuration()
+        # Ensure the returned configuration is of type Apify Configuration.
+        model_dump = stored_configuration.model_dump()
+        # The configuration will read env variables first and overridden with stored_configuration
+        _config = Configuration()
+        for key in model_dump:
+            setattr(_config, key, model_dump[key])
+        return _config
+
+
+# Ensure that ApifyServiceLocator is used to make sure Apify Configuration is used.
+service_locator = ApifyServiceLocator()
+crawlee.service_locator = service_locator
--- a/src/apify/_configuration.py
+++ b/src/apify/_configuration.py
@@ -1,21 +1,25 @@
 from __future__ import annotations
 
 import traceback
+import types
 from datetime import datetime, timedelta
 from decimal import Decimal
 from logging import getLogger
-from typing import Annotated, Any
+from typing import TYPE_CHECKING, Annotated, Any, cast
 
 from pydantic import AliasChoices, BeforeValidator, Field, model_validator
 from typing_extensions import Self, deprecated
 
 import crawlee
-from crawlee._service_locator import ServiceLocator
+from crawlee import service_locator
 from crawlee._utils.models import timedelta_ms
 from crawlee._utils.urls import validate_http_url
 from crawlee.configuration import Configuration as CrawleeConfiguration
 
 from apify._utils import docs_group
+
+if TYPE_CHECKING:
+    from crawlee._service_locator import ServiceLocator
 
 logger = getLogger(__name__)
 
@@ -431,20 +435,21 @@ class Configuration(CrawleeConfiguration):
         Mostly for the backwards compatibility. It is recommended to use the `service_locator.get_configuration()`
         instead.
         """
-        return service_locator.get_configuration()
+        return cast('Configuration', service_locator.get_configuration())
 
 
-class ApifyServiceLocator(ServiceLocator):
-    """Same as ServiceLocator from Crawlee, but it always returns Apify Configuration."""
+def _patch_service_locator() -> None:
+    """Patch the Crawlee ServiceLocator to ensure that Apify Configuration is always returned."""
+    old_get_configuration = crawlee.service_locator.get_configuration
 
-    def get_configuration(self) -> Configuration:
+    def get_configuration(self: ServiceLocator) -> Configuration:
         # ApifyServiceLocator can store any children of Crawlee Configuration, but in Apify context it is desired to
         # return Apify Configuration.
         if isinstance(self._configuration, Configuration):
             # If Apify configuration was already stored in service locator, return it.
             return self._configuration
 
-        stored_configuration = super().get_configuration()
+        stored_configuration = old_get_configuration()
         apify_configuration = Configuration()
 
         # Ensure the returned configuration is of type Apify Configuration.
@@ -457,7 +462,8 @@ class ApifyServiceLocator(ServiceLocator):
 
         return apify_configuration
 
+    crawlee.service_locator.get_configuration = types.MethodType(get_configuration, crawlee.service_locator)
+
 
 # Ensure that ApifyServiceLocator is used to make sure Apify Configuration is used.
-service_locator = ApifyServiceLocator()
-crawlee.service_locator = service_locator
+_patch_service_locator()
--- a/src/apify/_configuration.py
+++ b/src/apify/_configuration.py
@@ -445,17 +445,17 @@ class ApifyServiceLocator(ServiceLocator):
             return self._configuration
 
         stored_configuration = super().get_configuration()
+        apify_configuration = Configuration()
 
         # Ensure the returned configuration is of type Apify Configuration.
         # Most likely crawlee configuration was already set. Create Apify configuration from it.
-        # Env vars will have lower priority than the stored configuration.
-        model_dump = stored_configuration.model_dump()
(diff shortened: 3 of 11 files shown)


src/apify/_actor.py
```diff
@@ -595,6 +631,11 @@ async def set_value(
     def get_charging_manager(self) -> ChargingManager:
         """Retrieve the charging manager to access granular pricing information."""
         self._raise_if_not_initialized()
+        return self._get_charging_manager_implementation()
+
+    def _get_charging_manager_implementation(self) -> ChargingManagerImplementation:
```
janbuchar: Why not make this a cached property so that we could get rid of the _charging_manager completely?

tests/unit/test_proxy_configuration.py
```diff
@@ -532,9 +532,9 @@ async def test_initialize_when_status_page_unavailable(
 
     await proxy_configuration.initialize()
 
-    assert len(caplog.records) == 1
+    assert len(caplog.records) == 2
```
janbuchar: Soo what was added here? 😁

tests/unit/events/test_apify_event_manager.py
```diff
@@ -24,9 +24,8 @@
 
 async def test_lifecycle_local(caplog: pytest.LogCaptureFixture) -> None:
     caplog.set_level(logging.DEBUG, logger='apify')
-    config = Configuration.get_global_configuration()
 
-    async with ApifyEventManager(config):
+    async with ApifyEventManager(Configuration()):
```
janbuchar, line 28: Why is this change necessary?

tests/unit/actor/test_actor_lifecycle.py
```diff
@@ -150,6 +150,9 @@ async def handler(websocket: websockets.asyncio.server.ServerConnection) -> None
         port: int = ws_server.sockets[0].getsockname()[1]  # type: ignore[index]
         monkeypatch.setenv(ActorEnvVars.EVENTS_WEBSOCKET_URL, f'ws://localhost:{port}')
 
+        # Make sure there is a charging manager that can be mocked
+        Actor._get_charging_manager_implementation()
```
janbuchar: I'm not particularly happy about using a protected method _and_ relying on its internals. Perhaps there's a better way to mock the charging manager that wouldn't require this?

src/apify/_configuration.py
```diff
@@ -424,11 +425,35 @@ def disable_browser_sandbox_on_platform(self) -> Self:
...
+        """
+        apify_configuration = cls()
+
+        # Ensure the returned configuration is of type Apify Configuration.
+        # Most likely crawlee configuration was already set. Create Apify configuration from it.
+        # Due to known Pydantic issue https://github.com/pydantic/pydantic/issues/9516, creating new instance of
+        # Configuration from existing one in situation where environment can have some fields set by alias is very
+        # unpredictable. Use the stable workaround.
+        for name in configuration.model_fields:
+            setattr(apify_configuration, name, getattr(configuration, name))
```
janbuchar, line 463: This screams "fragile" to me. Wouldn't a combination of `cls.model_validate` and `configuration.model_dump` achieve the same result? Probably after setting up the right parameters...

tests/unit/actor/test_actor_lifecycle.py
Pijukatel: Ok, I can mock public method Actor.new_client instead

tests/unit/events/test_apify_event_manager.py
Pijukatel: Not necessary, but I wanted to avoid Configuration.get_global_configuration() as it warns that it is setting the implicit default configuration as no configuration was set explicitly.
I can also properly set the configuration before the test or just pass the configuration directly and avoid the service_locator related side effects

tests/unit/test_proxy_configuration.py
Pijukatel: It will throw this warning:
"'No configuration set, implicitly creating and using default Configuration.'"

Which is fine. Normally, the proxy configuration would be used within the context of the Actor and it would set the configuration.

It still works even this way, it just warns that no one set the configuration before and thus it is doing it now implicitly.

src/apify/_configuration.py
Pijukatel: I am not happy about that. But that is the current limitation of the Pydantic mentioned in the issue in the comment above the code.

Here is short reproducible snippet to help understand it in isolation:
```
import os
from typing import Annotated

from pydantic import AliasChoices, Field
from pydantic_settings import BaseSettings, SettingsConfigDict


class Settings(BaseSettings):

    model_config = SettingsConfigDict(validate_by_name=True, validate_by_alias=True)

    a: Annotated[
        str,
        Field(
            validation_alias=AliasChoices(
                'a_alias',
            ),
        ),
    ] = 'whatever'


# Aliases in the environment cause very weird behavior
os.environ["a_alias"] = "whatever"
# This will use the alias to create the settings instance
existing_settings = Settings()

# All these fail because of the env variable alias takes precedence in weird way and causes runtime error
new_settings = Settings.model_validate(existing_settings.model_dump())
new_settings = Settings(a=5)
new_settings = Settings.model_validate({'a': 5})
```

src/apify/_actor.py
Pijukatel: Done

janbuchar: Now drop the `get_` prefix please 🙂

src/apify/_actor.py
```diff
@@ -198,32 +193,87 @@ def __call__(
...
+        return self.config
 
     @property
     def config(self) -> Configuration:
         """The Configuration instance the Actor instance uses."""
-        return self._configuration
+        if self._configuration:
+            return self._configuration
+
+        return self._finalize_implicit_configuration()
```
janbuchar: There's `config`, `configuration`, `_configuration` and `_finalize_implicit_configuration`. Can we use `cached_property` to get rid of `_configuration`? Or something in a similar vein, 4 names for essentially the same thing is too much.

src/apify/_actor.py
```diff
@@ -198,32 +193,87 @@ def __call__(
...
+                'Storage client in service locator was set explicitly before Actor.init was called.'
+                'Using the existing storage client as implicit storage client for the Actor.'
+            )
+        self._local_storage_client = service_locator.get_storage_client()
+        return self._local_storage_client
 
     @property
     def event_manager(self) -> EventManager:
         """The EventManager instance the Actor instance uses."""
+        if not self._event_manager:
```
janbuchar: This also kinda looks like a `cached_property`.

tests/unit/storage_clients/test_file_system.py
```diff
@@ -75,7 +75,6 @@ async def test_pre_existing_input_used_by_actor(tmp_path: Path) -> None:
     path_to_input.mkdir(parents=True)
     (path_to_input / f'{configuration.input_key}.json').write_text(json.dumps(pre_existing_input))
 
-    # Remove this line after https://github.com/apify/apify-sdk-python/pull/576
```
janbuchar, line 79: Soo, why has this become permanent?

Pijukatel: Just forgot to delete it, thanks.

src/apify/_actor.py
```diff
@@ -198,32 +194,93 @@ def __call__(
...
     def config(self) -> Configuration:
         """The Configuration instance the Actor instance uses."""
-        return self._configuration
+        if self._configuration:
+            return self._configuration
+
+        return self._finalize_implicit_configuration()
+
+    def _finalize_implicit_configuration(self) -> Configuration:
+        """Set implicit configuration in the actor and return it.
```
vdusek:
```suggestion
        """Set implicit configuration in the Actor and return it.
```

tests/integration/test_actor_migration.py
```diff
@@ -0,0 +1,48 @@
+from __future__ import annotations
```
vdusek: We have some reboot-related tests already, does this cover something the existing tests don't?

Pijukatel: This is the only reboot test with Crawler. The other tests seemed to be testing lower-level components of the crawler.

src/apify/_actor.py
Pijukatel: `config` and `configuration` are I guess, some backwards compatibility, and they are the same. So no change in this PR.

I do not think we can get rid of `_configuration` as it can be set implicitly by `_finalize_implicit_configuration` or explicitly by the user.

src/apify/_actor.py
Pijukatel: uff that is awkward, done

src/apify/_actor.py
Pijukatel: Done

src/apify/_actor.py
Pijukatel: Done

src/apify/storage_clients/_file_system/_storage_client.py
```diff
@@ -32,10 +32,7 @@ async def create_kvs_client(
     ) -> FileSystemKeyValueStoreClient:
         configuration = configuration or Configuration.get_global_configuration()
         client = await ApifyFileSystemKeyValueStoreClient.open(
-            id=id,
-            name=name,
-            alias=alias,
-            configuration=configuration,
+            id=id, name=name, alias=alias, configuration=configuration
```
vdusek: ?

uv.lock
```diff
@@ -112,26 +112,26 @@ dev = [
 
 [[package]]
 name = "apify-client"
-version = "2.0.0"
+version = "2.1.0"
```
vdusek: please revert the changes in the lock file that are not related to this change

tests/unit/actor/test_configuration.py
```diff
@@ -16,6 +27,204 @@ def test_disable_browser_sandbox(
...
+async def test_file_system_storage_client_warning(caplog: pytest.LogCaptureFixture) -> None:
+    service_locator.set_storage_client(FileSystemStorageClient())
+    caplog.set_level('WARNING')
+    async with Actor():
+        ...
+
+    assert (
+        'Using `FileSystemStorageClient` in Actor context is not recommended and can lead to problems with reading '
+        'the input file. Use `ApifyFileSystemStorageClient` instead.'
+    ) in caplog.messages
```
vdusek: Ok, it's great that we have covered this this way, but `ApifyFileSystemStorageClient` is exported as `FileSystemStorageClient`, so please adjust the message accordingly.

tests/unit/actor/test_configuration.py
```diff
@@ -16,6 +27,204 @@ def test_disable_browser_sandbox(
...
+
+
+async def test_storage_retrieved_is_same_with_same_config() -> None:
+    """Test that retrieving storage is same if same configuration is used."""
+    async with Actor():
+        actor_kvs = await Actor.open_key_value_store()
+        crawler = BasicCrawler()
+        crawler_kvs = await crawler.get_key_value_store()
+
+    assert actor_kvs is crawler_kvs
```
vdusek: This is horrible that we have `get_...` and `open_...` variants ... couldn't we unify it? I would use `open_...` in Crawlee as well.

tests/unit/actor/test_configuration.py
```diff
@@ -16,6 +27,204 @@ def test_disable_browser_sandbox(
...
+    assert crawler._service_locator.get_configuration() is config_global
+
+
+async def test_crawler_uses_implicit_apify_config() -> None:
+    """Test that Actor is using implicit ApifyConfiguration in Actor context."""
+    async with Actor:
+        assert isinstance(Actor.config, ApifyConfiguration)
+
+
+async def test_storage_retrieved_is_different_with_different_config(tmp_path: Path) -> None:
```
vdusek: This could parametrized for KVS + Dataset or maybe even RQ as well?

(Same for the tests after this one)

tests/unit/actor/test_configuration.py
```diff
@@ -16,6 +27,204 @@ def test_disable_browser_sandbox(
...
+
+async def test_crawlers_own_configuration(tmp_path: Path) -> None:
+    """Test that crawlers can use own configurations without crashing."""
+    config_actor = ApifyConfiguration()
+    dir_1 = tmp_path / 'dir_1'
+    dir_2 = tmp_path / 'dir_2'
+    config_crawler_1 = ApifyConfiguration()
+    config_actor.storage_dir = str(dir_1)
+    config_crawler_2 = ApifyConfiguration()
+    config_crawler_2.storage_dir = str(dir_2)
```
vdusek, line 126: why you don't provide the storage_dir to the config constructor directly?

tests/unit/actor/test_configuration.py
```diff
@@ -16,6 +27,204 @@ def test_disable_browser_sandbox(
...
+
+
+async def test_storage_retrieved_is_different_with_different_config(tmp_path: Path) -> None:
+    """Test that retrieving storage depends on used configuration."""
+    dir_1 = tmp_path / 'dir_1'
+    dir_2 = tmp_path / 'dir_2'
+    config_actor = ApifyConfiguration()
+    config_actor.storage_dir = str(dir_1)
+    config_crawler = ApifyConfiguration()
+    config_crawler.storage_dir = str(dir_2)
```
vdusek, line 177: why you don't provide the storage_dir to the config constructor directly?

tests/unit/actor/test_configuration.py
```diff
@@ -1,6 +1,17 @@
...
+from crawlee import Request, service_locator
+from crawlee._types import BasicCrawlingContext
+from crawlee.configuration import Configuration as CrawleeConfiguration
+from crawlee.crawlers import BasicCrawler
+from crawlee.errors import ServiceConflictError
+from crawlee.storage_clients import FileSystemStorageClient
+
+from apify import Actor
+from apify import Configuration as ApifyConfiguration
+from apify.storage_clients._file_system import ApifyFileSystemStorageClient
```
vdusek: The more we touch only the public interface, the less private the better.

```suggestion
from apify.storage_clients import FileSystemStorageClient as ApifyFileSystemStorageClient
```

tests/unit/actor/test_configuration.py
```diff
@@ -1,6 +1,17 @@
+from pathlib import Path
```
vdusek, line 1: it's great that this behaviour and all the possible edge cases will be declared somewhere and tested

tests/unit/actor/test_apify_storage.py
```diff
@@ -0,0 +1,62 @@
+from datetime import datetime, timezone
```
vdusek: rename it to `test_apify_storages` please

tests/unit/actor/test_apify_storage.py
```diff
@@ -0,0 +1,62 @@
...
+from unittest import mock
+from unittest.mock import AsyncMock
+
+import pytest
+
+from crawlee.storage_clients.models import StorageMetadata
+from crawlee.storages import Dataset, KeyValueStore, RequestQueue
```
vdusek: import from `apify.storages` please

tests/unit/actor/test_actor_lifecycle.py
```diff
@@ -6,6 +6,7 @@
 import sys
```
vdusek, line 6: could you explain changes in this file?

tests/integration/test_actor_migration.py
vdusek: well, so the point of this test is that it tests the reboot with the crawler, so I would say it should be better named

src/apify/storage_clients/_apify/_utils.py
```diff
@@ -1,117 +1,138 @@
 from __future__ import annotations
 
+import base64
```
vdusek: 1) Encryption is not necessary in this case, we should use just hashing.
2) Use the similar mechanisms (separator, hashing, ...) as for the [computing of unique_key](https://github.com/apify/crawlee-python/blob/master/src/crawlee/_utils/requests.py).

src/apify/storage_clients/_apify/_storage_client.py
```diff
@@ -1,44 +1,68 @@
 from __future__ import annotations
```
vdusek: Will this work for the default unnamed storage + `force_cloud`? Or what will it do?

src/apify/_actor.py
```diff
@@ -5,6 +5,7 @@
 import sys
```
vdusek, line 5: Remove long-time deprecated `config` property (stay only with `configuration` - we refer to it everywhere with the whole name).

src/apify/_actor.py
```diff
@@ -465,13 +527,12 @@ async def open_key_value_store(
...
         self._raise_if_not_initialized()
         self._raise_if_cloud_requested_but_not_configured(force_cloud=force_cloud)
-        storage_client = self._cloud_storage_client if force_cloud else self._local_storage_client
+        storage_client = self._cloud_storage_client if force_cloud else self._get_local_storage_client()
 
         return await KeyValueStore.open(
             id=id,
             alias=alias,
             name=name,
-            configuration=self._configuration,
```
vdusek, line 474: ?

src/apify/_actor.py
```diff
@@ -506,13 +567,12 @@ async def open_request_queue(
...
         self._raise_if_cloud_requested_but_not_configured(force_cloud=force_cloud)
 
-        storage_client = self._cloud_storage_client if force_cloud else self._local_storage_client
+        storage_client = self._cloud_storage_client if force_cloud else self._get_local_storage_client()
 
         return await RequestQueue.open(
             id=id,
             alias=alias,
             name=name,
-            configuration=self._configuration,
```
vdusek, line 515: ?

src/apify/_actor.py
vdusek, line 1: This is not necessary related to this PR, but it is to the upcoming major release.

The usage of an Actor's context manager versus explicitly calling init/exit is inconsistent. Specifically, exit accepts arguments, while aexit cannot take explicit arguments. This should be unified.

Moreover, the direction of delegation should likely be reversed: exit should call aexit, not the other way around, to ensure argument parity. Same for the init and aenter.

vdusek (changes requested):
Another batch (from our discussion), I'll wait for your updates now.

tests/unit/actor/test_configuration.py
Pijukatel: Agreed, but not related to changes in this PR. I will do this in a separate PR. Pinky promise!

tests/unit/actor/test_configuration.py
Pijukatel: This is related to the Pydantic issue. Explained in this comment
https://github.com/apify/apify-sdk-python/pull/576#discussion_r2341428747

The alternative is type ignore. Pydantic does not handle this well.
`config_crawler_1 = ApifyConfiguration(crawlee_storage_dir=str(dir_1),  # type: ignore[call-arg])`

src/apify/storage_clients/_apify/_storage_client.py
```diff
@@ -9,36 +9,47 @@
...
+
     @override
     async def create_dataset_client(
         self,
         *,
         id: str | None = None,
         name: str | None = None,
         alias: str | None = None,
-        configuration: Configuration | None = None,
+        configuration: CrawleeConfiguration | None = None,
```
janbuchar, line 48: I guess I might just be out of the loop, but wasn't this supposed to work with some "stored" configuration so that we could just remove this parameter? Is there some reason why we need it?

src/apify/storage_clients/_apify/_utils.py
```diff
@@ -1,117 +1,144 @@
...
-
-        # get_record can return {key: ..., value: ..., content_type: ...}
-        if isinstance(record, dict) and 'value' in record:
-            record = record['value']
 
-        # Extract the actual data from the KVS record
-        if isinstance(record, dict) and alias_key in record:
-            storage_id = record[alias_key]
-            return str(storage_id)
+_StorageT = type[Dataset | KeyValueStore | RequestQueue]
```
janbuchar: This seems to be a brand new naming convention for type aliases. Perhaps _TStorage would be more in line with what we already have?

src/apify/_configuration.py
```diff
@@ -424,11 +425,35 @@ def disable_browser_sandbox_on_platform(self) -> Self:
...
+        if isinstance(global_configuration, Configuration):
+            # If Apify configuration was already stored in service locator, return it.
+            return global_configuration
 
-# Monkey-patch the base class so that it works with the extended configuration
-CrawleeConfiguration.get_global_configuration = Configuration.get_global_configuration  # type: ignore[method-assign]
+        return cls.from_configuration(global_configuration)
+
+    @classmethod
+    def from_configuration(cls, configuration: CrawleeConfiguration) -> Configuration:
```
janbuchar, line 446: The result should be cached so that two calls to get_global_configuration always return the exact same object.

tests/unit/actor/test_apify_storage.py
janbuchar: Also I don't think it should be in the tests/unit/actor directory, perhaps a level higher?

src/apify/_actor.py
janbuchar: I think that one private attribute for explicitly set configuration and a cached property (configuration) that performs the logic of _finalize_implicit_configuration if necessary would be much simpler. Plus, as @vdusek pointed out, the config property can be removed in favor of configuration.

src/apify/_actor.py
```diff
@@ -119,28 +123,17 @@ def __init__(
...
-                persist_state_interval=self._configuration.persist_state_interval,
-            )
-        )
+        self._configuration = configuration
+        self._configure_logging = configure_logging
+        self._apify_client: ApifyClientAsync | None = None
+        self._local_storage_client: StorageClient | None = None
 
-        self._charging_manager = ChargingManagerImplementation(self._configuration, self._apify_client)
+        self._charging_manager: ChargingManagerImplementation | None = None
```
janbuchar: Is this still used? I'm led to believe it's not

src/apify/_actor.py
```diff
@@ -607,7 +670,11 @@ async def set_value(
     def get_charging_manager(self) -> ChargingManager:
         """Retrieve the charging manager to access granular pricing information."""
         self._raise_if_not_initialized()
-        return self._charging_manager
+        return self.charging_manager_implementation
+
+    @cached_property
+    def charging_manager_implementation(self) -> ChargingManagerImplementation:
+        return ChargingManagerImplementation(self.config, self.apify_client)
```
janbuchar: I believe this should call _raise_if_not_initialized too

src/apify/_actor.py
```diff
@@ -607,7 +670,11 @@ async def set_value(
     def get_charging_manager(self) -> ChargingManager:
         """Retrieve the charging manager to access granular pricing information."""
         self._raise_if_not_initialized()
-        return self._charging_manager
+        return self.charging_manager_implementation
```
janbuchar: Do we really need this method now?

src/apify/_actor.py
Pijukatel: We keep inflating this PR. Deleting the deprecated `config` property is unrelated and could have been done before or after this PR. I added it, but please lets focus on changes tightly related to the PR.

As for `_finalize_implicit_configuration` I have to keep it due to the Actor instance/class duality. It has two uses:
1)
Someone wants to use `Actor.configuration` before the `init` was called. It is allowed, but now the configuration is finalized both in `Actor` and `service_locator` and it throws a warning that implicit configuration was used.
2)
Someone explicitly sets config in Actor(configuration=...) and when `init` is called, this configuration gets finalized - set in global `service_locator`

tests/integration/test_actor_migration.py
Pijukatel: renamed and moved

src/apify/storage_clients/_file_system/_storage_client.py
Pijukatel: `make format`

tests/unit/actor/test_actor_lifecycle.py
Pijukatel: Previously, the Actor._charging_manager was already set, and it could have been mocked. Now it is finalized during init, so to mock the internal client, I have to mock the `Actor.new_client` method that is used to create the client for the charging manager

tests/unit/actor/test_apify_storage.py
Pijukatel: Rename, moved

tests/unit/actor/test_apify_storage.py
Pijukatel: Done

src/apify/storage_clients/_file_system/_storage_client.py
vdusek: That is not true 🙂, revert it please

tests/unit/actor/test_configuration.py
vdusek: Sure, and we will probably want to do the changes in the Crawlee (rather open than get).

tests/unit/actor/test_configuration.py
Pijukatel: Updated

tests/unit/actor/test_configuration.py
Pijukatel: Updated

tests/unit/actor/test_configuration.py
vdusek: Ah, okay, I mean we've been using the approach type ignores in all other places, so I would stay with that (with appropriate comment). And once the Pydantic issue is resolved, we can fix it in all the places.

src/apify/storage_clients/_apify/_storage_client.py
Pijukatel: I tried that approach initially, but that led to more problems related to the order in which services can be set in the service locator. Since one service is dependent on another service, but they can be set independently... 
So imagine you do:
service_locator.set_configuration(Configuration1)
service_locator.set_storage_client(SomeStorageClient(Configuration2))
What now? Better to prevent this from even happening

src/apify/storage_clients/_apify/_utils.py
Pijukatel: Updated

src/apify/_configuration.py
Pijukatel: I added a warning to the only potential path where this could cause a problem. If someone gets there, it is not by intention, but for the sake of backward compatibility, it will still work in most cases as expected.

src/apify/storage_clients/_apify/_utils.py
Pijukatel: Done

tests/unit/actor/test_configuration.py
Pijukatel: The other approach requires ignore comment, and is potentially more fragile as it relies on somewhat internal Pydantic behavior, that can change. I will stay with this approach in this change.

tests/unit/actor/test_configuration.py
Pijukatel: https://github.com/apify/apify-sdk-python/pull/576/files#r2356047155

docs/04_upgrading/upgrading_to_v3.md
```diff
@@ -9,6 +9,42 @@ This page summarizes the breaking changes between Apify Python SDK v2.x and v3.0
 
 Support for Python 3.9 has been dropped. The Apify Python SDK v3.x now requires Python 3.10 or later. Make sure your environment is running a compatible version before upgrading.
 
+## Actor initialization and ServiceLocator changes
+
+`Actor` initialization and global `service_locator` services setup is more strict and predictable.
+- Services in `Actor` can't be changed after calling `Actor.init`, entering the `async with Actor` context manager or after requesting them from the `Actor`
+- Services in `Actor` can be different from services in Crawler
```
vdusek:
```suggestion
- Services in `Actor` can't be changed after calling `Actor.init`, entering the `async with Actor` context manager or after requesting them from the `Actor`.
- Services in `Actor` can be different from services in Crawler.
```

src/apify/_actor.py
Pijukatel: Preferably, all uses should go through the public property, and only the internall methods that have something to do with setting up the configuration for the actor should access private _configuration

src/apify/_actor.py
Pijukatel: Preferably, all uses should go through the public property, and only the internall methods that have something to do with setting up the configuration for the actor should access private _configuration

pyproject.toml
```diff
@@ -36,7 +36,7 @@ keywords = [
 dependencies = [
     "apify-client>=2.0.0,<3.0.0",
     "apify-shared>=2.0.0,<3.0.0",
-    "crawlee==0.6.13b37",
+    "crawlee @ git+https://github.com/apify/crawlee-python.git@master",
```
vdusek: I would rather use the newest beta: `crawlee==0.6.13b42`, as in the lockfile, there is stored a specific commit anyway, so this could be confusing.

src/apify/_actor.py
vdusek: Issue: https://github.com/apify/apify-sdk-python/issues/598 and resolving

src/apify/_actor.py
Pijukatel: Public usage of charging_manager should be allowed only if Actor is initialized and Actor is not initialized before `await self._charging_manager_implementation.__aenter__()`

uv.lock
Pijukatel: Reverted

src/apify/_actor.py
Pijukatel: Removed

src/apify/_actor.py
Pijukatel: https://github.com/apify/apify-sdk-python/pull/576/files#r2358849136

src/apify/storage_clients/_file_system/_storage_client.py
Pijukatel: Done

src/apify/_actor.py
vdusek: I would also be in favour of removing `_finalize_implicit_configuration` and inlining its functionality into the cached property `configuration`. We can call the cached property from the init, if I am not mistaken. So it's not really necessary.

src/apify/storage_clients/_apify/_storage_client.py
Pijukatel: This behavior is unchanged. Default unnamed storage with force cloud relies on configuration to give it default storage id.

src/apify/_actor.py
vdusek, line 1: Please look for occurrences of "actor" (lowercase) in docstrings/comments and replace them with "Actor".

src/apify/_actor.py
```diff
@@ -265,9 +323,10 @@ async def init(self) -> None:
         # If the Actor is running on the Apify platform, we set the cloud storage client.
         if self.is_at_home():
             service_locator.set_storage_client(self._cloud_storage_client)
-
+            self._local_storage_client = self._cloud_storage_client
+        else:
+            self._finalize_implicit_local_storage_client()
```
vdusek: Why are the configurations of the local and cloud storage clients done in different ways? Couldn't it be unified?

src/apify/storage_clients/_apify/_storage_client.py
Pijukatel: https://github.com/apify/apify-sdk-python/issues/599

src/apify/_actor.py
Pijukatel: Good catch

src/apify/storage_clients/_apify/_dataset_client.py
```diff
@@ -126,19 +127,19 @@ async def open(
...
-                id = resolved_id
-            else:
-                # Create a new storage and store the alias mapping
-                new_storage_metadata = DatasetMetadata.model_validate(
-                    await apify_datasets_client.get_or_create(),
-                )
-                id = new_storage_metadata.id
-                await store_alias_mapping(alias, 'dataset', id, configuration)
+            # Check if there is pre-existing alias mapping in the default KVS.
+            async with Alias(storage_type=Dataset, alias=alias, configuration=configuration) as _alias:
```
vdusek: The name of the class `Alias` is not good; here we can see why, AliasResolver or something like that would be better.

src/apify/storage_clients/_apify/_utils.py
vdusek, line 1: Why does the Alias class now have so many public methods? I can see only 2 valid use cases: resolve alias to ID, and store alias: ID mapping.

src/apify/storage_clients/_apify/_utils.py
```diff
@@ -1,117 +1,170 @@
...
-
-
-async def _get_default_kvs_client(configuration: Configuration) -> KeyValueStoreClientAsync:
+    _alias_map: ClassVar[dict[str, str]] = {}
+    """Map containing pre-existing alias storages and their ids. Global for all instances."""
+    _alias_init_lock: Lock | None = None
+    """Lock for creating alias storages. Only one alias storage can be created at the time. Global for all instances."""
+
+    ALIAS_STORAGE_KEY_SEPARATOR = ','
+    ALIAS_MAPPING_KEY = '__STORAGE_ALIASES_MAPPING'
```
vdusek: These could be private class attrs.

```suggestion
    _ALIAS_STORAGE_KEY_SEPARATOR = ','
    _ALIAS_MAPPING_KEY = '__STORAGE_ALIASES_MAPPING'
```

src/apify/storage_clients/_apify/_utils.py
vdusek, line 1: I would name the entire module `alias_resolver.py` (or a similar name) and the class accordingly.

src/apify/storage_clients/_apify/_utils.py
```diff
@@ -1,117 +1,170 @@
...
+            else:
+                record = {self.storage_key: storage_id}
+
+            # Store the mapping back in the KVS.
+            await default_kvs_client.set_record(self.ALIAS_MAPPING_KEY, record)
+        except Exception as exc:
+            logger.warning(f'Error storing alias mapping for {self.alias}: {exc}')
+
+
+async def get_default_kvs_client() -> KeyValueStoreClientAsync:
```
vdusek: private
```suggestion
async def _get_default_kvs_client() -> KeyValueStoreClientAsync:
```

src/apify/storage_clients/_apify/_utils.py
```diff
@@ -1,117 +1,170 @@
...
+        """Get additional cache key based on configuration.
+
+        Use only api_public_base_url and token as the relevant for differentiating storages.
+        """
+        if configuration.api_public_base_url is None or configuration.token is None:
+            raise ValueError("'Configuration.api_public_base_url' and 'Configuration.token' must be set.")
+        return compute_short_hash(f'{configuration.api_public_base_url}{configuration.token}'.encode())
+
+    @property
+    def storage_key(self) -> str:
```
vdusek: could be private

src/apify/storage_clients/_apify/_utils.py
```diff
@@ -1,117 +1,170 @@
...
+
+    ALIAS_STORAGE_KEY_SEPARATOR = ','
+    ALIAS_MAPPING_KEY = '__STORAGE_ALIASES_MAPPING'
+
+    def __init__(
+        self, storage_type: type[Dataset | KeyValueStore | RequestQueue], alias: str, configuration: Configuration
+    ) -> None:
+        self.storage_type = storage_type
+        self.alias = alias
+        self.additional_cache_key = self.get_additional_cache_key(configuration)
```
vdusek: private attrs

```suggestion
        self._storage_type = storage_type
        self._alias = alias
        self._additional_cache_key = self.get_additional_cache_key(configuration)
```

src/apify/storage_clients/_apify/_utils.py
```diff
@@ -1,117 +1,170 @@
...
+
+        The lock is shared for all instances of the Alias class.
+        It is created in async method to ensure that some event loop is already running.
+        """
+        if cls._alias_init_lock is None:
+            cls._alias_init_lock = Lock()
+        return cls._alias_init_lock
+
+    @classmethod
+    async def get_alias_map(cls) -> dict[str, str]:
```
vdusek: why public?

src/apify/storage_clients/_apify/_utils.py
```diff
@@ -1,117 +1,170 @@
...
+                    cls._alias_map = record['value']
+                else:
+                    cls._alias_map = record
+            else:
+                cls._alias_map = dict[str, str]()
+
+        return cls._alias_map
+
+    @classmethod
+    def get_additional_cache_key(cls, configuration: Configuration) -> str:
```
vdusek: I don't think this belongs under this class

tests/unit/test_apify_storages.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+@pytest.mark.parametrize(
+    ('storage', '_storage_client'),
+    [
+        (Dataset, ApifyDatasetClient),
+        (KeyValueStore, ApifyKeyValueStoreClient),
+        (RequestQueue, ApifyRequestQueueClient),
+    ],
+)
```
vdusek, line 23: pytest param please


e.g.

```python
@pytest.mark.parametrize(
    argnames='true_positive',
    argvalues=[
        pytest.param('http://www.something.com', id='standard_http_with_www'),
        pytest.param('https://www.something.net', id='standard_https_with_www'),
        pytest.param('http://nowww.cz', id='http_no_www'),
        pytest.param('https://with-hypen.com', id='https_with_hyphen'),
        pytest.param('http://number1.com', id='http_with_number_in_domain'),
        pytest.param('http://www.number.123.abc', id='http_with_subdomains_and_numbers'),
        pytest.param('http://many.dots.com', id='http_with_multiple_subdomains'),
        pytest.param('http://a.com', id='http_short_domain'),
        pytest.param('http://www.something.com/somethignelse', id='http_with_path_no_extension'),
        pytest.param('http://www.something.com/somethignelse.txt', id='http_with_path_and_extension'),
        pytest.param('http://non-english-chars-áíéåü.com', id='http_with_non_english_chars'),
        pytest.param('http://www.port.com:1234', id='http_with_port'),
        pytest.param('http://username:password@something.else.com', id='http_with_authentication'),
    ],
)
```

pyproject.toml
Pijukatel: Done

src/apify/storage_clients/_apify/_dataset_client.py
Pijukatel: Updated

src/apify/storage_clients/_apify/_utils.py
Pijukatel: Updated

src/apify/storage_clients/_apify/_utils.py
Pijukatel: private

src/apify/storage_clients/_apify/_utils.py
Pijukatel: private

src/apify/storage_clients/_apify/_utils.py
Pijukatel: Ok, util function now

tests/unit/test_apify_storages.py
Pijukatel: There is no benefit for these straightforward params with just two arguments. It is just extra boilerplate.

src/apify/_actor.py
Pijukatel: Because on Apify platform, they are the same and locally they do not have to be.

src/apify/_actor.py
Pijukatel: done

tests/unit/test_apify_storages.py
vdusek: There is a benefit - using `pytest.param` gives you explicit IDs for each case. That means when a test fails, you can see which input caused the failure just from the test output - without needing to inspect the data further/log.

tests/unit/actor/test_configuration.py
vdusek: https://github.com/apify/crawlee-python/issues/1415

src/apify/_actor.py
vdusek: That's true, of course - I was commenting on the deletion.

src/apify/_actor.py
vdusek: I mean this:

```python
if self.is_at_home():
    self._local_storage_client = self._cloud_storage_client
```

is really wrong.

src/apify/storage_clients/_apify/_utils.py
vdusek, line 1: Please follow the general method ordering: dunder methods -> classmethods -> staticmethods -> public instance methods -> private methods. It is not implemented in the ruff yet, but probably will be (e.g. https://github.com/astral-sh/ruff/issues/2425).

src/apify/_actor.py
vdusek: Done

tests/unit/test_apify_storages.py
Pijukatel: That is useful for cryptic and complex parameters. Here it is out of the box done implicitly. See the example Pytest log. There is no doubt about which parameters were used
```
FAILED tests/unit/test_apify_storages.py::test_get_additional_cache_key[Dataset-ApifyDatasetClient] - assert False
```

src/apify/storage_clients/_apify/_utils.py
Pijukatel: Ordered

tests/unit/test_apify_storages.py
vdusek: Interesting... why did I believe there was just `test_get_additional_cache_key1`, `test_get_additional_cache_key2`, ...?

Issue #460, opened 2025-04-22
Ensure compatibility with `lazy-object-proxy` v1.11.0


vdusek:
Example of broken job https://github.com/apify/apify-sdk-python/actions/runs/14596645969/job/40944164914?pr=459 :

```
ERROR tests/unit/actor/test_request_list.py::test_url_no_commas_regex_true_positives[https_with_hyphen] - KeyError: '__wrapped__'
ERROR tests/unit/actor/test_request_list.py::test_url_no_commas_regex_true_positives[http_with_number_in_domain] - KeyError: '__wrapped__'
ERROR tests/unit/actor/test_request_list.py::test_url_no_commas_regex_true_positives[http_with_subdomains_and_numbers] - KeyError: '__wrapped__'
ERROR tests/unit/actor/test_request_list.py::test_url_no_commas_regex_true_positives[http_with_multiple_subdomains] - KeyError: '__wrapped__'
ERROR tests/unit/actor/test_request_list.py::test_url_no_commas_regex_true_positives[http_short_domain] - KeyError: '__wrapped__'
ERROR tests/unit/actor/test_request_list.py::test_url_no_commas_regex_true_positives[http_with_path_no_extension] - KeyError: '__wrapped__'
ERROR tests/unit/actor/test_request_list.py::test_url_no_commas_regex_true_positives[http_with_path_and_extension] - KeyError: '__wrapped__'
ERROR tests/unit/actor/test_request_list.py::test_url_no_commas_regex_true_positives[http_with_non_english_chars] - KeyError: '__wrapped__'
ERROR tests/unit/actor/test_request_list.py::test_url_no_commas_regex_true_positives[http_with_port] - KeyError: '__wrapped__'
ERROR tests/unit/actor/test_request_list.py::test_url_no_commas_regex_true_positives[http_with_authentication] - KeyError: '__wrapped__'
ERROR tests/unit/actor/test_configuration.py::test_disable_browser_sandbox[False-True-True] - KeyError: '__wrapped__'
ERROR tests/unit/actor/test_request_list.py::test_url_no_commas_regex_false_positives[invalid_domain_single_level] - KeyError: '__wrapped__'
ERROR tests/unit/actor/test_configuration.py::test_disable_browser_sandbox[True-False-True] - KeyError: '__wrapped__'
ERROR tests/unit/actor/test_request_list.py::test_url_no_commas_regex_false_positives[invalid_domain_no_tld] - KeyError: '__wrapped__'
[...]

PR #583, opened 2025-09-04
chore: relax `lazy-object-proxy` version upper bound

Mantisus:
### Description

- After upgrading to Python 3.10, there are no compatibility issues.

### Issues

- Closes: #460


vdusek (approved):
Oh, nice

Issue #598, opened 2025-09-18
Unify Actor's context manager with manual init/exit methods


vdusek:
The usage of an Actor's context manager versus explicitly calling init/exit is inconsistent. Specifically, exit accepts arguments, while aexit cannot take explicit arguments. This should be unified.

Moreover, the direction of delegation should likely be reversed: exit should call aexit, not the other way around, to ensure argument parity. Same for the init and aenter.

PR #600, opened 2025-09-19
fix: Unify Actor context manager with init & exit methods

vdusek:
### Description

To align the behavior between the two possible approaches of using an Actor - as a context manager or by explicitly calling `init`/`exit`/`fail` - the following changes were implemented:

- The Actor's arguments (`exit_code`, `status_message`, `event_listeners_timeout`, `cleanup_timeout`) were added to the `Actor.__init__` as well.
- `exit_code` and `status_message` are now public properties with setters, allowing them to be updated within a context manager block in response to error conditions.
- The previous approach of calling `__aenter__` from `init` and `__aexit__` from `exit` was reversed, ensuring fewer inconsistencies for future changes (since `__aenter__` and `__aexit__` have a fixed number of arguments).

Other changes:
- Related docstrings were updated.
- The ordering of methods in the Actor was adjusted (see [ruff issue #2425](https://github.com/astral-sh/ruff/issues/2425)).
- The interface of init, exit, and fail remains unchanged.
- Many more tests have been added.
- The documentation page about the Actor lifecycle has been completely rewritten.

### Issues

- Closes: #598

### Testing

- The current test set cover the changes.

### DIFF
--- a/src/apify/_actor.py
+++ b/src/apify/_actor.py
@@ -158,40 +158,47 @@ class _ActorType:
         This method must be called exactly once per Actor instance. Re-initializing an Actor or having multiple
         active Actor instances is not standard usage and may lead to warnings or unexpected behavior.
         """
-        self.log.info('Initializing Actor...')
-        self.log.info('System info', extra=get_system_info())
-
         if self._is_initialized:
             raise RuntimeError('The Actor was already initialized!')
 
+        # Initialize configuration first - it's required for the next steps.
+        if self._configuration:
+            # User provided explicit configuration - register it in the service locator.
+            service_locator.set_configuration(self.configuration)
+        else:
+            # No explicit configuration provided - trigger creation of default configuration.
+            _ = self.configuration
+
+        # Configure logging based on the configuration, any logs before this point are lost.
+        if self._configure_logging:
+            _configure_logging()
+        self.log.debug('Logging configured')
+
+        self.log.info('Initializing Actor', extra=get_system_info())
+        self.log.debug('Configuration initialized')
+
+        # Warn about non-standard usage patterns.
         if _ActorType._is_any_instance_initialized:
             self.log.warning('Repeated Actor initialization detected - this is non-standard usage, proceed with care.')
 
-        if self._configuration:
-            # Set explicitly the configuration in the service locator.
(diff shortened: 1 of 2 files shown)


vdusek:
IMO that scenario is a misuse. You've got 2 approaches and you should choose one of them. Not combining them.

tests/integration/test_actor_lifecycle.py
```diff
@@ -115,7 +117,7 @@ async def main() -> None:
     actor = await make_actor(label='with-actor-fail', main_func=main)
     run_result = await run_actor(actor)
 
-    assert run_result.exit_code == 91
+    assert run_result.exit_code == ActorExitCodes.ERROR_USER_FUNCTION_THREW.value
```
janbuchar: I usually prefer "inlining" the constants to avoid tautological tests. Not a hill to die on in this case, just saying

docs/02_concepts/code/01_class_manual.py
```diff
@@ -0,0 +1,24 @@
+import asyncio
+
+from apify import Actor
+
+
+async def main() -> None:
```
janbuchar, line 6: Shouldn't the part before `exit` be in a `try` block and the `exit` in a `finally` clause? For educational purposes?

docs/02_concepts/code/01_instance_context.py
```diff
@@ -0,0 +1,27 @@
...
+        # Get input
+        actor_input = await actor.get_input()
+        actor.log.info('Actor input: %s', actor_input)
+
+        # Your Actor logic here
+        data = {'message': 'Hello from Actor instance!', 'input': actor_input}
+        await actor.push_data(data)
+
+        # Set status message
+        await Actor.set_status_message('Actor completed successfully')
```
janbuchar: is this using `Actor` instead of `actor` on purpose? If yes, please add a comment that explains it

vdusek: it's not on purpose, good catch, thank you

docs/02_concepts/code/01_class_manual.py
vdusek: yes, we can do it that way

Issue #599, opened 2025-09-18
Convenient handling of default unnamed storages when running locally


Pijukatel:
Currently, when outside of Apify platform when using default unnamed storages the system relies on preoprely set configuration to get the default storage id.

This can be changed after unnamed default storages are enabled.

New behavior should in this situation take the id from configuration, but if it is default value, then ignore it and treat it as normal alias storage.

PR #606, opened 2025-09-22
refactor!: Make default Apify storages use alias mechanism

Pijukatel:
### Description

- Using unnamed Apify-based storage locally outside of the Apify platform will use the alias mechanism.

### Issues
Closes: #599 

### Testing

Added unit tests

### DIFF
--- a/docs/04_upgrading/upgrading_to_v3.md
+++ b/docs/04_upgrading/upgrading_to_v3.md
@@ -50,7 +50,7 @@ async def main():
 - `Configuration.default_dataset_id` changed from `'default'` to `None`.
 - `Configuration.default_request_queue_id` changed from `'default'` to `None`.
 
-As a consequence of this change, using default storage without specifying its `id` in `Configuration` will use unnamed storage.
+Previously using the default storage without specifying its `id` in `Configuration` would lead to using specific storage with id `'default'`. Now it will use newly created unnamed storage with `'id'` assigned by the Apify platform, consecutive calls to get the default storage will return the same storage.
 
 ## Storages
 
--- a/docs/04_upgrading/upgrading_to_v3.md
+++ b/docs/04_upgrading/upgrading_to_v3.md
@@ -45,6 +45,13 @@ async def main():
 ## Removed Actor.config property
 - `Actor.config` property has been removed. Use `Actor.configuration` instead.
 
+## Default storage ids in configuration changed to None
+- `Configuration.default_key_value_store_id` changed from `'default'` to `None`.
+- `Configuration.default_dataset_id` changed from `'default'` to `None`.
+- `Configuration.default_request_queue_id` changed from `'default'` to `None`.
+
+As a consequence of this change, using default storage without specifying its `id` in `Configuration` will use unnamed storage.
+
 ## Storages
 
 <!-- TODO -->
--- a/src/apify/storage_clients/_apify/_dataset_client.py
+++ b/src/apify/storage_clients/_apify/_dataset_client.py
@@ -124,8 +124,10 @@ class ApifyDatasetClient(DatasetClient):
         )
         apify_datasets_client = apify_client_async.datasets()
 
-        # Normalize 'default' alias to None
-        alias = None if alias == 'default' else alias
+        # Normalize unnamed default storage in cases where not defined in `configuration.default_dataset_id` to unnamed
+        # storage aliased as `__default__`
+        if not any([alias, name, id, configuration.default_dataset_id]):
+            alias = '__default__'
 
         if alias:
             # Check if there is pre-existing alias mapping in the default KVS.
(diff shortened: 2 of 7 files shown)


docs/04_upgrading/upgrading_to_v3.md
```diff
@@ -45,6 +45,13 @@ async def main():
 ## Removed Actor.config property
 - `Actor.config` property has been removed. Use `Actor.configuration` instead.
 
+## Default storage ids in configuration changed to None
+- `Configuration.default_key_value_store_id` changed from `'default'` to `None`.
+- `Configuration.default_dataset_id` changed from `'default'` to `None`.
+- `Configuration.default_request_queue_id` changed from `'default'` to `None`.
+
+As a consequence of this change, using default storage without specifying its `id` in `Configuration` will use unnamed storage.
```
vdusek: TBH, this doesn't explain much - even I'm not sure what it would mean for me or what the consequences are.

Pijukatel: I added a more detailed explanation.

docs/04_upgrading/upgrading_to_v3.md
```diff
@@ -101,3 +101,21 @@ async def main():
...
+
+## Removed Actor.config property
+- `Actor.config` property has been removed. Use `Actor.configuration` instead.
+
+## Default storage ids in configuration changed to None
+- `Configuration.default_key_value_store_id` changed from `'default'` to `None`.
+- `Configuration.default_dataset_id` changed from `'default'` to `None`.
+- `Configuration.default_request_queue_id` changed from `'default'` to `None`.
+
+Previously using the default storage without specifying its `id` in `Configuration` would lead to using specific storage with id `'default'`. Now it will use newly created unnamed storage with `'id'` assigned by the Apify platform, consecutive calls to get the default storage will return the same storage.
```
janbuchar, line 113: "specific storage with id `'default'`" id or name? I'm pretty sure it would be a named storage.

src/apify/storage_clients/_apify/_dataset_client.py
```diff
@@ -150,6 +152,11 @@ async def open(
         # If none are provided, try to get the default storage ID from environment variables.
         elif id is None:
             id = configuration.default_dataset_id
+            if not id:
```
janbuchar, line 155: But... this shouldn't happen thanks to the normalization step that you added, right?

src/apify/storage_clients/_apify/_utils.py
```diff
@@ -76,7 +76,7 @@ async def _get_alias_map(cls) -> dict[str, str]:
         Returns:
             Map of aliases and storage ids.
         """
-        if not cls._alias_map:
+        if not cls._alias_map and Configuration.get_global_configuration().is_at_home:
```
janbuchar, line 79: So this change guarantees that if you open the default storage locally with force_cloud, the mapping won't be stored to the platform? Can you add a test?

docs/04_upgrading/upgrading_to_v3.md
Pijukatel: On latest stable release v2.7.3 it was `id="default"`. For example following test would most likely throw
`RuntimeError: Dataset with id "default" does not exist!`

```
async def test_actor_default() -> None:
    async with Actor:
        await Actor.open_dataset(force_cloud=True)
```

(I am not saying it was correct, but that was the state.)

src/apify/storage_clients/_apify/_dataset_client.py
Pijukatel: It will not happen, but mypy is not able to figure it out. Probably the code could be refactored to make it clearer to mypy, but I would leave it after the release, as it will be non-breaking pure refactoring, similar to https://github.com/apify/apify-sdk-python/pull/595#issuecomment-3292247854

src/apify/storage_clients/_apify/_utils.py
Pijukatel: Added

docs/04_upgrading/upgrading_to_v3.md
janbuchar: Huh, weird.

src/apify/storage_clients/_apify/_dataset_client.py
janbuchar: Can't we just use https://docs.python.org/3/library/typing.html#typing.assert_never? I don't believe that we'll ever come back to change this.

Pijukatel: There is too much runtime checking. The mypy has no clue about max 1 kwargs of three runtime checks and so on. Let's not bend the code too much for the mypy's sake.

Anyway any such refactoring is non-breaking, so no need to rush it in before the release.

PR #617, opened 2025-10-06
fix: Also load input from a file with a .json extension in file system storage

janbuchar:
- fixes bug reported in https://apify.slack.com/archives/C0L33UM7Z/p1759754068120539


vdusek (approved):
LGTM, just wondering whether not to use the default value `INPUT.json` (with extension), instead of `INPUT`. https://github.com/apify/apify-sdk-python/blob/master/src/apify/_configuration.py#L198

vdusek (approved):
LGTM

vdusek (approved):
LGTM

PR #623, opened 2025-10-09
fix: Handle None result in single consumer request queue client

janbuchar:
- closes https://github.com/apify/crawlee-python/issues/1472
- I could not resist looking into this 🙂
- we will need to investigate the root cause, too - why the request is missing from the cache... but it might be just the platform being its usual self


Pijukatel:
Seems to me, that this could be the root cause:
https://github.com/apify/apify-sdk-python/issues/630

PR #629, opened 2025-10-16
Update test to included statistics before reboot

Pijukatel:
Temporarily point to unmerged crawlee version that contains the fix.

TODO: Finalize PR after Crawlee change is merged and released.


Pijukatel:
Will be handled in https://github.com/apify/apify-sdk-python/pull/646

PR #639, opened 2025-10-21
docs: Improve docs of Apify storage clients and export SQL storage client

vdusek:
- Export `SqlStorageClient` from Crawlee.
- Improve docs of `ApifyStorageClient` and `SmartApifyStorageClient`.

### DIFF
--- a/src/apify/storage_clients/_apify/_storage_client.py
+++ b/src/apify/storage_clients/_apify/_storage_client.py
@@ -36,12 +36,18 @@ class ApifyStorageClient(StorageClient):
 
     ### Single mode
 
-    The `single` mode is optimized for scenarios with only one consumer. It makes fewer API calls and is therefore
-    faster and more cost-efficient, but it comes with several constraints. Only one client should consume the request
-    queue at a time. Multiple producers can add new requests, but forefront requests might not be processed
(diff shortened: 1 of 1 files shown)


src/apify/storage_clients/_apify/_storage_client.py
```diff
@@ -21,23 +21,44 @@
...
+
+    The request queue client supports two access modes controlled by the `request_queue_access` parameter:
+
+    ### Single mode
+
+    The `single` mode is optimized for scenarios with only one consumer. It makes fewer API calls and is therefore
+    faster and more cost-efficient, but it comes with several constraints. Only one client should consume the request
+    queue at a time. Multiple producers can add new requests, but forefront requests might not be processed
+    immediately, since this mode relies on local head estimation rather than frequent forefront fetching. Requests
+    can be added or marked as handled by other clients, but they must not be deleted or modified, as such changes
```
Pijukatel: > marked as handled by other clients

Depending on the scenario this could be ignored by the client. If the request is already fully cached locally them marking it as handled by other client will be ignored by this client. It is not going to break the client, but it could sometimes lead to processing request that was already marked as handled by other client.

If the request was not yet cached locally, then marking it as handled is not an issue.

PR #653, opened 2025-10-29
fix: make the `ActorRun.stats` optional

Mantisus:
This PR fixes a validation error for the response received from the endpoint - https://docs.apify.com/api/v2/actor-run-get
When, for some reason, the `stats` data is missing in the response. However, this does not comply with the API documentation.


janbuchar:
Closing this because a fix on the platform side in under way

Issue #641, opened 2025-10-22
Improve Pytest environment handling in `Actor._get_default_exit_process`


vdusek:
### Description

Method [`Actor._get_default_exit_process`](https://github.com/apify/apify-sdk-python/blob/v3.0.3/src/apify/_actor.py#L1292:L1310) currently contains the following logic:

```python
# Check if running in Pytest by detecting the relevant environment variable.
if os.getenv('PYTEST_CURRENT_TEST'):
    self.log.debug('Running in Pytest, setting default `exit_process` to False.')
    return False
```

This behavior is incorrect - production code should not adjust its defaults based on test environment detection. The logic should be inverted: instead of the SDK adapting to Pytest during normal execution, the test environment should explicitly configure the `exit_process` flag as needed.

### Expected fix:
 
- Remove the Pytest-specific condition from the production code.
- Handle `exit_process=False` setup within the test environment (e.g., via test fixtures or explicit configuration).

PR #654, opened 2025-10-30
chore: remove the logic of `Actor._get_default_exit_process` for pytest from production code

Mantisus:
### Issues

Closes: #641

### DIFF
--- a/tests/unit/actor/test_actor_log.py
+++ b/tests/unit/actor/test_actor_log.py
@@ -36,7 +36,7 @@ async def test_actor_logs_messages_correctly(caplog: pytest.LogCaptureFixture) -
             # Test that exception in Actor.main is logged with the traceback
             raise RuntimeError('Dummy RuntimeError')
 
-    # We skip the first entry, as it is related to the initialization of `lazy_object_proxy.Proxy`` for `Actor`.
+    # We skip the first entry, as it is related to the initialization of `lazy_object_proxy.Proxy` for `Actor`.
     records = caplog.records[1:]
 
     # Updated expected number of log records (additional debug messages added)


tests/unit/conftest.py
```diff
@@ -226,3 +226,21 @@ def proxyless_async_client(*args: Any, **kwargs: Any) -> impit.AsyncClient:
     monkeypatch.setattr(impit, 'AsyncClient', proxyless_async_client)
     yield
     monkeypatch.undo()
+
+
+@pytest.fixture(autouse=True)
+def patched_actor_exit_proc(monkeypatch: pytest.MonkeyPatch) -> Iterator[None]:
```
Pijukatel: Why not just this?


```
@pytest.fixture(autouse=True)
def mock_actor_sys_exit():
    with mock.patch('apify._actor.sys.exit') as mock_exit:
        yield mock_exit
```

Mantisus: [Test](https://github.com/apify/apify-sdk-python/blob/master/tests/unit/actor/test_actor_log.py#L14) checks behavior in different environments.

This PR fully preserves the old behavior and does not directly affect the tests.

Pijukatel: Yes, but I think it is desired to slightly modify the tests since we are not really interested in log containing such artificial entries like `Running in Pytest`

On top of it, the fixture probably does not have to be `autosue` and can be applied only to relevant tests. I tried running without the fixture, and actually, it seems to me that only the test you have mentioned needs the fixture.

Mantisus: Yes, it seems that this is only used for this test and only to obtain more controllable logs.

Mantisus: I dropped it and just updated the test.

Issue #529, opened 2025-08-05
Improve Apify RQ storage client resilience to concurrent processing


vdusek:
The current implementation of the Apify Request Queue storage client (both in the v2 and v3 versions of the SDK) is not resilient to concurrent processing scenarios, which leads to flaky behavior in tests that simulate parallel workers. This is evident in `test_concurrent_processing_simulation`, which has been skipped.

PR #662, opened 2025-11-04
chore: Use `shared` mode `ApifyStorageClient` for `test_concurrent_processing_simulation`

Mantisus:
Use `ApifyStorageClient(request_queue_access='shared')` for `test_concurrent_processing_simulation`. This is the expected usage for concurrent access. And fix the error in the test

### Issues

Closes: #529

### DIFF
--- a/tests/integration/actor/test_actor_request_queue.py
+++ b/tests/integration/actor/test_actor_request_queue.py
@@ -417,17 +417,20 @@ async def test_concurrent_processing_simulation(
             # Simulate concurrent workers
             async def worker() -> int:
                 processed = 0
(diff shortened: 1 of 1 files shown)


Pijukatel:
> > Could you stress test it a little locally since the CI is now not running integration tests on fork PRs?
> 
> Out of about 70 runs, this test failed once due to an actor timeout (execution time exceeding 10 minutes).
> 
> But there really is a problem. Normally, it takes less than 10 seconds to execute, but in some runs, it takes up to 5 minutes.
> 
> UPD: I will try to find the reasons for this.

Have you been able to figure this out?

PR #663, opened 2025-11-05
refactor: Extract API storage client creation from the Apify storage code

Pijukatel:
### Description

- Extract API storage client creation from the Apify storage `open` method to `ApiClientFactory`
- Create specific variants of the factory and use them in `open` methods of Apify storages.


src/apify/storage_clients/_apify/_utils.py
```diff
@@ -219,3 +237,103 @@ def create_apify_client(configuration: Configuration) -> ApifyClientAsync:
...
+
+            case (str(), None, None, _):
+                return await self._open_by_alias(self._alias)
+
+            case (None, None, None, str()):
+                resource_client = self._get_resource_client(id=self._default_id)
+                raw_metadata = await resource_client.get()
+                metadata = self._get_metadata(raw_metadata)
+                if not raw_metadata:
+                    # Do we want this??? Backwards compatibility, so probably yes.
```
Pijukatel: One edge case from the discussion

src/apify/storage_clients/_apify/_utils.py
```diff
@@ -219,3 +237,103 @@ def create_apify_client(configuration: Configuration) -> ApifyClientAsync:
...
+            id = await _alias.resolve_id()
+
+            if id:
+                # There was id, storage has to exist, fetch metadata to confirm it.
+                resource_client = self._get_resource_client(id=id)
+                raw_metadata = await resource_client.get()
+                if raw_metadata:
+                    return resource_client, self._get_metadata(raw_metadata)
+                # If we do not raise here, we will behave same as for default storage. We create it even though it
+                # should exist already. Consistency or throw an error???
```
Pijukatel: The other edge case from the discussion

src/apify/storage_clients/_apify/_api_client_factory.py
```diff
@@ -0,0 +1,183 @@
...
+    configuration: Configuration,
+    id: str | None = None,
+    name: str | None = None,
+    alias: str | None = None,
+) -> DatasetClientAsync: ...
+
+
+async def create_api_client(
+    *,
+    storage_type: Literal['key_value_store', 'request_queue', 'dataset'],
```
Pijukatel: Is there any benefit in keeping this virtual mapping between class and literal?
Why not use the class directly?

src/apify/storage_clients/_apify/_api_client_creation.py
```diff
@@ -0,0 +1,183 @@
...
+        name: Storage name (global scope, persists across runs). Mutually exclusive with id and alias.
+        alias: Storage alias (run scope, creates unnamed storage). Mutually exclusive with id and name.
+
+    Returns:
+        The storage client for the opened or created storage.
+
+    Raises:
+        ValueError: If configuration is invalid, multiple identifiers are provided, or storage cannot be found.
+    """
+    if sum(1 for param in [id, name, alias] if param is not None) > 1:
```
Pijukatel, line 75: The function is very long, so I would extract at least the first part to a standalone method to improve the readability. 
`def _create_api_client`

src/apify/storage_clients/_apify/_request_queue_client.py
```diff
@@ -225,11 +222,22 @@ async def open(
...
-        ).get_client_with_metadata()
+        api_client = await create_api_client(
+            storage_type='request_queue',
+            configuration=configuration,
+            alias=alias,
+            name=name,
+            id=id,
+        )
+
+        # Fetch metadata separately
```
Pijukatel, line 233: This is an additional API call for something that we already had available in `create_api_client`. I agree that the method that returns a tuple is worse than the method that returns a single value, but it is basically an internal factory method used in three and only three places, so in my view, the practicality beats purity in this instance. Especially since the PR was partially triggered by a Slack discussion about excessive API calls during storage creation.

src/apify/storage_clients/_apify/_api_client_factory.py
vdusek: import hell (circular imports) - this is a better approach, trust me, I've been there :smile:

src/apify/storage_clients/_apify/_request_queue_client.py
vdusek: Only the request queue client needs this, and it's just one additional call during creation, so I'd say it's not worth it.

src/apify/storage_clients/_apify/_api_client_creation.py
vdusek: Okay, we can split "create_storage_api_client" and "create_api_client".

src/apify/storage_clients/_apify/_alias_resolving.py
```diff
@@ -0,0 +1,272 @@
...
+        storage_id = await alias_resolver.resolve_id()
+
+        if storage_id:
+            # Check if storage with this ID exists
+            resource_client = get_resource_client_by_id(storage_id)
+            raw_metadata = await resource_client.get()
+            if raw_metadata:
+                return resource_client
+
+        # Create new unnamed storage and store alias mapping
```
Pijukatel, line 105: Why is this handled differently compared to the re-creation of the missing default storage, which is handled like this
```
            # Default storage does not exist. Create a new one.
            if not raw_metadata:
                raw_metadata = await collection_client.get_or_create()
                resource_client = get_resource_client(raw_metadata['id'])
```

vdusek: It is a different case. For the default storage, we either find it or create it. But for the ID, you can either find it or fail.

Pijukatel: We are already past the code that deals with a specific ID. If it finds it, it returns it, if not we get to this code, which creates new storage. `raw_metadata = await collection_client.get_or_create()` creates new storage with some new id

PR #664, opened 2025-11-06
fix: Fix crash in `Actor.push_data` with PPE and a strict charging limit

janbuchar:
- reported [via Slack](https://apify.slack.com/archives/CGZSN9DQC/p1762419168355659)

### DIFF
--- a/src/apify/_actor.py
+++ b/src/apify/_actor.py
@@ -614,18 +614,20 @@ class _ActorType:
             else None
         )
 
+        # Push as many items as we can charge for
+        pushed_items_count = min(max_charged_count, len(data)) if max_charged_count is not None else len(data)
+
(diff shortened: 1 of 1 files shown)


Mantisus (approved):
LGTM

Pijukatel (approved):
It would be nice to add test for the combination of inputs that caused the bug.

janbuchar:
> It would be nice to add test for the combination of inputs that caused the bug.

the steps to reproduce it were basically to run an Actor with a spending limit that makes it impossible to charge even a single occurence of an event and then call `push_data` with that event. This triggers a `push_data` call with `count=0`. And now that I think of it, this may not have resolved the actual reported issue :grin:

PR #668, opened 2025-11-07
fix: Fix Actor.charge behavior when the budget is overdrawn

janbuchar:
- follow up to #665
- I believe this fixes the original issue for good :disappointed_relieved:


vdusek (approved):
LGTM

Pijukatel (approved):

Issue #666, opened 2025-11-07
Synchronize `Actor.charge` calls to prevent race conditions in `Actor.push_data`


janbuchar:
- calls to `ChargingManager.charge` should be safe because the "critical section" does not yield
- however, when calling PPE-aware `Actor.push_data` there might be some interleaving of the `push_data` and `charge` API calls that may lead to unintended behavior
- it would probably be best to hold all `Actor.charge` calls if there is an `Actor.push_data` in progress

PR #684, opened 2025-11-20
fix: Synchronize charge operations to prevent race conditions

vdusek:
Closes: #666

### DIFF
--- a/src/apify/_actor.py
+++ b/src/apify/_actor.py
@@ -355,6 +355,14 @@ class _ActorType:
         return ChargingManagerImplementation(self.configuration, self.apify_client)
 
     @cached_property
+    def _charge_lock(self) -> asyncio.Lock:
+        """Lock to synchronize charge operations.
+
+        Prevents race conditions between Actor.charge and Actor.push_data calls.
+        """
+        return asyncio.Lock()
+
+    @cached_property
     def _storage_client(self) -> SmartApifyStorageClient:
         """Storage client used by the Actor.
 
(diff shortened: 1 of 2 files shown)


src/apify/_charging.py
```diff
@@ -223,58 +229,60 @@ def calculate_chargeable() -> dict[str, int | None]:
                 chargeable_within_limit=calculate_chargeable(),
             )
 
-        # START OF CRITICAL SECTION - no awaits here
+        # Acquire lock to prevent race conditions between concurrent charge calls
+        # (e.g., when Actor.push_data with charging is called concurrently with Actor.charge).
+        async with self._charge_lock:
```
janbuchar: This, on its own, doesn't fix the issue of multiple interleaved, PPE-aware `Actor.push_data` calls. The lock would need to be held for the whole duration of the push_data+charge sequence.

vdusek: Thanks, so I moved the lock to the Actor class, take a look.

src/apify/_actor.py
```diff
@@ -606,30 +614,34 @@ async def push_data(self, data: dict | list[dict], charged_event_name: str | Non
...
+        if charged_event_name is None:
+            dataset = await self.open_dataset()
+            await dataset.push_data(data)
+            return None
 
-        # Push as many items as we can charge for
-        pushed_items_count = min(max_charged_count, len(data)) if max_charged_count is not None else len(data)
+        # If charging is requested, acquire the charge lock to prevent race conditions between concurrent
+        # push_data calls. We need to hold the lock for the entire push_data + charge sequence.
+        async with self._charge_lock:
```
janbuchar, line 625: In theory, this could still get interleaved with an `Actor.charge` call with an unrelated event, which would result in pushing more items than the user can afford.

A watertight solution would be to use a read-write lock where plain `Actor.charge` would acquire the read-lock and `Actor.push_data` would acquire the write-lock.

But since there is no built-in read-write lock for asyncio, we should probably just leave this the way it is 🙂

janbuchar (approved):
LGTM

PR #689, opened 2025-11-25
feat: add `ACTOR_PERMISSION_LEVEL` to the configuration

### LINKED ISSUE #690: Read `ACTOR_PERMISSION_LEVEL` envvar
The Apify worker environment sets a new variable `ACTOR_PERMISSION_LEVEL`. 

parity tracking issue for https://github.com/apify/apify-sdk-js/pull/516

stepskop:
This adds the `ACTOR_PERMISSION_LEVEL` environmental variable as `actor_permission_level` to the Configuration,

Closes #690

### DIFF
--- a/src/apify/_configuration.py
+++ b/src/apify/_configuration.py
@@ -64,7 +64,6 @@ class Configuration(CrawleeConfiguration):
     actor_permission_level: Annotated[
         str | None,
         Field(
-            alias='actor_permission_level',
             description='Permission level the Actor is run under.',
         ),
     ] = None


tests/unit/actor/test_actor_log.py
stepskop, line 1: 😅

src/apify/_actor.py
stepskop, line 1: Same as for JS SDK...
I am not sure if this should not be somehow conditional, as most of the time it would be defined only when run on Apify. Locally, it won't be defined if the user doesn't do so.

However, I can imagine some cases when my code depends on the permission level, so I am passing it "artificially" even locally to test things out.

danpoletaev (commented):
Same as for JS SDK, consider moving log message to worker 🙃

src/apify/_configuration.py
```diff
@@ -61,6 +61,17 @@ class Configuration(CrawleeConfiguration):
         ),
     ] = None
 
+    actor_permission_level: Annotated[
+        str | None,
+        Field(
+            validation_alias=AliasChoices(
+                'actor_permission_level',
+                'apify_actor_permission_level',
```
janbuchar: Does it really make sense to have multiple choices for a newly introduced variable? The aliases exist mainly for compatibility with outdated documentation 😁

stepskop: Actually, I wasn't sure how the name was inferred, so it was just to **be** sure 😅.

So in this case, the `actor_permission_level` should be enough, right?

The name of the env var is `ACTOR_PERMISSION_LEVEL` and it is part of `ActorEnvVars` in `apify-shared-python`.

https://github.com/apify/apify-shared-python/blob/782a8334e2d0e5de38aa28e10f4f1f1038497955/src/apify_shared/consts.py#L136

e.g.: I assume that it won't be prefixed.

janbuchar: yeah, we can keep just `actor_permission_level` and drop `apify_actor_permission_level`

src/apify/_configuration.py
```diff
@@ -61,6 +61,14 @@ class Configuration(CrawleeConfiguration):
         ),
     ] = None
 
+    actor_permission_level: Annotated[
+        str | None,
+        Field(
+            alias='actor_permission_level',
```
vdusek: I believe the alias has no purpose here if it only matches the attribute name.

stepskop: I see. So if I get it correctly, we don't need it since the name of the property would be the same. Cool.

vdusek (approved):
lgtm

Issue #703, opened 2025-12-08
Test `test_actor_handles_migrating_event_correctly` is flaky


vdusek:
- Linux Python 3.14.
- Failing log run e.g. [here](https://github.com/apify/apify-sdk-python/actions/runs/20029622226/job/57435192024?pr=702)
- Failing test summary:
```
=========================== short test summary info ============================
FAILED tests/unit/actor/test_actor_lifecycle.py::test_actor_handles_migrating_event_correctly - assert EventPersistStateData(is_migrating=False) == EventPersistStateData(is_migrating=True)
  
  Full diff:
  - EventPersistStateData(is_migrating=True)
  ?                                    ^^^
  + EventPersistStateData(is_migrating=False)
  ?                                    ^^^^
================= 1 failed, 244 passed, 336 warnings in 22.38s =================
```

PR #704, opened 2025-12-08
chore: Fix flaky `test_actor_handles_migrating_event_correctly`

Pijukatel:
### Description:
Do run run persist state event every 500 ms to prevent flakiness. It is enough to check the persist state event we send manually and the Crawler final one.

### Issues:

Closes: #703

### DIFF
--- a/tests/unit/actor/test_actor_lifecycle.py
+++ b/tests/unit/actor/test_actor_lifecycle.py
@@ -284,9 +284,8 @@ async def test_actor_handles_migrating_event_correctly(monkeypatch: pytest.Monke
 
                 await asyncio.sleep(1)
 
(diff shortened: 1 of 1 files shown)


tests/unit/actor/test_actor_lifecycle.py
```diff
@@ -285,7 +284,7 @@ async def handler(websocket: websockets.asyncio.server.ServerConnection) -> None
 
                 await asyncio.sleep(1)
 
-    assert len(persist_state_events_data) >= 3
+    assert len(persist_state_events_data) >= 2
```
vdusek, line 288: > It is enough to check the persist state event we send manually and the Crawler final one.

Could we just please add a comment describing this?

PR #707, opened 2025-12-11
chore: Add test coverage reports

vdusek:
- Add a code coverage badge to the README.
- Update the `pre_release` and `run_code_checks` test pipeline invocations.
- Remove the unit test call from the `release` workflow. Since this workflow is triggered only manually, the caller should ensure the package is OK (and also often the flakiness factor).
- Temporarily skip four newly failing integration tests to unblock coverage - these will be investigated right after this.
- Update pytest invocations in the Makefile.
- Collect coverage from both unit and integration tests and combine them into a single report.
- PRs for Crawlee and Client will follow.
- `CODECOV_TOKEN` is a repository-level secret.
- Relates to: https://github.com/apify/crawlee-python/issues/767

### DIFF
--- a/.github/workflows/pre_release.yaml
+++ b/.github/workflows/pre_release.yaml
@@ -55,7 +55,7 @@ jobs:
 
   unit_tests:
     name: Unit tests
-    uses: apify/workflows/.github/workflows/python_unit_tests.yaml@add-code-cov-to-python
+    uses: apify/workflows/.github/workflows/python_unit_tests.yaml@main
     secrets: inherit
     with:
       python-versions: '["3.10", "3.11", "3.12", "3.13", "3.14"]'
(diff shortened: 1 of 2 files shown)


janbuchar (approved):

.github/workflows/release.yaml
```diff
@@ -55,12 +55,6 @@ jobs:
     with:
       python-versions: '["3.10", "3.11", "3.12", "3.13", "3.14"]'
 
-  unit_tests:
```
janbuchar, line 58: On second thought, are you sure? I think the optimal approach is to check if the branch is green...

vdusek: I'm sure. I do most of the releases, and having to restart them multiple times due to flaky tests (mostly in Crawlee) is quite painful. Since this workflow is triggered manually, I think it's reasonable to expect the caller to ensure the branch is already green. For that reason, I'd prefer to remove the test step here.

PR #709, opened 2025-12-12
test: Fix Actor logging integration test

vdusek:
- Fix Actor logging integration test
- The remaing 3 are caused by limited permissions - follow-up issue: https://github.com/apify/apify-sdk-python/issues/715

### DIFF
--- a/tests/integration/actor/test_actor_log.py
+++ b/tests/integration/actor/test_actor_log.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING
 
-from apify import Actor, __version__
+from apify import Actor
 
 if TYPE_CHECKING:
     from .conftest import MakeActorFunction, RunActorFunction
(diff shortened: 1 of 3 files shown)


tests/integration/actor/test_actor_log.py
```diff
@@ -53,41 +50,25 @@ async def main() -> None:
...
-    assert run_log_lines.pop(0) == '      Traceback (most recent call last):'
-    assert run_log_lines.pop(0) == '        File "/usr/src/app/src/main.py", line 33, in main'
-    assert run_log_lines.pop(0) == "          raise RuntimeError('Dummy RuntimeError')"
-    assert run_log_lines.pop(0) == '      RuntimeError: Dummy RuntimeError'
-    assert run_log_lines.pop(0) == '[apify] INFO  Exiting Actor ({"exit_code": 91})'
+    # Join all lines to make it easier to search for expected content
+    full_log = '\n'.join(run_log_lines)
+
+    # Verify expected log messages are present (order-independent checks)
+    assert '[apify] DEBUG Debug message' in full_log
```
janbuchar, line 60: I will allow this. Honestly, I'd be fine if you just removed this test 😄

vdusek: Well, the logging can be problematic due to our custom configuration in Actor -> https://github.com/apify/apify-sdk-python/blob/master/src/apify/_actor.py#L169-L172 . We already had cases with log statements before `configure_logging`, which were supposed to log info messages, but it wasn't working, and this went unnoticed for a long time. Because of that, I would prefer to keep this test :).

PR #712, opened 2025-12-14
chore(ci): Align GitHub workflows across Python projects

vdusek:


.github/workflows/_release_docs.yaml
```diff
@@ -25,29 +31,12 @@ jobs:
         uses: actions/checkout@v6
         with:
           token: ${{ secrets.APIFY_SERVICE_ACCOUNT_GITHUB_TOKEN }}
+          ref: ${{ github.event_name == 'workflow_call' && inputs.ref || github.ref }}
 
       - name: Set up Node
         uses: actions/setup-node@v6
         with:
           node-version: ${{ env.NODE_VERSION }}
-          cache: npm
```
janbuchar, line 33: What was this for, do we really not need it?

.github/workflows/manual_release_stable.yaml
```diff
@@ -102,6 +95,9 @@ jobs:
       - name: Publish package to PyPI
         uses: pypa/gh-action-pypi-publish@release/v1
 
+      # TODO: add job for publish package to Conda
```
janbuchar: I don't think this is necessary?

.github/workflows/_release_pre.yaml
```diff
@@ -72,3 +64,6 @@ jobs:
       # Publishes the package to PyPI using PyPA official GitHub action with OIDC authentication.
       - name: Publish package to PyPI
         uses: pypa/gh-action-pypi-publish@release/v1
+
+      # TODO: add job for publish package to Conda
```
janbuchar: This, too, seems unnecessary

.github/workflows/_release_docs.yaml
vdusek: Fuh, thanks for noticing!

vdusek: I fixed this and the rest in other PRs as well

PR #732, opened 2026-01-08
docs: merge overview pages

TC-MO:


docs/01_overview/index.mdx
```diff
@@ -0,0 +1,163 @@
+---
```
marcel-rbro, line 1: General note:
Is it possible to move the overview page one level up, so that it is not in the Overview section? It's kinda weird UX-wise.
image

docs/01_overview/index.mdx
```diff
@@ -0,0 +1,163 @@
...
+handler.setFormatter(ActorLogFormatter())
+${''}
+apify_logger = logging.getLogger('apify')
+apify_logger.setLevel(logging.DEBUG)
+apify_logger.addHandler(handler)
+${''}
+asyncio.run(main())`
+        }</CodeBlock>
+    </TabItem>
+</Tabs>
```
marcel-rbro, line 120: Consider adding some code comments?

docs/01_overview/index.mdx
```diff
@@ -0,0 +1,163 @@
...
+When you create an Actor using the Apify CLI, the Apify SDK for Python is installed for you automatically.
+
+If you are not developing Apify Actors and you just need to access the Apify API from Python,
+consider using the [Apify API client for Python](https://docs.apify.com/api/client/python) directly.
+
+## Quick start
+
+### Creating Actors
+
+To create and run Actors through Apify Console, refer to the [Console documentation](https://docs.apify.com/academy/getting-started/creating-actors#choose-your-template).
```
marcel-rbro: "through Apify Console"
Is through used in this context? I'd rather just use "in".

```suggestion
To create and run Actors in Apify Console, refer to the [Console documentation](https://docs.apify.com/academy/getting-started/creating-actors#choose-your-template).
```

docs/01_overview/index.mdx
```diff
@@ -0,0 +1,163 @@
...
+    async with Actor:
+        input = await Actor.get_input()
+        response = requests.get(input['url'])
+        soup = BeautifulSoup(response.content, 'html.parser')
+        await Actor.push_data({ 'url': input['url'], 'title': soup.title.string })
+```
+
+## Requirements
+
+The Apify SDK requires Python version 3.8 or above to run Python Actors locally.
```
marcel-rbro: The docs previously said that the required version is 3.10 or above. Is this updated version correct?

docs/01_overview/index.mdx
```diff
@@ -0,0 +1,163 @@
...
+        }</CodeBlock>
+    </TabItem>
+</Tabs>
+
+If you want to modify the Actor structure, you need to make sure that your Actor is executable as a module, via `python -m src`, as that is the command started by `apify run` in the Apify CLI.
+We recommend keeping the entrypoint for the Actor in the `src/__main__.py` file.
+
+## Adding dependencies
+
+First, add them in the [`requirements.txt`](https://pip.pypa.io/en/stable/reference/requirements-file-format/) file in the Actor source folder.
```
marcel-rbro:
```suggestion
First, add the dependencies in the [`requirements.txt`](https://pip.pypa.io/en/stable/reference/requirements-file-format/) file in the Actor source folder.
```

docs/01_overview/index.mdx
```diff
@@ -0,0 +1,163 @@
...
+        }</CodeBlock>
+    </TabItem>
+    <TabItem value="win" label="Windows">
+        <CodeBlock language="powershell">{
+`.venv\\Scripts\\activate`
+        }</CodeBlock>
+    </TabItem>
+</Tabs>
+
+Then install the dependencies:
```
marcel-rbro:
```suggestion
Finally, install the dependencies:
```

docs/01_overview/index.mdx
TC-MO: Yeah I was afraid that might happen. I think if we rename it to index.md it should not be separate openable section

docs/01_overview/index.mdx
TC-MO: Good point, I'll reach out and confirm the required version as client & sdk diverge here

vdusek (changes requested):
When running it locally, clicking "Docs" or "Get Started" redirects to
http://localhost:3000/sdk/python/docs/overview/introduction, which results in a 404.

Image

docs/01_overview/index.mdx
TC-MO: based on `uv.lock` seems like 3.10 or above is required @vdusek can you confirm so we can update accordingly?

B4nan: we test only on python 3.10+, afaik we dropped python 3.9 with the current v3 line

TC-MO: Thank you , I'll update accordingly that 3.10 or above is required

docs/01_overview/index.mdx
```diff
@@ -0,0 +1,163 @@
...
+sidebar_label: Overview
+---
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+import CodeBlock from '@theme/CodeBlock';
+
+The Apify SDK for Python is the official library for creating [Apify Actors](https://docs.apify.com/platform/actors) in Python.
+
+```python
```
patrikbraborec: Does it have make sense to have the code example as the first thing in the overview?

docs/01_overview/index.mdx
```diff
@@ -0,0 +1,163 @@
...
+When you create an Actor using the Apify CLI, the Apify SDK for Python is installed for you automatically.
+
+If you are not developing Apify Actors and you just need to access the Apify API from Python,
+consider using the [Apify API client for Python](https://docs.apify.com/api/client/python) directly.
+
+## Quick start
+
+### Creating Actors
+
+To create and run Actors in Apify Console, refer to the [Console documentation](https://docs.apify.com/academy/getting-started/creating-actors#choose-your-template).
```
patrikbraborec: I think we should refer to the platform docs, not academy imho

docs/01_overview/index.mdx
```diff
@@ -0,0 +1,163 @@
...
+
+For example, to create an Actor from the "[beta] Python SDK" template, you can use the [`apify create` command](https://docs.apify.com/cli/docs/reference#apify-create-actorname).
+
+```bash
+apify create my-first-actor --template python-start
+```
+
+This will create a new folder called `my-first-actor`, download and extract the "Getting started with Python" Actor template there, create a virtual environment in `my-first-actor/.venv`, and install the Actor dependencies in it.
+
+### Running the Actor
```
patrikbraborec: I think we could show example of the result (i.e. terminal output)

docs/01_overview/index.mdx
```diff
@@ -0,0 +1,163 @@
...
+To run the Actor, you can use the [`apify run` command](https://docs.apify.com/cli/docs/reference#apify-run):
+
+```bash
+cd my-first-actor
+apify run
+```
+
+This will activate the virtual environment in `.venv` (if no other virtual environment is activated yet), then start the Actor, passing the right environment variables for local running, and configure it to use local storages from the `storage` folder.
+
+The Actor input, for example, will be in `storage/key_value_stores/default/INPUT.json`.
```
patrikbraborec, line 75: This is quite important information, and we should enhance it as many people struggle with this (they do not know where they should define INPUT.json locally)

TC-MO: perhaps an admonition, let me try few different versions

TC-MO: Admonition wouldn't work after all since I would have to. do it for whole paragraph. I opted to use bullet list (even though I hate them :D) to better flow with the whole doc, while making the content in question more scannable and accentuated. Let me know if you think any further changes are needed

patrikbraborec: Thank you!

vdusek:
> @vdusek should we be worried about tests failing or do they have no bearing on docs change that we want to do?

I'll restart it, but there's no need to worry - the (integration) tests rely only on the codebase, not the docs.

Issue #735, opened 2026-01-13
Consider adding `Actor.use_state`


Pijukatel:
Context: https://github.com/apify/apify-docs/pull/2176#discussion_r2686522872

B4nan:
@Mantisus can you look into this please?

barjin:
Given we want to [split the Crawlee state](https://github.com/apify/crawlee-python/issues/1627) to have a separate one for each Crawler instance, it doesn't really make sense to me to add SDK support on top of this.

JS and Python SDK sharing the same state (key) on `Actor.useState` / `Actor.use_state` sounds nice and IMO sells the idea it's actually the "Actor's state", not some random KVS value.

PR #738, opened 2026-01-16
feat: Add `Actor` method `use_state`

Mantisus:
### Description

Add `Actor` method `use_state`. Allows the user to save data and states in kvs with periodic persistence.

### Issues

Closes: #735

### Testing

Added unit tests


vdusek:
Let's please wait until the implementation is clarified in https://github.com/apify/apify-sdk-python/issues/735.

PR #746, opened 2026-01-21
test: Add pytest-rerunfailures for integration tests

vdusek:
Add `pytest-rerunfailures` as a dev dependency and configure all integration tests to rerun up to 3 times on failure automatically. This helps handle occasional platform-related instabilities without manual test reruns.

Uses `pytest_collection_modifyitems` hook to apply the flaky marker to all integration tests, avoiding the need to decorate each test individually.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -72,6 +72,7 @@ dev = [
     "pytest-asyncio<2.0.0",
     "pytest-cov<8.0.0",
     "pytest-httpserver<2.0.0",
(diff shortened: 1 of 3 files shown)


Pijukatel (changes requested):
I would prefer to apply any retries only on per-test basis. Most often I feel that the flakiness is due to specific test design or sometimes it actually exposes real issue. Such blanket retry could lead to hiding issues.

Can you please link some example failures you want to deal with this?

PR #747, opened 2026-01-22
docs: Migrate from npm to yarn and relax web dependencies

vdusek:
- Migrate from `npm` to `yarn` for consistency.
- Same setup as in Crawlee.
- All 3 Python documentation projects (`crawlee-python`, `apify-sdk-python`, `apify-client-python`) now use an identical yarn setup.
- Relax website dependency constraints to reduce the number of unnecessary Renovate PRs.
- Switch GitHub Actions to major-version tags for the same reason (only typos, for actionlint it isn't possible, unfortunately).


website/package.json
```diff
@@ -29,26 +28,28 @@
...
-        "markdownlint": "^0.40.0",
-        "markdownlint-cli": "^0.47.0",
+        "markdownlint": ">=0.39.0 <1.0.0",
+        "markdownlint-cli": ">=0.46.0 <1.0.0",
         "path-browserify": "^1.0.1",
         "prettier": "^3.0.0",
-        "rimraf": "^6.0.0"
-    }
+        "rimraf": "^6.0.0",
+        "typescript": "^5.0.0"
```
B4nan: this is not python :] you dont have to add upper constraints for major versions, it works like this by design. 0.x can be a breaking release, so npm/yarn/whatever will only install things it its range (only patches for those). please remove this, IMO it's a noise. lets stick to `^`, i dont think we use this "python way" anywhere in JS projects

B4nan: i think i understand what were you trying to do, but i kinda doubt it would work this way with renovate (you would still get PRs for update from 0.39 to 0.40), not to mention it feels a bit dangerous (although, docs, they either build or not, but it sometimes happens you get the build passing but the behaviour changes and you get visual bugs). we want to get PRs for any 0.x bumps, those can be breaking, technically those are major releases, and we want PRs for any major releases

pyproject.toml
```diff
@@ -240,10 +240,14 @@ shell = "uv run ruff format --check && uv run ruff check"
 [tool.poe.tasks.format]
 shell = "uv run ruff check --fix && uv run ruff format"
 
+[tool.poe.tasks.update-docs-theme]
+shell = "corepack enable && yarn up @apify/docs-theme"
+cwd = "website"
+
 [tool.poe.tasks.build-docs]
-shell = "./build_api_reference.sh && npm ci && npm run build"
+shell = "./build_api_reference.sh && corepack enable && yarn && uv run yarn build"
```
B4nan, line 248: why `uv run yarn build`? you can do `yarn build`, why make it longer? 🙃 or is there any difference?

barjin: afaik we're calling the Python code analyzers in the docs build (for the API docs) and `uv run` runs the script in the correct project environment (with the right dependencies installed).

We're already running `poe` with `uv run` in the CI, but locally this still might be useful imo

B4nan: ok, thought its gonna be something like that, lets keep it then

website/package.json
vdusek: > this is not python :] you dont have to add upper constraints for major versions, it works like this by design. 0.x can be a breaking release, so npm/yarn/whatever will only install things it its range (only patches for those)

I know what the `^` operator does, including how it works with zero-based versions :slightly_smiling_face: (it is even a valid operator in `poetry`, which we were using before `uv`).  So all of these changes are intentional and have a reason.

The reason is that Renovate tends to "spam" updates for `markdownlint` and `markdownlint-cli`, which release new minor (major) versions almost monthly. As a result, we end up with dedicated PRs just to update a dev dependency of the documentation website (which btw. we don't even use anywhere 😄).

Also, this is a 10+ year old project that likely intends to stay on zero-based versioning forever (which might make sense for linters).

(Maybe production dependencies (`docusaurus-gtm-plugin`), or other dev dependencies (`@apify/eslint-config-ts`, `@apify/tsconfig`) can be left as they are, as there are no new "major" releases that often.)

> not to mention it feels a bit dangerous

Could you elaborate on this - in what way would it be dangerous? With a lockfile in place, we always install the exact pinned version, right? Even if a breaking change were introduced in a minor or even patch release, it wouldn't affect us until the lockfile is updated. Meaning we would find out that there is an issue during the lock file maintenance PR from renovate.

> (although, docs, they either build or not, but it sometimes happens you get the build passing, but the behaviour changes and you get visual bugs)

On top of what I wrote above, the dev dependencies shouldn't affect that, right?

B4nan: > which btw. we don't even use anywhere 😄

Then let's drop it and reintroduce it once we actually use it. Or use it right ahead.

> Could you elaborate on this - in what way would it be dangerous? With a lockfile in place, we always install the exact pinned version, right? Even if a breaking change were introduced in a minor or even patch release, it wouldn't affect us until the lockfile is updated. Meaning we would find out that there is an issue during the lock file maintenance PR from renovate.

This is not about lockfiles, but about automatic dependency updates. Build can pass, but there might be visual differences. Of course, not from a package like markdown lint.

The important bit here is that I doubt this will work the way you expect. You will still get PRs for those major bumps from 0.x regardless of the upper bound. Also I dont think we should just "treat 0.x releases as non breaking", they can be breaking, we do want to have separate PRs for them.

> On top of what I wrote above, the dev dependencies shouldn't affect that, right?

They surely can, a dev dep in a website can be used to produce the build, hence it can break some visuals.

vdusek: Alright, I'll remove `markdownlint` and `markdownlint-cli` for now and open a new issue to add them to CI later. This will require more work - when I tried them, they didn't pass right away, and the documentation content will need some updates to satisfy the linter.

For the other dependencies, I'll revert the changes.

Thanks for the explanation.

B4nan (approved):

PR #776, opened 2026-02-11
fix: handle TimeoutError in Actor __aexit__ to prevent resource leaks

vdusek:
## Summary
- When Actor cleanup exceeded `_cleanup_timeout`, the unhandled `TimeoutError` from `asyncio.wait_for` left the Actor in a broken state
- `_is_initialized` was never reset to `False`, event/charging managers could be left partially shut down, and `sys.exit()` was never called even when `_exit_process` was `True`
- Now the `TimeoutError` is caught, critical resource cleanup (event manager, charging manager) is forced even after timeout, and `_is_initialized` is always reset via a `finally` block

## Test plan
- [ ] Verify existing tests pass
- [ ] Test Actor exit with a very short `_cleanup_timeout` and a slow event listener to confirm the timeout path logs a warning and still completes shutdown cleanly

### DIFF
--- a/src/apify/_actor.py
+++ b/src/apify/_actor.py
@@ -255,12 +255,7 @@ class _ActorType:
         try:
             await asyncio.wait_for(finalize(), self._cleanup_timeout.total_seconds())
         except TimeoutError:
-            self.log.warning('Actor cleanup timed out, forcing shutdown of event manager and charging manager')
-            # Ensure critical resources are cleaned up even after timeout
-            with suppress(Exception):
-                await self.event_manager.__aexit__(None, None, None)
-            with suppress(Exception):
-                await self._charging_manager_implementation.__aexit__(None, None, None)
+            self.log.exception('Actor cleanup timed out')
         finally:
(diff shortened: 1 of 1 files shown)


src/apify/_actor.py
```diff
@@ -252,8 +252,17 @@ async def finalize() -> None:
...
-        self._is_initialized = False
+        try:
+            await asyncio.wait_for(finalize(), self._cleanup_timeout.total_seconds())
+        except TimeoutError:
+            self.log.warning('Actor cleanup timed out, forcing shutdown of event manager and charging manager')
+            # Ensure critical resources are cleaned up even after timeout
+            with suppress(Exception):
+                await self.event_manager.__aexit__(None, None, None)
+            with suppress(Exception):
+                await self._charging_manager_implementation.__aexit__(None, None, None)
```
janbuchar: This is just overengineering. If `finalize` times out, it is very likely that at least one of those two `__aexit__` calls will also hang. Just let the damn thing crash.

PR #777, opened 2026-02-11
fix: stop silently swallowing exceptions in request queue

vdusek:
## Summary
- Errors in `mark_request_as_handled` and `reclaim_request` in both the single and shared request queue clients were logged at `DEBUG` level, making API failures (network timeouts, auth errors, rate limits) effectively invisible
- Upgraded `logger.debug` to `logger.exception` in all 4 exception handlers so these errors are visible and include full tracebacks

## Test plan
- [x] Verify existing tests pass
- [x] Confirm that API errors in `_update_request` now produce visible log output with tracebacks

### DIFF
--- a/src/apify/storage_clients/_apify/_request_queue_shared_client.py
+++ b/src/apify/storage_clients/_apify/_request_queue_shared_client.py
@@ -235,8 +235,8 @@ class ApifyRequestQueueSharedClient:
                 processed_request=processed_request,
                 hydrated_request=request,
             )
-        except Exception as exc:
-            logger.debug(f'Error marking request {request.unique_key} as handled: {exc!s}')
+        except Exception:
+            logger.exception(f'Error marking request {request.unique_key} as handled')
             return None
         else:
             return processed_request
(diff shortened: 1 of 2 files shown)


src/apify/storage_clients/_apify/_request_queue_shared_client.py
```diff
@@ -236,7 +236,7 @@ async def mark_request_as_handled(self, request: Request) -> ProcessedRequest |
                 hydrated_request=request,
             )
         except Exception as exc:
-            logger.debug(f'Error marking request {request.unique_key} as handled: {exc!s}')
+            logger.exception(f'Error marking request {request.unique_key} as handled: {exc!s}')
```
janbuchar: Actually, no. I'd prefer to use `logger.error(..., exc_info=exc)` so that the traceback is included. `logger.exception` is a wrapper around that that pulls the "current exception" from thin air.

vdusek: Makes sense, thanks

PR #781, opened 2026-02-11
test: split integration tests into integration and e2e suites

vdusek:
## Summary

- Moved Actor-based tests (which build and deploy Actors on the Apify platform) from `tests/integration/actor/` to `tests/e2e/`
- Flattened `tests/integration/apify_api/` into `tests/integration/` (removed unnecessary nesting)
- Created dedicated `tests/e2e/conftest.py` merging fixtures from `tests/integration/conftest.py` and the old `tests/integration/actor/conftest.py`
- Added `e2e-tests` and `e2e-tests-cov` poe tasks in `pyproject.toml`
- Added `e2e_tests` CI job in `.github/workflows/_tests.yaml`
- Updated READMEs for both test directories
- Added ruff lint exception for e2e tests (matching integration tests config)

This gives a clearer separation:
- **Integration tests** (`tests/integration/`): Make real API calls but run locally, no Actor builds needed
- **E2E tests** (`tests/e2e/`): Build and deploy Actors on the Apify platform

Test collection: 77 integration + 64 e2e = 141 total (unchanged).

Relates: #785

## Test plan

- [x] `uv run poe lint` — passes
- [x] `uv run poe unit-tests` — 249 tests pass
- [x] `uv run pytest --collect-only tests/integration` — 77 tests collected
- [x] `uv run pytest --collect-only tests/e2e` — 64 tests collected
- [x] CI integration tests pass
- [x] CI e2e tests pass

### DIFF
--- a/.github/workflows/_tests.yaml
+++ b/.github/workflows/_tests.yaml
@@ -86,11 +86,10 @@ jobs:
       }}
 
     strategy:
-      # E2E tests build and run Actors on the platform. Limit parallel workflows to 2 to allow running both
-      # the oldest and newest Python versions simultaneously, while not exceeding the platform's limits. Two
-      # parallel workflows with 8 pytest workers each is a compromise between good test parallelization and
-      # not being rate-limited by the platform.
-      max-parallel: 2
+      # E2E tests build and run Actors on the platform. Limit parallel workflows to 1 to avoid exceeding
+      # the platform's memory limits. A single workflow with 16 pytest workers provides good test
+      # parallelization while staying within platform constraints.
+      max-parallel: 1
       matrix:
         os: ["ubuntu-latest"]
         python-version: ["3.10", "3.14"]
@@ -98,7 +97,7 @@ jobs:
     runs-on: ${{ matrix.os }}
 
     env:
-      TESTS_CONCURRENCY: "8"
+      TESTS_CONCURRENCY: "16"
 
     steps:
       - name: Checkout repository


janbuchar:
> What problem does this actually solve?

https://apify.slack.com/archives/C06ENBV1B9Q/p1770812487540959

.github/workflows/_tests.yaml
```diff
@@ -20,12 +20,116 @@ jobs:
...
-      python_versions: '["3.10", "3.14"]'
-      operating_systems: '["ubuntu-latest"]'
-      python_version_for_codecov: "3.14"
-      operating_system_for_codecov: ubuntu-latest
-      tests_concurrency: "16"
+    name: Integration tests (${{ matrix.python-version }}, ${{ matrix.os }})
+
+    if: >-
+      ${{
+        (github.event_name == 'pull_request' && github.event.pull_request.head.repo.owner.login == 'apify') ||
```
janbuchar, line 27: What problem does the `owner` check solve? Tests don't run in forks, do they?

.github/workflows/_tests.yaml
```diff
@@ -20,12 +20,116 @@ jobs:
...
-    with:
-      python_versions: '["3.10", "3.14"]'
-      operating_systems: '["ubuntu-latest"]'
-      python_version_for_codecov: "3.14"
-      operating_system_for_codecov: ubuntu-latest
-      tests_concurrency: "16"
+    name: Integration tests (${{ matrix.python-version }}, ${{ matrix.os }})
+
+    if: >-
+      ${{
```
janbuchar, line 26: I don't think these braces are necessary in this context, but I may be wrong.

.github/workflows/_tests.yaml
vdusek: They don't, thanks to this condition (they are skipped). Otherwise, they would be executed, and they would fail.

.github/workflows/_tests.yaml
vdusek: It is copied from the `apify/workflows`, and we know it works, so I would probably stay with that :slightly_smiling_face:.

PR #784, opened 2026-02-12
test: add e2e tests for Crawlee crawlers as Apify Actors

vdusek:
## Summary

- Add 6 new E2E tests verifying each Crawlee crawler type (BasicCrawler, HttpCrawler, BeautifulSoupCrawler, ParselCrawler, PlaywrightCrawler, AdaptivePlaywrightCrawler) works correctly when deployed as an Actor on the Apify platform.
- Each test deploys an Actor that crawls a local 5-page e-commerce test server, exercises link discovery (`enqueue_links` / `add_requests`), data extraction (`push_data`), and KVS storage (`Actor.set_value`).
- Shared test infrastructure in `conftest.py`: ASGI test server, Playwright Dockerfile template, product data expectations, and a `_verify_crawler_results` helper that checks run status, dataset contents, and KVS records.

## Motivation

- Ensuring that Crawlee crawlers can run on the Apify platform.
- There are many interactions with Apify storages, which also helps validate the Apify storage client.

## Issue

- Relates: #785

## Test plan

- [x] CI passes


Pijukatel (approved):

vdusek:
> Maybe I'll be able to check this out later in more detail, for now I'll just share one thing - we should collocate the "Test Actors" with their testcases, instead of having two big folders, actors and tests.

I absolutely agree. However, since this PR is only adding new E2E tests to the existing setup, I'd prefer not to change the structure here. Let's handle this in a dedicated PR and discuss the structure beforehand.

PR #796, opened 2026-02-16
fix: Eliminate race condition in _fetch_requests_from_url

vdusek:
## Summary

- `_fetch_requests_from_url` used `add_done_callback` + `asyncio.create_task` to process HTTP responses, but `asyncio.gather` only awaited the HTTP request tasks — not the callback-spawned processing tasks
- This caused `created_requests` to be returned before processing completed, yielding empty or incomplete results
- Refactored to gather HTTP responses first, then process each response sequentially, ensuring all extracted URLs are collected before returning

## Test plan

- [x] CI passes

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/src/apify/request_loaders/_apify_request_list.py
+++ b/src/apify/request_loaders/_apify_request_list.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import asyncio
 import re
+from itertools import chain
 from typing import Annotated, Any
 
 from pydantic import BaseModel, Field, TypeAdapter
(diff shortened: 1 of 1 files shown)


src/apify/request_loaders/_apify_request_list.py
```diff
@@ -114,49 +113,34 @@ async def _fetch_requests_from_url(
...
+        )
 
-            Use the regular expression to find all matching links in the response body, then create `Request`
-            objects from these links and the provided input attributes.
-            """
-            response = await (task.result()).read()
-            matches = re.finditer(URL_NO_COMMAS_REGEX, response.decode('utf-8'))
+        # Process each response and extract links.
+        for request_input, http_response in zip(remote_url_requests_inputs, responses, strict=True):
+            response_body = await http_response.read()
```
Pijukatel: This will do the callback processing run one by one. I am not sure in which situation this could block for a while, but maybe it would be better to do a second gather:
- gather for first tasks: send request, create callback task, append callback task to second list
- gather for second tasks: finish all callbacks

vdusek: or maybe just one gather :thinking:, check it out

Issue #762, opened 2026-02-05
Adapt to Apify's "multiple storages" functionality


janbuchar:
This is a feature that should complement the unnamed non-default storages:

1. Apify passes a JSON object with additional storage IDs to the Actor via environment - `ACTOR_STORAGE_IDS` - we should add this to the `Configuration` class. It is an object with three properties - `datasets`, `keyValueStores` and `requestQueues`. Each of these properties contains a `name -> ID` dictionary.
2. We should extend the `AliasResolver` class to utilize this so that `Dataset.open(alias="additional-dataset")` just works (™️). If an unknown alias is requested, we can either fall back to the current logic (create it, put an ID into the default KV-store) or shout at the user.

PR #797, opened 2026-02-17
feat: Support Actor schema storages with Alias mechanism

Pijukatel:
### Description

- Update `Configuration` to include `actor_storages` that is loaded from `actor_storages_json` env variable.
- Update `AliasResolver` to be able to resolve alias mapping from `Configuration`.

### Issues

- Closes: #762

### Testing

- Added E2E test
- Added unit tests
- Manual Actor test

### Checklist

- [x] CI passed

### DIFF
--- a/src/apify/_configuration.py
+++ b/src/apify/_configuration.py
@@ -35,15 +35,32 @@ def _transform_to_list(value: Any) -> list[str] | None:
 
 
 class ActorStorages(TypedDict):
+    """Mapping of storage aliases to their IDs, grouped by storage type.
+
+    Populated from the `ACTOR_STORAGES_JSON` env var that the Apify platform sets when an Actor declares
+    named storages in its `actor.json` schema. Each key maps a user-defined alias (e.g. `'custom'`)
+    to the platform-assigned storage ID.
+    """
+
     key_value_stores: dict[str, str]
     datasets: dict[str, str]
     request_queues: dict[str, str]
 
 
-def _load_storage_keys(
-    data: None | str | ActorStorages,
-) -> ActorStorages | None:
-    """Load storage keys."""
+def _load_storage_keys(data: None | str | ActorStorages) -> ActorStorages | None:
+    """Parse the `ACTOR_STORAGES_JSON` value into a normalized `ActorStorages` dict.
+
+    The platform provides this as a JSON string with camelCase keys (`keyValueStores`, `requestQueues`, `datasets`).
+    This validator deserializes the JSON when needed and normalizes the keys to snake_case, falling back to empty
+    dicts for missing storage types.
+
+    Args:
+        data: Raw value - `None` when the env var is unset, a JSON string from the env var, or an already-parsed
+        `ActorStorages` dict when set programmatically.
+
+    Returns:
+        Normalized storage mapping, or `None` if the input is `None`.
+    """
     if data is None:
         return None
     storage_mapping = json.loads(data) if isinstance(data, str) else data
@@ -470,7 +487,7 @@ class Configuration(CrawleeConfiguration):
         ActorStorages | None,
         Field(
             alias='actor_storages_json',
-            description='Storage IDs for the actor',
+            description='Mapping of storage aliases to their platform-assigned IDs.',
         ),
         BeforeValidator(_load_storage_keys),
     ] = None
--- a/tests/e2e/test_schema_storages/actor_source/main.py
+++ b/tests/e2e/test_schema_storages/actor_source/main.py
@@ -4,4 +4,6 @@ from apify import Actor
 async def main() -> None:
     async with Actor:
         assert Actor.configuration.actor_storages
-        assert (await Actor.open_dataset(alias='custom')).id == Actor.configuration.actor_storages['datasets']['custom']
+        dataset = await Actor.open_dataset(alias='custom')
+        expected_id = Actor.configuration.actor_storages['datasets']['custom']
+        assert dataset.id == expected_id


src/apify/storage_clients/_apify/_alias_resolving.py
```diff
@@ -262,3 +262,41 @@ async def _get_default_kvs_client(configuration: Configuration) -> KeyValueStore
             raise ValueError("'Configuration.default_key_value_store_id' must be set.")
 
         return apify_client_async.key_value_store(key_value_store_id=configuration.default_key_value_store_id)
+
+    @classmethod
+    async def register_aliases(cls, configuration: Configuration) -> None:
```
vdusek: 2. `register_aliases` lacks error handling

**`src/apify/storage_clients/_apify/_alias_resolving.py:265-302`** — The existing `store_mapping` method wraps KVS operations in `try/except` (line 236), but `register_aliases` does not. If `get_record` or `set_record` fails (e.g., network issue during Actor init), the entire Actor startup will crash.

```python
# register_aliases — no error handling:
existing_mapping = ((await client.get_record(...)) or {'value': {}}).get('value', {})
await client.set_record(cls._ALIAS_MAPPING_KEY, existing_mapping)
```

Consider wrapping in try/except with a warning, consistent with `store_mapping`.

src/apify/storage_clients/_apify/_alias_resolving.py
```diff
@@ -262,3 +262,41 @@ async def _get_default_kvs_client(configuration: Configuration) -> KeyValueStore
...
+        """Load alias mapping from configuration to the default kvs."""
```
vdusek: `register_aliases` doesn't acquire the alias lock

**`src/apify/storage_clients/_apify/_alias_resolving.py:268`** — The `register_aliases` classmethod reads and writes both the KVS record and `cls._alias_map` without acquiring `_alias_init_lock`. Meanwhile, `open_by_alias` (which calls `store_mapping`) **does** acquire the lock via the async context manager. If a user calls `open_dataset(alias='custom')` concurrently with Actor init, you could get a race condition on both the in-memory map and the KVS record.

src/apify/storage_clients/_apify/_alias_resolving.py
```diff
@@ -262,3 +262,41 @@ async def _get_default_kvs_client(configuration: Configuration) -> KeyValueStore
...
+                f'Conflicting default dataset ids: {configuration.default_dataset_id=},'
+                f" {configuration.actor_storages.datasets.get('default')=}"
+            )
+
+        for mapping, storage_type in (
+            (configuration.actor_storages.key_value_stores, 'KeyValueStore'),
+            (configuration.actor_storages.datasets, 'Dataset'),
+            (configuration.actor_storages.request_queues, 'RequestQueue'),
+        ):
+            for storage_alias, storage_id in mapping.items():
```
vdusek: Redundant `AliasResolver` instantiation in the loop

**`src/apify/storage_clients/_apify/_alias_resolving.py:285-293`** — A new `AliasResolver` instance is created for each `(storage_alias, storage_id)` pair just to access `_storage_key`. Each instantiation calls `hash_api_base_url_and_token(configuration)`, which is the same every time. Consider either:
- Extracting `_storage_key` as a `@staticmethod` / `@classmethod` that takes components directly, or
- Computing `_additional_cache_key` once outside the loop

```python
# Current — creates N instances just for the key:
for storage_alias, storage_id in mapping.items():
    configuration_mapping[
        cls(
            storage_type=storage_type,
            alias='__default__' if storage_alias == 'default' else storage_alias,
            configuration=configuration,
        )._storage_key
    ] = storage_id
```

src/apify/storage_clients/_apify/_alias_resolving.py
```diff
@@ -262,3 +262,41 @@ async def _get_default_kvs_client(configuration: Configuration) -> KeyValueStore
...
+                configuration_mapping[
+                    cls(  # noqa: SLF001# It is ok in own classmethod.
```
vdusek: Malformed noqa comment

```suggestion
                    cls(  # noqa: SLF001  # It is ok in own classmethod.
```

src/apify/storage_clients/_apify/_alias_resolving.py
```diff
@@ -262,3 +262,41 @@ async def _get_default_kvs_client(configuration: Configuration) -> KeyValueStore
...
+
+    @classmethod
+    async def register_aliases(cls, configuration: Configuration) -> None:
+        """Load alias mapping from configuration to the default kvs."""
+        if configuration.actor_storages is None:
+            return
+
+        configuration_mapping = {}
+
+        if configuration.default_dataset_id != configuration.actor_storages.datasets.get('default'):
```
vdusek: Conflict check is only for datasets

**`src/apify/storage_clients/_apify/_alias_resolving.py:274-278`** — There's a warning log for conflicting default dataset IDs, but no equivalent check for default KVS or default RQ. Either check all three storage types or drop the check entirely (since the platform should be authoritative).

```python
if configuration.default_dataset_id != configuration.actor_storages.datasets.get('default'):
    logger.warning(...)
# No similar check for KVS or RQ
```

tests/integration/test_storages.py
```diff
@@ -125,3 +128,53 @@ async def test_actor_implicit_storage_init(apify_token: str) -> None:
...
         assert await Actor.open_request_queue() is not await Actor.open_request_queue(force_cloud=True)
+
+
+async def test_actor_storages_alias_resolving(apify_token: str) -> None:
+    """Test that `AliasResolver.register_aliases` correctly updates default KVS with Actor storages."""
+
+    # Actor storages
+    datasets = {'default': 'default_dataset_id', 'custom': 'custom_dataset_id'}
+    request_queues = {'default': 'default_dataset_id', 'custom': 'custom_dataset_id'}
+    key_value_stores = {'default': 'default_dataset_id', 'custom': 'custom_dataset_id'}
```
vdusek:
```suggestion
    datasets = {'default': 'default_dataset_id', 'custom': 'custom_dataset_id'}
    request_queues = {'default': 'default_request_queue_id', 'custom': 'custom_request_queue_id'}
    key_value_stores = {'default': 'default_key_value_store_id', 'custom': 'custom_key_value_store_id'}
```

tests/integration/test_storages.py
```diff
@@ -125,3 +128,53 @@ async def test_actor_implicit_storage_init(apify_token: str) -> None:
...
+                    (storage_type, '__default__' if storage_alias == 'default' else storage_alias, client_cache_key)
+                )
+            ] = storage_id
+    expected_mapping.update(pre_existing_mapping)
+
+    try:
+        configuration.default_key_value_store_id = default_kvs.id
+        await AliasResolver.register_aliases(configuration=configuration)
+        assert await default_kvs.get_value(AliasResolver._ALIAS_MAPPING_KEY) == expected_mapping
+    finally:
```
vdusek: Integration test doesn't clean up `_alias_map`

**`tests/integration/test_storages.py:176-180`** — The `finally` block only drops the KVS but doesn't reset `AliasResolver._alias_map`. While the test isolation fixture does this between tests, it's good practice to clean up what you dirty — especially since `_alias_map` is a class variable that persists across the process.

tests/e2e/test_schema_storages/actor_source/actor.json
```diff
@@ -0,0 +1,24 @@
+{
```
vdusek, line 1: spaces please

src/apify/_configuration.py
```diff
@@ -446,6 +470,15 @@ class Configuration(CrawleeConfiguration):
...
     ] = None
 
+    actor_storages: Annotated[
+        ActorStorages | None,
+        Field(
+            alias='actor_storages_json',
+            description='Storage IDs for the actor',
+        ),
+        BeforeValidator(_load_storage_keys),
+    ] = None
```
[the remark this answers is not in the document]
Pijukatel: The implementation on the platform is `ACTOR_STORAGES_JSON` now, if it changes the code should reflect it as well

tests/integration/test_storages.py
```diff
@@ -125,3 +128,53 @@ async def test_actor_implicit_storage_init(apify_token: str) -> None:
...
+        token=apify_token,
+        actor_storages=ActorStorages(
+            datasets=datasets, request_queues=request_queues, key_value_stores=key_value_stores
+        ),
+    )
+    storage_client = ApifyStorageClient()
+    service_locator.set_configuration(configuration)
+    service_locator.set_storage_client(storage_client)
+
+    client_cache_key = cast('tuple', storage_client.get_storage_client_cache_key(configuration))[-1]
```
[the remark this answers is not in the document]
Pijukatel: no, TC006 [*] Add quotes to type expression in `typing.cast()`

src/apify/storage_clients/_apify/_alias_resolving.py
Pijukatel: Will skip for mapping that contains only default storages, but it should crash for actor that contains the mapping in such case. Otherwise, someone could try to access such storage by alias and it would create new storage instead of reusing the pre-existing one.

src/apify/storage_clients/_apify/_alias_resolving.py
Pijukatel: Good point, updated

src/apify/storage_clients/_apify/_alias_resolving.py
Pijukatel: Removed. This should be handled by platform tests

src/apify/storage_clients/_apify/_alias_resolving.py
Pijukatel: Refactored

tests/e2e/test_schema_storages/actor_source/actor.json
Pijukatel: reformated

tests/integration/test_storages.py
Pijukatel: It cleans the "unique dirty stuff", but keeps the generic for the auto-use fixture.

There is no benefit in repeating the cleanup/setup done by autosue fixtures in individual tests.

src/apify/storage_clients/_apify/_alias_resolving.py
Pijukatel: Reworked to avoid any API calls at all. It is completely based on the `Configuration`

tests/e2e/test_schema_storages/actor_source/main.py
```diff
@@ -0,0 +1,7 @@
+from apify import Actor
+
+
+async def main() -> None:
+    async with Actor:
+        assert Actor.configuration.actor_storages
+        assert (await Actor.open_dataset(alias='custom')).id == Actor.configuration.actor_storages.datasets['custom']
```
valekjo: Suggestion: Would it also make sense to check that `default` is `default`?

src/apify/storage_clients/_apify/_alias_resolving.py
```diff
@@ -201,6 +194,18 @@ async def resolve_id(self) -> str | None:
         Returns:
             Storage id if it exists, None otherwise.
         """
+        # First try to find the alias in the configuration mapping to avoid any API calls.
+        # This mapping is maintained by the Apify platform and does not have to be maintained in the default KVS.
+        if self._configuration.actor_storages and self._alias != 'default':
```
valekjo, line 198: Question: I don't understand why we're treating 'default' differently.

Pijukatel: The way to access the default storage is without an alias. Loading up default aliases from the configuration mapping would allow an alternative way to access the default storage through "default" alias. In Python that is generally frowned upon:
`There should be one-- and preferably only one --obvious way to do it.`

tests/e2e/test_schema_storages/actor_source/main.py
Pijukatel: That should probably be tested on the platform level. If the env vars are inconsistent. (default id is different in mapping env and in dedicated env var, then our code can't do much about that)

src/apify/storage_clients/_apify/_alias_resolving.py
valekjo: Ah, sometimes I wish Javascript was like this :D

src/apify/storage_clients/_apify/_alias_resolving.py
```diff
@@ -5,6 +5,8 @@
 from logging import getLogger
 from typing import TYPE_CHECKING, ClassVar, Literal, overload
 
+from propcache import cached_property
```
vdusek: What kind of `cached_property` is this? Shouldn't we import it from functools? Otherwise, we'd need to add another production dependency, and I don't think we want to do that just for this decorator 🙂.

src/apify/_configuration.py
```diff
@@ -34,6 +35,29 @@ def _transform_to_list(value: Any) -> list[str] | None:
     return value if isinstance(value, list) else str(value).split(',')
 
 
+@dataclasses.dataclass
+class ActorStorages:
```
vdusek: I think using a `dataclass` here is not the right choice, as it would allow parsing objects like `{"datasets": "not-a-dict"}`, if I'm not mistaken. It should be a Pydantic model, since we consume data from the outside.

But mostly, I'm not convinced that introducing a new structure for this is a good idea, especially since it's just a typed wrapper. It would have to become part of the public API for cases where someone instantiates `Configuration` directly in code, and it would make the DX worse.

src/apify/storage_clients/_apify/_alias_resolving.py
```diff
@@ -220,30 +225,21 @@ async def store_mapping(self, storage_id: str) -> None:
...
             # Store the mapping back in the KVS.
-            await default_kvs_client.set_record(self._ALIAS_MAPPING_KEY, record)
+            await default_kvs_client.set_record(key=self._ALIAS_MAPPING_KEY, value=value)
         except Exception as exc:
             logger.warning(f'Error storing alias mapping for {self._alias}: {exc}')
 
-    @property
+    @cached_property
     def _storage_key(self) -> str:
-        """Get a unique storage key used for storing the alias in the mapping."""
```
vdusek, line 241: Why are you removing the docstring?

tests/unit/actor/test_configuration.py
```diff
@@ -300,3 +300,26 @@ def test_actor_pricing_info_from_json_env_var(monkeypatch: pytest.MonkeyPatch) -
...
     assert config.actor_pricing_info.pricing_model == 'PAY_PER_EVENT'
+
+
+def test_actor_storage_json_env_var(monkeypatch: pytest.MonkeyPatch) -> None:
+    """Test that actor_storages_json is parsed from JSON env var."""
+    import json
+
+    datasets = {'default': 'default_dataset_id', 'custom': 'custom_dataset_id'}
+    request_queues = {'default': 'default_dataset_id', 'custom': 'custom_dataset_id'}
+    key_value_stores = {'default': 'default_dataset_id', 'custom': 'custom_dataset_id'}
```
vdusek: I already wrote that all three storage types use `'default_dataset_id'` / `'custom_dataset_id'` as values.

Pijukatel: Fixed

src/apify/_configuration.py
Pijukatel: Changed to Pydantic Model

src/apify/storage_clients/_apify/_alias_resolving.py
Pijukatel: Back

src/apify/_configuration.py
vdusek: I'm sorry, but the point of the comment was:

> I'm not convinced that introducing a new structure for this is a good idea, especially since it's just a typed wrapper. It would have to become part of the public API for cases where someone instantiates Configuration directly in code, and it would make the DX worse.

Pijukatel: Ok it is a `TypedDict` now

src/apify/_configuration.py
```diff
@@ -446,6 +466,15 @@ class Configuration(CrawleeConfiguration):
         BeforeValidator(lambda data: json.loads(data) if isinstance(data, str) else data or None),
     ] = None
 
+    actor_storages: Annotated[
```
janbuchar, line 486: I'm not sure about the naming here - we usually strip the `actor_` prefix. But I do agree that a plain `storages` would look odd. So let's keep it this way, I guess.

src/apify/_configuration.py
```diff
@@ -34,6 +34,26 @@ def _transform_to_list(value: Any) -> list[str] | None:
     return value if isinstance(value, list) else str(value).split(',')
 
 
+class ActorStorages(TypedDict):
+    key_value_stores: dict[str, str]
+    datasets: dict[str, str]
+    request_queues: dict[str, str]
```
vdusek, line 47: Can we add a docstring? For example something like this:

```suggestion
class ActorStorages(TypedDict):
    """Mapping of storage aliases to their IDs, grouped by storage type.

    Populated from the `ACTOR_STORAGES_JSON` env var that the Apify platform sets when an Actor declares
    named storages in its `actor.json` schema. Each key maps a user-defined alias (e.g. `'custom'`)
    to the platform-assigned storage ID.
    """

    key_value_stores: dict[str, str]
    datasets: dict[str, str]
    request_queues: dict[str, str]
```

src/apify/_configuration.py
```diff
@@ -34,6 +34,26 @@ def _transform_to_list(value: Any) -> list[str] | None:
...
+
+
+def _load_storage_keys(
+    data: None | str | ActorStorages,
+) -> ActorStorages | None:
+    """Load storage keys."""
```
vdusek: Could we use a more descriptive docstring, please? The current docstring, "Load storage keys.", for the `_load_storage_keys` method doesn't really provide any valuable information.

```suggestion
def _load_storage_keys(data: None | str | ActorStorages) -> ActorStorages | None:
    """Parse the `ACTOR_STORAGES_JSON` value into a normalized `ActorStorages` dict.

    The platform provides this as a JSON string with camelCase keys (`keyValueStores`, `requestQueues`, `datasets`).
    This validator deserializes the JSON when needed and normalizes the keys to snake_case, falling back to empty
    dicts for missing storage types.

    Args:
        data: Raw value - `None` when the env var is unset, a JSON string from the env var, or an already-parsed
        `ActorStorages` dict when set programmatically.

    Returns:
        Normalized storage mapping, or `None` if the input is `None`.
    """
```

src/apify/_configuration.py
```diff
@@ -446,6 +466,15 @@ class Configuration(CrawleeConfiguration):
         BeforeValidator(lambda data: json.loads(data) if isinstance(data, str) else data or None),
     ] = None
 
+    actor_storages: Annotated[
+        ActorStorages | None,
+        Field(
+            alias='actor_storages_json',
+            description='Storage IDs for the actor',
```
vdusek:
```suggestion
            description='Mapping of storage aliases to their platform-assigned IDs.',
```

tests/e2e/test_schema_storages/actor_source/main.py
```diff
@@ -0,0 +1,7 @@
+from apify import Actor
+
+
+async def main() -> None:
+    async with Actor:
+        assert Actor.configuration.actor_storages
+        assert (await Actor.open_dataset(alias='custom')).id == Actor.configuration.actor_storages['datasets']['custom']
```
vdusek: It would be better to split this into multiple lines. That way, if an error occurs, it will be easier to identify where the problem is, e.g.:


```suggestion
        dataset = await Actor.open_dataset(alias='custom')
        expected_id = Actor.configuration.actor_storages['datasets']['custom']
        assert dataset.id == expected_id
```

tests/unit/storage_clients/test_alias_resolver.py
```diff
@@ -76,3 +76,24 @@ async def test_get_alias_map_returns_in_memory_map() -> None:
...
     assert result == {}
+
+
+async def test_configuration_storages_alias_resolving() -> None:
+    """Test that `AliasResolver` correctly resolves ids of storages registered in Configuration."""
+
+    # Actor storages
+    datasets = {'default': 'default_dataset_id', 'custom': 'custom_Dataset_id'}
+    request_queues = {'default': 'default_request_queue_id', 'custom': 'custom_RequestQueue_id'}
+    key_value_stores = {'default': 'default_key_value_store_id', 'custom': 'custom_KeyValueStore_id'}
```
vdusek, line 87: I assume the mixing of camel and snake cases is not intended

```suggestion
    datasets = {'default': 'default_dataset_id', 'custom': 'custom_dataset_id'}
    request_queues = {'default': 'default_request_queue_id', 'custom': 'custom_request_queue_id'}
    key_value_stores = {'default': 'default_key_value_store_id', 'custom': 'custom_key_value_store_id'}
```

src/apify/_configuration.py
vdusek: looks good

tests/unit/storage_clients/test_alias_resolver.py
Pijukatel: That value can be any string, so I just used one that made it more convenient to write the test. 

Look at the assert loop and the fstring:
f'custom_{storage_type}_id'

tests/e2e/test_schema_storages/actor_source/main.py
```diff
@@ -0,0 +1,9 @@
+from apify import Actor
```
vdusek, line 1: One more thing, this won't cost us anything and could be more helpful for testing.

Could you please add a custom KVS and RQ to `actor.json` as well? Then, in this test, open them and verify that the expected IDs are returned.

Pijukatel: I tried, but the platform does not support them in the schema yet. Currently, it is not planned.

vdusek: Hmm :thinking:, alright then

PR #803, opened 2026-02-18
test: Fix flaky integration tests for RQ in shared mode

vdusek:
## Summary
- Add retry logic to `fetch_next_request()` calls in `test_request_reclaim_functionality` to handle propagation delays in shared request queue access mode
- The `[shared]` variant was flaky because newly added or reclaimed requests may not be immediately visible
- Introduces a `_fetch_next_request_with_retry()` helper that polls up to 10 times with 1-second intervals

## Test plan
- [ ] CI integration tests pass consistently for both `[single]` and `[shared]` variants

### DIFF
--- a/tests/integration/test_request_queue.py
+++ b/tests/integration/test_request_queue.py
@@ -26,6 +26,23 @@ if TYPE_CHECKING:
     from apify.storage_clients._apify._models import ApifyRequestQueueMetadata
 
 
+async def fetch_next_request_with_exp_backoff(rq: RequestQueue, max_retries: int = 5) -> Request | None:
+    """Fetch the next request with exponential backoff retries.
+
+    In shared request queue mode, there is a propagation delay before newly added or reclaimed requests become visible
+    (see https://github.com/apify/apify-sdk-python/issues/808). This helper retries with exponential backoff to handle
+    that delay in integration tests.
+    """
+    for attempt in range(max_retries):
+        result = await rq.fetch_next_request()
+        if result is not None:
+            return result
+        delay = 2**attempt
+        Actor.log.info(f'fetch_next_request returned None, retrying in {delay}s (attempt {attempt + 1}/{max_retries})')
+        await asyncio.sleep(delay)
+    return None
+
+
 async def test_add_and_fetch_requests(request_queue_apify: RequestQueue) -> None:
     """Test basic functionality of adding and fetching requests."""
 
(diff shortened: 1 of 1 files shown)


tests/integration/test_request_queue.py
```diff
@@ -218,8 +228,8 @@ async def test_request_reclaim_functionality(request_queue_apify: RequestQueue)
     await rq.add_request('https://example.com/test')
     Actor.log.info('Added test request')
 
-    # Fetch and reclaim the request
-    request = await rq.fetch_next_request()
+    # Fetch and reclaim the request (retry to handle propagation delay in shared mode)
+    request = await _fetch_next_request_with_retry(rq)
```
janbuchar: Since the integration test shows intended usage, I think we're not actually fixing anything. We're just sweeping an issue under the rug. 

In other words, this is not a flaky test, this shows the code under test being unreliable.

vdusek: Yes, but how can we handle this in the case of a shared client? 🙂 If I understand it correctly, a single client adds requests directly to the local `queue_head` deque, which makes them immediately available. However, a shared client can't do that, right? If retries aren't an option, I'd suggest skipping this test for the shared client for now, opening an issue, and keeping it skipped until we resolve it somehow.

janbuchar: I guess either skip it or add a well-aimed `if mode == "shared": sleep()`. That's probably more upfront about what we're dealing with.

vdusek: Alright, see updates and https://github.com/apify/apify-sdk-python/issues/808.

(You must approve, because you requested changes before)

tests/integration/test_request_queue.py
```diff
@@ -260,20 +289,24 @@ async def test_request_reclaim_with_forefront(request_queue_apify: RequestQueue)
...
-    assert next_request is not None, f'next_request={next_request}'
-    assert next_request.url == first_request.url, (
-        f'next_request.url={next_request.url}',
-        f'first_request.url={first_request.url}',
-    )
+    # The reclaimed request should be fetched first again.
+    # In shared mode, there is a propagation delay before the reclaimed request becomes visible
+    # (see https://github.com/apify/apify-sdk-python/issues/808).
+    if rq_access_mode == 'shared':
+        next_request = await fetch_next_request_with_exp_backoff(rq)
```
janbuchar, line 303: This looks like a massive overkill to me, can't we just give it 5 seconds and call it a day?

janbuchar (approved):

tests/integration/test_request_queue.py
vdusek: Ahh, I don't wanna spend any more time on this. The goal is to keep it from failing until we resolve #808. Once that's done, this will be removed anyway, so I'd leave it as is for now.

PR #813, opened 2026-02-25
test: Fix flaky `is_finished` checks in shared RQ integration tests

vdusek:
## Summary

- Introduced a generic `call_with_exp_backoff` helper in `tests/integration/_utils.py` that retries any async callable with exponential backoff until it returns a truthy value, replacing the previous `fetch_next_request_with_exp_backoff`
- Applied the helper to all `is_finished()` checks in shared request queue mode across 9 integration tests to handle API propagation delays
- Renamed local variables that shadowed `request: pytest.FixtureRequest` to fix type checker errors

## Test plan

- [x] Integration tests pass in CI (both `single` and `shared` RQ modes)
- [x] No more flaky `is_finished` failures in shared mode

## Motivation

One of the failed run: https://github.com/apify/apify-sdk-python/actions/runs/22375456542/job/64764491226?pr=812

### DIFF
--- a/tests/integration/_utils.py
+++ b/tests/integration/_utils.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import asyncio
-from typing import TYPE_CHECKING, TypeVar, cast
+from typing import TYPE_CHECKING, TypeVar
 
 from crawlee._utils.crypto import crypto_random_object_id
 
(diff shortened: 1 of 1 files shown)


tests/integration/_utils.py
```diff
@@ -1,7 +1,35 @@
...
 
+from apify import Actor
+
+if TYPE_CHECKING:
+    from collections.abc import Awaitable, Callable
+
+T = TypeVar('T')
+
+
+async def call_with_exp_backoff(fn: Callable[[], Awaitable[T]], *, max_retries: int = 3) -> T:
```
Pijukatel: Shouldn't this be `-> T | None` 

and drop the cast? It should still pass the type check.

Issue #760, opened 2026-02-03
Keep track of synthetic apify-default-dataset-item events


janbuchar:
- Port over https://github.com/apify/apify-sdk-js/pull/528
    - We should **wait until the JS changes are released and tested** to prevent duplicating efforts
    - The monkey patching part may not be necessary - the `ApifyStorageClient` Python class might be a good place for the synthetic event accounting logic
- Related - #705

Issue #705, opened 2025-12-08
calculate_max_event_charge_count_within_limit doesn't do it's job when pushing data to default dataset in PPE actor


devsaswave:
Currently when testing a user providing Maximum cost per run to an PPE actor run (not input but env variable seen from actor side)

EVENT_NAME = "apify-default-dataset-item"

```
async with Actor:
  charging_manager = Actor.get_charging_manager()
  remaining = charging_manager.calculate_max_event_charge_count_within_limit(EVENT_NAME)
  await Actor.push_data({"test":"Yes"})
  remaining = charging_manager.calculate_max_event_charge_count_within_limit(EVENT_NAME)
  await Actor.push_data({"test":"Yes"})
  remaining = charging_manager.calculate_max_event_charge_count_within_limit(EVENT_NAME)
```

remaining is always the same 

Unless if i do 

`await Actor.push_data({"test":"Yes"}, EVENT_NAME)`

Then calculate_max_event_charge_count_within_limit() works corretly
But i have to catch and ignore the error raised by push_data() because as mentioned in the logs 

`apify_client._errors.ApifyApiError: Event "apify-default-dataset-item" is system event and cannot be charged.`

PR #814, opened 2026-02-26
fix: Keep track of synthetic apify-default-dataset-item events

Mantisus:
### Description

- Add `_DatasetClientPPEMixin` to dataset storage clients to automatically charge for the synthetic `apify-default-dataset-item` event on every push to the default dataset.
- Update `Actor.push_data()` to account for the combined explicit + synthetic event price when enforcing budget limits.

### Issues

- Closes: #760
- Closes: #705

### Testing

- Add new unit and E2E test.

### DIFF
--- a/src/apify/_actor.py
+++ b/src/apify/_actor.py
@@ -26,7 +26,7 @@ from crawlee.events import (
     EventSystemInfoData,
 )
 
-from apify._charging import ChargeResult, ChargingManager, ChargingManagerImplementation
+from apify._charging import DEFAULT_DATASET_ITEM_EVENT, ChargeResult, ChargingManager, ChargingManagerImplementation
 from apify._configuration import Configuration
 from apify._consts import EVENT_LISTENERS_TIMEOUT
 from apify._crypto import decrypt_input_secrets, load_private_key
@@ -613,12 +613,8 @@ class _ActorType:
             storage_client=self._storage_client.get_suitable_storage_client(force_cloud=force_cloud),
         )
 
-    @overload
-    async def push_data(self, data: dict | list[dict]) -> None: ...
-    @overload
-    async def push_data(self, data: dict | list[dict], charged_event_name: str) -> ChargeResult: ...
     @_ensure_context
-    async def push_data(self, data: dict | list[dict], charged_event_name: str | None = None) -> ChargeResult | None:
+    async def push_data(self, data: dict | list[dict], charged_event_name: str | None = None) -> ChargeResult:
         """Store an object or a list of objects to the default dataset of the current Actor run.
 
         Args:
@@ -626,32 +622,46 @@ class _ActorType:
             charged_event_name: If provided and if the Actor uses the pay-per-event pricing model,
                 the method will attempt to charge for the event for each pushed item.
         """
-        if not data:
-            return None
-
-        data = data if isinstance(data, list) else [data]
-
         if charged_event_name and charged_event_name.startswith('apify-'):
             raise ValueError(f'Cannot charge for synthetic event "{charged_event_name}" manually')
 
         charging_manager = self.get_charging_manager()
 
+        if not data:
+            charged_event_name = charged_event_name or DEFAULT_DATASET_ITEM_EVENT
+            charge_limit_reached = charging_manager.is_event_charge_limit_reached(charged_event_name)
+
+            return ChargeResult(
+                event_charge_limit_reached=charge_limit_reached,
+                charged_count=0,
+                chargeable_within_limit=charging_manager.compute_chargeable(),
+            )
+
+        data = data if isinstance(data, list) else [data]
+
+        dataset = await self.open_dataset()
+
         # Acquire the charge lock to prevent race conditions between concurrent
         # push_data calls. We need to hold the lock for the entire push_data + charge sequence.
-        async with charging_manager.charge_lock:
-            # No explicit charging requested; synthetic events are handled within dataset.push_data.
+        async with charging_manager.charge_lock():
+            # Synthetic events are handled within dataset.push_data, only get data for `ChargeResult`.
             if charged_event_name is None:
-                dataset = await self.open_dataset()
+                before = charging_manager.get_charged_event_count(DEFAULT_DATASET_ITEM_EVENT)
                 await dataset.push_data(data)
-                return None
+                after = charging_manager.get_charged_event_count(DEFAULT_DATASET_ITEM_EVENT)
+                return ChargeResult(
+                    event_charge_limit_reached=charging_manager.is_event_charge_limit_reached(
+                        DEFAULT_DATASET_ITEM_EVENT
+                    ),
+                    charged_count=after - before,
+                    chargeable_within_limit=charging_manager.compute_chargeable(),
+                )
 
-            pushed_items_count = self.get_charging_manager().compute_push_data_limit(
+            pushed_items_count = charging_manager.compute_push_data_limit(
                 items_count=len(data),
                 event_name=charged_event_name,
                 is_default_dataset=True,
             )
-
-            dataset = await self.open_dataset()
 
             if pushed_items_count < len(data):
                 await dataset.push_data(data[:pushed_items_count])
@@ -659,7 +669,7 @@ class _ActorType:
                 await dataset.push_data(data)
 
             # Only charge explicit events; synthetic events will be processed within the client.
-            return await self.get_charging_manager().charge(
+            return await charging_manager.charge(
                 event_name=charged_event_name,
                 count=pushed_items_count,
             )
@@ -723,10 +733,9 @@ class _ActorType:
             event_name: Name of the event to be charged for.
             count: Number of events to charge for.
         """
-        # Acquire lock to prevent race conditions with concurrent charge/push_data calls.
+        # charging_manager.charge() acquires charge_lock internally.
         charging_manager = self.get_charging_manager()
-        async with charging_manager.charge_lock:
-            return await charging_manager.charge(event_name, count)
+        return await charging_manager.charge(event_name, count)
 
     @overload
     def on(
--- a/src/apify/_charging.py
+++ b/src/apify/_charging.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import asyncio
 import math
 from contextvars import ContextVar
 from dataclasses import dataclass
@@ -18,7 +17,7 @@ from apify._models import (
     PricePerDatasetItemActorPricingInfo,
     PricingModel,
 )
-from apify._utils import docs_group, ensure_context
+from apify._utils import ReentrantLock, docs_group, ensure_context
 from apify.log import logger
 from apify.storages import Dataset
 
@@ -53,7 +52,7 @@ class ChargingManager(Protocol):
     - Apify platform documentation: https://docs.apify.com/platform/actors/publishing/monetize
     """
 
-    charge_lock: asyncio.Lock
+    charge_lock: ReentrantLock
     """Lock to synchronize charge operations. Prevents race conditions between `charge` and `push_data` calls."""
 
     async def charge(self, event_name: str, count: int = 1) -> ChargeResult:
@@ -114,6 +113,12 @@ class ChargingManager(Protocol):
             Max number of items that can be pushed within the budget.
         """
 
+    def is_event_charge_limit_reached(self, event_name: str) -> bool:
+        """Return True if the remaining budget is insufficient to charge even a single event of the given type."""
+
+    def compute_chargeable(self) -> dict[str, int | None]:
+        """Compute the maximum number of events of each type that can be charged within the current budget."""
+
 
 @docs_group('Charging')
 @dataclass(frozen=True)
@@ -170,7 +175,7 @@ class ChargingManagerImplementation(ChargingManager):
         self._not_ppe_warning_printed = False
         self.active = False
 
-        self.charge_lock = asyncio.Lock()
+        self.charge_lock = ReentrantLock()
 
     async def __aenter__(self) -> None:
         """Initialize the charging manager - this is called by the `Actor` class and shouldn't be invoked manually."""
@@ -244,13 +249,6 @@ class ChargingManagerImplementation(ChargingManager):
 
     @_ensure_context
     async def charge(self, event_name: str, count: int = 1) -> ChargeResult:
-        def calculate_chargeable() -> dict[str, int | None]:
-            """Calculate the maximum number of events of each type that can be charged within the current budget."""
-            return {
-                event_name: self.calculate_max_event_charge_count_within_limit(event_name)
-                for event_name in self._pricing_info
-            }
-
         # For runs that do not use the pay-per-event pricing model, just print a warning and return
         if self._pricing_model != 'PAY_PER_EVENT':
             if not self._not_ppe_warning_printed:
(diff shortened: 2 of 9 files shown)


src/apify/storage_clients/_apify/_dataset_client.py
```diff
@@ -13,6 +13,7 @@
 from crawlee.storage_clients.models import DatasetItemsListPage, DatasetMetadata
 
 from ._api_client_creation import create_storage_api_client
+from apify.storage_clients._ppe_dataset_mixin import _DatasetClientPPEMixin
```
vdusek: just regarding the naming:

- The class itself should not have the `_` prefix, since we import it from another module.
- We also use this convention: `DatasetClientPpeMixin` - short names should be capitalized as well (e.g.  `HttpCrawler`).

Mantisus: Updated

src/apify/storage_clients/_ppe_dataset_mixin.py
```diff
@@ -0,0 +1,31 @@
...
+    def is_default_dataset(self, value: bool) -> None:
+        self._is_default_dataset = value
+
+    async def _calculate_limit_for_push(self, items_count: int) -> int:
+        if self.is_default_dataset and (charging_manager := charging_manager_ctx.get()):
+            max_charged_count = charging_manager.calculate_max_event_charge_count_within_limit(
+                event_name=DEFAULT_DATASET_ITEM_EVENT
+            )
+            return min(max_charged_count, items_count) if max_charged_count is not None else items_count
+        return items_count
```
vdusek: This can be just sync method

src/apify/storage_clients/_file_system/_dataset_client.py
```diff
@@ -0,0 +1,49 @@
...
+from crawlee.storage_clients._file_system import FileSystemDatasetClient
+
+from apify._configuration import Configuration as ApifyConfiguration
+from apify.storage_clients._ppe_dataset_mixin import DatasetClientPpeMixin
+
+if TYPE_CHECKING:
+    from crawlee.configuration import Configuration
+
+
+class ApifyFileSystemDatasetClient(FileSystemDatasetClient, DatasetClientPpeMixin):
```
vdusek, line 15: Could you please add a docstring? Explaining why this is necessary, how it differs from the Crawlee's FS client, ...

src/apify/storage_clients/_apify/_dataset_client.py
```diff
@@ -48,6 +49,8 @@ def __init__(
 
         Preferably use the `ApifyDatasetClient.open` class method to create a new instance.
         """
+        super().__init__()
```
vdusek: How does this work in the case of multiple inheritance? Shouldn't we use rather:

```python
DatasetClient.__init__(self)
DatasetClientPpeMixin.__init__(self)
```

?

src/apify/storage_clients/_file_system/_dataset_client.py
```diff
@@ -0,0 +1,49 @@
...
+            configuration=configuration,
+        )
+
+        if isinstance(configuration, ApifyConfiguration) and all(v is None for v in (id, name, alias)):
+            dataset_client.is_default_dataset = True
+
+        return dataset_client
+
+    @override
+    async def push_data(self, data: list[dict[str, Any]] | dict[str, Any]) -> None:
```
vdusek, line 50: Unlike in the Apify storage client, there is no lock around budget check + push + charge, can't this cause troubles?

src/apify/storage_clients/_ppe_dataset_mixin.py
```diff
@@ -0,0 +1,31 @@
+from apify._charging import DEFAULT_DATASET_ITEM_EVENT, charging_manager_ctx
+
+
+class DatasetClientPpeMixin:
+    """A mixin for dataset clients to add support for PPE pricing model and tracking synthetic events."""
+
+    def __init__(self) -> None:
+        self._is_default_dataset: bool = False
```
vdusek: AFAIK redundant explicit type

```suggestion
        self._is_default_dataset = False
```

src/apify/storage_clients/_ppe_dataset_mixin.py
```diff
@@ -0,0 +1,31 @@
...
+
+    @property
+    def is_default_dataset(self) -> bool:
+        return self._is_default_dataset
+
+    @is_default_dataset.setter
+    def is_default_dataset(self, value: bool) -> None:
+        self._is_default_dataset = value
```
vdusek: Is there any reason why not just make `_is_default_dataset` a public property instead? 🙂

src/apify/storage_clients/_apify/_dataset_client.py
```diff
@@ -108,12 +111,16 @@ async def open(
...
         )
 
-        return cls(
+        dataset_client = cls(
             api_client=api_client,
             api_public_base_url='',  # Remove in version 4.0, https://github.com/apify/apify-sdk-python/issues/635
             lock=asyncio.Lock(),
         )
 
+        dataset_client.is_default_dataset = (await dataset_client.get_metadata()).id == configuration.default_dataset_id
```
vdusek: In the FS dataset client, we have this:

```python
if isinstance(configuration, ApifyConfiguration) and all(v is None for v in (id, name, alias)):
    dataset_client.is_default_dataset = True
```

Isn't this an inconsistency of default storage detection between clients?

src/apify/storage_clients/_file_system/_storage_client.py
```diff
@@ -48,3 +49,20 @@ async def create_kvs_client(
         )
         await self._purge_if_needed(client, configuration)
         return client
+
+    @override
+    async def create_dataset_client(
```
vdusek, line 54: We do not call `_purge_if_needed` here

src/apify/storage_clients/_apify/_dataset_client.py
```diff
@@ -128,21 +135,19 @@ async def drop(self) -> None:
...
+            items = items[:limit]
 
-            # Handle singular items
-            else:
-                items = await self._check_and_serialize(data)
-                await self._api_client.push_items(items=items)
+            async for chunk in self._chunk_by_size(payloads_generator(items)):
+                await self._api_client.push_items(items=chunk)
+
+            await self._charge_for_items(count_items=limit)
```
vdusek, line 153: Here, we charge `limit`, but in `Actor` we do:

```python
charge(event_name=charged_event_name, count=pushed_items_count)
```

Is that intentional? (`limit` vs `pushed_items_count`)

src/apify/_actor.py
```diff
@@ -637,6 +628,9 @@ async def push_data(self, data: dict | list[dict], charged_event_name: str | Non
 
         data = data if isinstance(data, list) else [data]
 
+        if charged_event_name and charged_event_name.startswith('apify-'):
+            raise ValueError(f'Cannot charge for synthetic event "{charged_event_name}" manually')
+
```
vdusek, line 627: Potential race condition from Claude, please think it through.

### Race condition: non-charging path doesn't hold `_charge_lock`

`src/apify/_actor.py` — When `push_data` is called **without** a `charged_event_name`, it takes the "no charging" path and calls `dataset.push_data(data)` directly without acquiring `_charge_lock`. However, the dataset client's mixin will still check the budget and charge for synthetic `apify-default-dataset-item` events. Between `_calculate_limit_for_push` (budget read) and `_charge_for_items` (budget write), there are `await` points (the actual API push), during which a concurrent coroutine holding `_charge_lock` could mutate the charging state. This is a TOCTOU race that could cause the budget to be exceeded.

src/apify/storage_clients/_apify/_dataset_client.py
Mantisus: It works because it inherits from an abstract class and a mixin. But I agree that explicit initialization is better, and it will be consistent with `ApifyFileSystemDatasetClient`.

src/apify/storage_clients/_ppe_dataset_mixin.py
Mantisus: Fair enough. 🙂

src/apify/storage_clients/_apify/_dataset_client.py
Mantisus: In the `Actor`, we call `charge` only for the event received from the user. Inside the client, only for the synthetic events `apify-default-dataset-item`.

src/apify/_actor.py
Mantisus: This is possible if the user mixes calls to `Actor.push_data` with and without `charged_event_name` when loading data into the default dataset. I assume that this is possible if the user does not charge the full price for incomplete data.

Updated so that both flows are executed inside `_charge_lock`

src/apify/storage_clients/_file_system/_storage_client.py
Mantisus: Fixed

src/apify/storage_clients/_apify/_dataset_client.py
Mantisus: Updated for a more consistent approach.

src/apify/storage_clients/_file_system/_dataset_client.py
Mantisus: Good catch. Updated the method so that the `check + push + charge` is inside the lock.

src/apify/storage_clients/_ppe_dataset_mixin.py
Mantisus: Updated

src/apify/storage_clients/_file_system/_dataset_client.py
Mantisus: Added

src/apify/_charging.py
```diff
@@ -81,6 +90,28 @@ def get_charged_event_count(self, event_name: str) -> int:
     def get_max_total_charge_usd(self) -> Decimal:
         """Get the configured maximum total charge for this Actor run."""
 
+    def calculate_push_data_limit(
```
vdusek: please use "compute" instead of "calculate" (I know calculate was used before as well, but let's fix it)

src/apify/_charging.py
```diff
@@ -337,6 +371,25 @@ def get_charged_event_count(self, event_name: str) -> int:
     def get_max_total_charge_usd(self) -> Decimal:
         return self._max_total_charge_usd
 
+    @ensure_context
+    def calculate_push_data_limit(
```
vdusek: please use "compute" instead of "calculate" (I know calculate was used before as well, but let's fix it)

src/apify/storage_clients/_file_system/_dataset_client.py
```diff
@@ -0,0 +1,67 @@
...
+        async with self._lock:
+            items = data if isinstance(data, list) else [data]
+            limit = self._calculate_limit_for_push(len(items))
+
+            new_item_count = self._metadata.item_count
+            for item in items:
```
vdusek: Shouldn't we iterate only over `items[:limit]`?

src/apify/_charging.py
```diff
@@ -31,6 +32,14 @@
 
 run_validator = TypeAdapter[ActorRun | None](ActorRun | None)
 
+DEFAULT_DATASET_ITEM_EVENT = 'apify-default-dataset-item'
+
+# Context variable to hold the current `ChargingManager` instance, if any. This allows PPE-aware dataset clients to
+# access the charging manager without needing to pass it explicitly.
+charging_manager_ctx: ContextVar[ChargingManagerImplementation | None] = ContextVar(
```
vdusek: Shouldn't we use the protocol as a type hint rather than its implementation?

```suggestion
charging_manager_ctx: ContextVar[ChargingManager | None] = ContextVar(
```

tests/unit/actor/test_charging_manager.py
```diff
@@ -247,6 +247,78 @@ async def test_get_max_total_charge_usd(mock_client: MagicMock) -> None:
...
+        max_total_charge_usd=Decimal('3.00'),
+    )
+    cm = ChargingManagerImplementation(config, mock_client)
+    async with cm:
+        # explicit price only = 1.00/item, budget = 3.00, max = floor(3/1) = 3
+        result = cm.calculate_push_data_limit(5, 'scrape', is_default_dataset=False)
+        assert result == 3
+
+
+async def test_calculate_push_data_limit_exhausted_budget(mock_client: MagicMock) -> None:
```
vdusek: please use "compute" instead of "calculate" (I know calculate was used before as well, but let's fix it)

src/apify/_actor.py
```diff
@@ -637,21 +637,23 @@ async def push_data(self, data: dict | list[dict], charged_event_name: str | Non
...
-        # No charging, just push the data without locking.
-        if charged_event_name is None:
-            dataset = await self.open_dataset()
-            await dataset.push_data(data)
-            return None
+        if charged_event_name and charged_event_name.startswith('apify-'):
+            raise ValueError(f'Cannot charge for synthetic event "{charged_event_name}" manually')
 
-        # If charging is requested, acquire the charge lock to prevent race conditions between concurrent
+        # Acquire the charge lock to prevent race conditions between concurrent
```
vdusek, line 644: Both flows (with and without `charged_event_name`) now execute under `_charge_lock`. However, the mixin's `_calculate_limit_for_push` + `_charge_for_items` in dataset clients operates independently of this lock. When a user calls `dataset.push_data()` directly (not through `Actor.push_data()`), the budget check and charge in the mixin are not protected by any lock from the Actor side.

Possible solution? Move the lock to the charging manager? Or expose the Actor's lock to the dataset client?

Not sure whether this is really a problem in practise. @janbuchar?

src/apify/_charging.py
Mantisus: Updated

src/apify/storage_clients/_file_system/_dataset_client.py
Mantisus: Fixed

src/apify/_charging.py
Mantisus: Updated

tests/unit/actor/test_charging_manager.py
Mantisus: Updated

src/apify/_actor.py
Mantisus: Yes, initially I thought about it and decided that using both `Actor.push_data` and `Dataset.push_data` was quite a strong mix of different approaches. But after thinking about it more deeply, I agree that even if it's a rare case, it's still worth handling.

src/apify/_charging.py
```diff
@@ -199,8 +239,10 @@ async def __aexit__(
         if not self.active:
             raise RuntimeError('Exiting an uninitialized ChargingManager')
 
+        charging_manager_ctx.set(None)
         self.active = False
 
+    @_ensure_context
     @_ensure_context
```
vdusek: Duplicate `@_ensure_context` decorators 🙂, applies to all occurences here

src/apify/storage_clients/_file_system/_dataset_client.py
```diff
@@ -0,0 +1,67 @@
...
+    ) -> Self:
+
+        dataset_client = await super().open(
+            id=id,
+            name=name,
+            alias=alias,
+            configuration=configuration,
+        )
+
+        dataset_client.is_default_dataset = all(v is None for v in (id, name, alias))
```
vdusek, line 45: The Apify client also matches when `id == configuration.default_dataset_id`.

src/apify/storage_clients/_ppe_dataset_mixin.py
```diff
@@ -0,0 +1,46 @@
...
+                event_name=DEFAULT_DATASET_ITEM_EVENT,
+                count=count_items,
+            )
+
+    @asynccontextmanager
+    async def _charge_lock(self) -> AsyncIterator[None]:
+        """Context manager to acquire the charge lock if PPE charging manager is active."""
+        charging_manager = charging_manager_ctx.get()
+        if charging_manager:
+            if charging_manager.charge_lock.locked():
```
vdusek: Hmm, `asyncio.Lock.locked()` tells whether any coroutine currently holds the lock. Doesn't that mean we would skip acquiring the lock and run unsynchronized? That would defeat the purpose of the lock in the first place. Am I missing something?

src/apify/storage_clients/_file_system/_dataset_client.py
Mantisus: Yes, because `configuration.default_dataset_id` is only used with the Apify client. Unless I'm mistaken, it is not used for the File System.

src/apify/_charging.py
Mantisus: 🤦. Fixed

src/apify/storage_clients/_ppe_dataset_mixin.py
Mantisus: When `Actor.push_data` is called, a lock is set, and `Dataset.push_data` is called inside it, which leads to this internal lock. Without this check, it will cause a deadlock; the flow will wait for the lock to be unlocked, which itself is blocked.
When calling `Dataset.push_data` directly, the lock will be set as usual.

src/apify/storage_clients/_file_system/_dataset_client.py
vdusek: you're right, thanks

src/apify/storage_clients/_ppe_dataset_mixin.py
vdusek: Thanks for the explanation.

tests/unit/actor/test_actor_charge.py
```diff
@@ -143,6 +144,104 @@ async def test_max_event_charge_count_within_limit_tolerates_overdraw() -> None:
         assert max_count == 0
 
 
+async def test_push_data_combined_price_limits_items() -> None:
+    """Test that push_data limits items when the combined explicit + synthetic event price exceeds the budget."""
+    async with setup_mocked_charging(
+        Configuration(max_total_charge_usd=Decimal('3.00'), test_pay_per_event=True)
+    ) as setup:
+        setup.charging_mgr._pricing_info['scrape'] = PricingInfoItem(Decimal('1.00'), 'Scrape')
+        setup.charging_mgr._pricing_info['apify-default-dataset-item'] = PricingInfoItem(
```
janbuchar: Perhaps the pricing info setup could also be an argument to `setup_mocked_charging`?

src/apify/storage_clients/_apify/_dataset_client.py
```diff
@@ -13,6 +13,7 @@
 from crawlee.storage_clients.models import DatasetItemsListPage, DatasetMetadata
 
 from ._api_client_creation import create_storage_api_client
+from apify.storage_clients._ppe_dataset_mixin import DatasetClientPpeMixin
```
janbuchar: I'd expect a relative import here. Also, the PPE mixin could probably live under `apify.storage_clients._apify`

src/apify/storage_clients/_file_system/_dataset_client.py
```diff
@@ -0,0 +1,67 @@
...
+            for item in items[:limit]:
+                new_item_count += 1
+                await self._push_item(item, new_item_count)
+
+            # now update metadata under the same lock
+            await self._update_metadata(
+                update_accessed_at=True,
+                update_modified_at=True,
+                new_item_count=new_item_count,
+            )
```
janbuchar: This is basically copied straight from `FileSystemDatasetClient.push_items` - couldn't you just call `super.push_items` with a truncated items array?

src/apify/_actor.py
```diff
@@ -639,21 +631,25 @@ async def push_data(self, data: dict | list[dict], charged_event_name: str | Non
...
-        async with self._charge_lock:
-            max_charged_count = self.get_charging_manager().calculate_max_event_charge_count_within_limit(
-                charged_event_name
-            )
+        async with charging_manager.charge_lock:
+            # No explicit charging requested; synthetic events are handled within dataset.push_data.
+            if charged_event_name is None:
+                dataset = await self.open_dataset()
+                await dataset.push_data(data)
+                return None
```
janbuchar: The JS version actually removed the return type overload and always returns a `ChargeResult`. The reasoning is that even without an explicit event name, you might want to know how many synthetic events were charged. 

Can you make this consistent with that?

src/apify/_actor.py
```diff
@@ -727,8 +724,9 @@ async def charge(self, event_name: str, count: int = 1) -> ChargeResult:
             count: Number of events to charge for.
         """
         # Acquire lock to prevent race conditions with concurrent charge/push_data calls.
-        async with self._charge_lock:
-            return await self.get_charging_manager().charge(event_name, count)
+        charging_manager = self.get_charging_manager()
+        async with charging_manager.charge_lock:
+            return await charging_manager.charge(event_name, count)
```
janbuchar: So `charging_manager.charge` doesn't acquire the lock by default? That would be fairly dangerous.

tests/unit/actor/test_actor_charge.py
Mantisus: Good point, updated

src/apify/storage_clients/_file_system/_dataset_client.py
Mantisus: Updated.

src/apify/_actor.py
Mantisus: Sure, updated 🙂

src/apify/_actor.py
Mantisus: I kept the original implementation here. But, yes, that could be a problem.

Updated.

PR #822, opened 2026-03-02
refactor: Use `ensure_context` decorator in Actor class

vdusek:
## Summary

- Replace manual `_raise_if_not_initialized()` calls with the `@ensure_context` decorator from Crawlee, matching the pattern already used in `ChargingManager`
- Add `active` attribute to `_ActorType` (replaces `_is_initialized`) for compatibility with `ensure_context`
- Remove the now-unused `_raise_if_not_initialized` method

Reopens #400

## Test plan

- [x] CI passes

### DIFF
--- a/src/apify/_actor.py
+++ b/src/apify/_actor.py
@@ -15,7 +15,6 @@ from pydantic import AliasChoices
 from apify_client import ApifyClientAsync
 from apify_shared.consts import ActorEnvVars, ActorExitCodes, ApifyEnvVars
 from crawlee import service_locator
-from crawlee._utils.context import ensure_context
 from crawlee.errors import ServiceConflictError
 from crawlee.events import (
     Event,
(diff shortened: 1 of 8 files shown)


Pijukatel (approved):
The tiny small comment is that we now have one more public `.active` , which is not really needed. In a gigantic class like Actor, that further increases the clutter.

What about making the `@ensure_context` decorator with a parameter?

Then we can do something like this



```
_ensure_context = ensure_context("_active")

class Actor:
    ...
    @_ensure_context
    async def open_dataset
```

vdusek:
> The tiny small comment is that we now have one more public .active , which is not really needed. In a gigantic class like Actor, that further increases the clutter.
>
> What about making the @ensure_context decorator with a parameter?

Sure, good idea. I've implemented a new version of `ensure_context` that accepts an argument for the SDK.

I implemented it in the SDK rather than changing the one in the Crawlee; this way, we don't have to restrict the lower bound version of Crawlee.

We can unify it again in the next major version of Crawlee & SDK (to have only 1 version of ensure context).

PR #855, opened 2026-04-14
chore(website): migrate to pnpm and enable minimum release age

B4nan:
## Summary

Migrates the `website/` subdirectory from Yarn 4 to pnpm 10 as part of an org-wide supply-chain hardening migration. The Python root (managed by uv) is unchanged.

Enables pnpm's `minimumReleaseAge` (1 day / 1440 minutes) to block installing packages published within the last 24h, with exclusions for first-party scopes `@apify/*` and `@crawlee/*`. Renovate is updated with `internalChecksFilter: strict` and a matching 0-day package rule so our own releases are not held back.

## Changes

- **website/package.json** — `packageManager` now `pnpm@10.24.0`; adds `name: apify-sdk-python-website` and `private: true`; all scripts rewritten from `yarn X` to `pnpm X`
- **website/pnpm-workspace.yaml** (new) — `minimumReleaseAge: 1440` plus exclusions for `@apify/*` and `@crawlee/*`
- **website/.npmrc** (new) — `node-linker=hoisted` + workspace hoisting settings to mirror the previous yarn `node-modules` layout
- **website/.yarnrc.yml, website/yarn.lock** — removed (no `.yarn/patches/` existed in this repo, so nothing was preserved)
- **website/pnpm-lock.yaml** (new) — generated via `pnpm install`
- **website/docusaurus.config.js** — renamed `future.experimental_faster` -> `future.faster` (Docusaurus 3.10 rename, forward-compatible)
- **.github/actions/pnpm-install/action.yml** (new) — composite action copied from `apify-client-js`, extended with a `working-directory` input so it can install inside `website/` without changing the caller's default directory
- **.github/workflows/_release_docs.yaml** — `git add website/yarn.lock` -> `git add website/pnpm-lock.yaml` in the auto-commit step of the docs-theme update
- **.github/workflows/manual_release_stable.yaml** — removed `corepack enable && yarn install`, installs via the new composite action with `working-directory: website`; `npx docusaurus` -> `pnpm exec docusaurus`
- **pyproject.toml** — poe tasks `update-docs-theme`, `build-docs`, `run-docs` now invoke `pnpm` directly (no more `corepack enable && yarn`)
- **renovate.json** — adds `internalChecksFilter: strict` and a `packageRule` matching `@apify/*` / `@crawlee/*` with `minimumReleaseAge: 0 days`. `ignoreDeps` unchanged (did not contain `yarn`).
- **.gitignore** — replaced `website/.yarn` with `website/.pnpm-store`

## Verification

- `pnpm install` and `pnpm install --frozen-lockfile` both succeed in `website/`
- Existing peer-dependency warnings (eslint v10, typescript v6, styled-components) pre-date this migration and are carried over unchanged
- Local docusaurus build was attempted but could not complete outside CI because the typedoc plugin spawns `python` (not `python3`) and this laptop doesn't expose `python` on PATH; in CI `actions/setup-python` provides `python` so this will resolve there. This is orthogonal to the pnpm migration.
- `pnpm lint` surfaces an eslint v10 flat-config error that pre-dates this PR (no `eslint.config.js` present). Not introduced here.


vdusek:
could you do it for crawlee-python as well?

renovate.json
```diff
@@ -19,9 +19,14 @@
             "groupSlug": "dev-dependencies",
             "automerge": true,
             "automergeType": "branch"
+        },
+        {
+            "matchPackageNames": ["@apify/*", "@crawlee/*"],
+            "minimumReleaseAge": "0 days"
```
Pijukatel, line 25: I am not sure how this will work with python related stuiff and UV. I think we should add to pyproject.toml

```
  [tool.uv.exclude-newer-package]                                                                                                                                                                           
  ...our packages...  = false
```

Pijukatel: This specifically
```
[tool.uv.exclude-newer-package]
apify-client = false 
apify-shared = false 
crawlee = false 
apify_fingerprint_datapoints = false
```

B4nan: This is here mainly for the docs. Good point that we should extend it to allow the client/sdk too, that's not covered at all. Note that this is renovate config, we want this in both places ideally.

Pijukatel: Yes, but without this if we turn on renovate, it will do weird stuff on Python deps as renovate will take newest ones, but uv will refuse them. So if we update renovate we need to update `pyproject.toml` as well

Pijukatel (approved):

vdusek (commented):
From Claude:

One inconsistency compared to apify-client-python #724:

This PR correctly replaces `website/.yarn` with `website/.pnpm-store` in `.gitignore`, but apify-client-python only removes `website/.yarn` without adding `website/.pnpm-store`. Just flagging in case you want to fix it there.

Otherwise this looks consistent with the other repos.

B4nan:
Is this automated? I don't need the same hint 3 times 🙃 I am managing this from a single session, not per repo.

B4nan:
So `.pnpm-store` shouldn't be necessary, pnpm wont use a project scoped folder by default. Also frozen lockfile flag was misused, it is supposed to be used in the CI which is handled in the shared install action.

vdusek (approved):

PR #857, opened 2026-04-15
ci: Extract docs versioning into reusable `manual_version_docs.yaml` workflow

vdusek:
## Summary

- Extract the inline `version_docs` job from `manual_release_stable.yaml` into a standalone `manual_version_docs.yaml` workflow
- The new workflow can be triggered manually (for docs fixes) or called from the release pipeline
- Fix `api:version` ENOENT bug by running docusaurus commands through `uv run` (Python deps not available to bare `npx`)
- Clean up ALL versions for the same major version, not just exact major.minor match
- Remove non-docs artifacts (pyproject.toml, .gitignore, caches) from versioned doc snapshots

### DIFF
--- a/.github/workflows/manual_release_stable.yaml
+++ b/.github/workflows/manual_release_stable.yaml
@@ -105,68 +105,12 @@ jobs:
   version_docs:
     name: Version docs
     needs: [release_prepare, changelog_update, pypi_publish]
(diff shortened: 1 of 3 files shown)


Pijukatel (approved):
Test in production

PR #866, opened 2026-04-18
ci: Fix docs release failing on detached HEAD

vdusek:
## Summary

Same fix as apify/apify-client-python#741 — the SDK workflows have the identical bug.

When `manual_release_docs.yaml` is invoked with a commit SHA as `ref` (from `doc_release_post_publish` in `on_master.yaml` or from `manual_release_stable.yaml`), `actions/checkout` leaves the repo in detached HEAD state and the hand-rolled `Commit the updated package.json and lockfile` step's `git push` fails with `fatal: You are not currently on a branch.`

The same latent bug exists in `manual_version_docs.yaml`'s `EndBug/add-and-commit` step (no `new_branch` set). Neither has fired yet because no `feat`/`fix`/`perf`/`refactor`/`style` commit has landed on master since the reusable-workflow extraction merged.

## Fix

- Replace the hand-rolled git block in `manual_release_docs.yaml` with `EndBug/add-and-commit@v10`.
- Set `new_branch: ${{ github.event.repository.default_branch }}` in both workflows — EndBug does not handle detached HEAD by itself (its default push is `git push origin --set-upstream` with no refspec, which fails the same way). Setting `new_branch` makes it create/checkout a local default branch at current HEAD before committing.

### DIFF
--- a/.github/workflows/manual_release_docs.yaml
+++ b/.github/workflows/manual_release_docs.yaml
@@ -74,13 +74,13 @@ jobs:
         run: uv run poe update-docs-theme
 
       - name: Commit the updated package.json and lockfile
-        run: |
-          git config user.name 'GitHub Actions'
-          git config user.email 'github-actions[bot]@users.noreply.github.com'
-          git add website/package.json
-          git add website/yarn.lock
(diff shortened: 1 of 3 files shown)


vdusek:
exactly the same changes as in https://github.com/apify/apify-client-python/pull/741 that were already approved

PR #867, opened 2026-04-20
fix: avoid InvalidStateError on mid-stream platform websocket disconnect

vdusek:
## Summary

Guard `set_result(False)` in `ApifyEventManager._process_platform_messages` with a `done()` check. Previously, a websocket disconnect *after* a successful connect would try to re-resolve the already-resolved `_connected_to_platform_websocket` future, raising `InvalidStateError` and masking the original `ConnectionClosedError` in logs.

Adds a regression test that forces a mid-stream abnormal close from the server and asserts the background task does not surface `InvalidStateError`.

### DIFF
--- a/tests/unit/events/test_apify_event_manager.py
+++ b/tests/unit/events/test_apify_event_manager.py
@@ -30,14 +30,12 @@ async def _platform_ws_server(
 ) -> AsyncGenerator[tuple[set[websockets.asyncio.server.ServerConnection], asyncio.Event]]:
     """Create a local WebSocket server that simulates Apify platform events.
 
-    Binds explicitly to ``127.0.0.1`` instead of ``localhost`` so that only a
-    single IPv4 socket is created.  On Windows, ``localhost`` resolves to both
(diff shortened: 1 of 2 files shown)


tests/unit/events/test_apify_event_manager.py
```diff
@@ -316,6 +314,33 @@ def migrating_listener(data: Any) -> None:
...
+    later drops, the outer `except` in `_process_platform_messages` must not call `set_result(False)` on the
+    already-resolved future.
+    """
+    async with (
+        _platform_ws_server(monkeypatch) as (connected_ws_clients, client_connected),
+        ApifyEventManager(Configuration.get_global_configuration()) as event_manager,
+    ):
+        await client_connected.wait()
+
+        # Force an abnormal close from the server so the client's `async for` raises ConnectionClosedError.
```
Pijukatel, line 333: Should we maybe assert that the error is `ConnectionClosedError` in the end of the test?

Pijukatel (approved):

PR #870, opened 2026-04-20
fix: preserve `max_total_charge_usd=0` instead of treating it as unlimited

vdusek:
## Summary

`configuration.max_total_charge_usd or Decimal('inf')` at `src/apify/_charging.py:162`, `:416`, and `:439` treats `Decimal(0)` as falsy and silently replaces it with `Decimal('inf')`. A user who explicitly sets `max_total_charge_usd=0` (or `ACTOR_MAX_TOTAL_CHARGE_USD=0`) to disable charging ends up with **unlimited** charging — the opposite of the intended behavior.

The configuration layer already preserves `0` as a distinct value (and `test_max_total_charge_usd_zero_is_preserved` asserts this), so the invariant is thrown away only at the charging manager boundary.

Replaced the three occurrences with an explicit `… if … is not None else Decimal('inf')` check, matching the already-correct pattern used on the API-fetch path (line 438).

### DIFF
--- a/src/apify/_charging.py
+++ b/src/apify/_charging.py
@@ -159,7 +159,9 @@ class ChargingManagerImplementation(ChargingManager):
     LOCAL_CHARGING_LOG_DATASET_NAME = 'charging-log'
 
     def __init__(self, configuration: Configuration, client: ApifyClientAsync) -> None:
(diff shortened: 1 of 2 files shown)


Pijukatel (approved):
Good catch, I think this deserves a test.

PR #884, opened 2026-05-03
test: update `test_request_deduplication_edge_cases`

### LINKED ISSUE #786: Test `test_request_deduplication_edge_cases[shared]` is flaky
Test `test_request_deduplication_edge_cases[shared]` is flaky. An `IndexError` inside Crawlee's `RequestQueue.add_request()`) can occur, see https://github.com/apify/apify-sdk-python/actions/runs/21956322870/job/63421260583?pr=784. Skip it for now.

Mantisus:
Closes: #786

Re-enables `test_request_deduplication_edge_cases` in shared mode (previously skipped as flaky).

The flakiness came from `add_request` occasionally returning `None` transiently due to platform-side propagation in shared request-queue mode. Instead of a fixed retry loop with no delay between attempts, the test now uses the `poll_until_condition` helper, which polls `add_request` with backoff until it returns a result (or times out). This gives transient `None` results time to clear and is consistent with how other integration tests handle eventually-consistent API state.

### DIFF
--- a/tests/integration/test_request_queue.py
+++ b/tests/integration/test_request_queue.py
@@ -743,11 +743,9 @@ async def test_request_deduplication_edge_cases(request_queue_apify: RequestQueu
 
     results = list[bool]()
     for url, expected_duplicate in urls_and_deduplication_expectations:
-        # add_request may transiently return None due to platform-side issues, retry up to 3 times.
-        for _ in range(3):
-            result = await rq.add_request(url)
-            if result is not None:
-                break  # Successfully added.
+        # In shared mode, `add_request` may transiently return None until the operation propagates,
+        # so poll with backoff until it returns a result.
+        result = await poll_until_condition(lambda url=url: rq.add_request(url), lambda result: result is not None)
 
         assert result is not None
         results.append(result.was_already_present)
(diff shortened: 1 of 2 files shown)


vdusek (changes requested):
Thanks, but I would rather see the root cause investigation and actually trace where the empty `processedRequests` comes from, rather than retry in test workaround 🙂.

janbuchar:
> Thanks, but I would rather see the root cause investigation and actually trace where the empty `processedRequests` comes from, rather than retry in test workaround 🙂.

It's a frustrating answer, but due to the nature of the Request Queue, this sometimes just... happens. It is an established fact that the API leaks implementation details. Maybe we should make sure we plug that leak at the SDK level at least, though.

vdusek:
> It's a frustrating answer, but due to the nature of the Request Queue, this sometimes just... happens. It is an established fact that the API leaks implementation details. Maybe we should make sure we plug that leak at the SDK level at least, though.

Yeah, in that case, we should try to address this in the `ApifyStorageClient`

vdusek (approved):
I just utilized the `poll_until_condition`, otherwise LGTM.

PR #896, opened 2026-05-13
chore: bump docusaurus-plugin-typedoc-api

barjin:
Bumps the API reference rendering plugin version.


vdusek (approved):
thanks, just let's use "docs" so that it gets updated 🙂

PR #899, opened 2026-05-15
fix: Update `push_data` annotation with `JsonSerializable` instead of Any

Mantisus:
Follow-up: https://github.com/apify/crawlee-python/pull/1889


Mantisus:
@vdusek, well, it caused issues, and we should update the SDK typing after Crawlee is updated.

vdusek:
Hi @Mantisus , could you please make this ready for a review? Let's release a patch of Crawlee, and the last patch of SDK, and then jump into BCs for SDK v4 :slightly_smiling_face: .

Mantisus:
> Hi @Mantisus , could you please make this ready for a review? Let's release a patch of Crawlee, and the last patch of SDK, and then jump into BCs for SDK v4 🙂 .

It's ready, but we need to release the crawlee patch first so I can unlink it from the master branch. 🙂

vdusek:
Makes sense. I was waiting for [apify/crawlee-python#1904](https://github.com/apify/crawlee-python/issues/1904) to be merged. Now there is v1.7.1 - and just please refresh the lock file without bumping the constraints in `pyproject.toml`. If I am not mistaken, this isn't a runtime issue, only type checking, so we don't want to constrain versions over it.

vdusek (approved):
otherwise LGTM

Mantisus:
> If I am not mistaken, this isn't a runtime issue, only type checking, so we don't want to constrain versions over it.

Right. Updated lock file.

PR #900, opened 2026-05-15
docs: use the common `ApiLink` component from the shared theme

### LINKED ISSUE #889: `ApiLink` generates incorrect URL for versioned docs
`ApiLink` always resolves to `/reference/<...>` regardless of the active Docusaurus version. For the next version, it should resolve to `/reference/next/<...>`.

This causes the docs CI check to fail when a PR introduces new public classes, since they don't exist in the stable API docs yet.

Parity tracking for https://github.com/apify/crawlee-python/issues/1867 and https://github.com/apify/apify-client-python/issues/780.

---

Note - the right way to go about this is likely moving `ApiLink` to a shared theme. Currently, there are small differences between the `ApiLink` implementations in the projects, though, so it's not straightforward.

barjin:
Uses the common `ApiLink` implementation from the shared theme package.

Closes #889


vdusek (approved):
lgtm

PR #906, opened 2026-05-20
fix: forward missing run kwargs in Actor.start/call/call_task

vdusek:
## Summary

`ActorClient.call` (apify-client) exposes several run-configuration kwargs that the SDK wrapper `Actor.call` was silently dropping, forcing users to bypass the SDK and reach into `Actor.apify_client.actor(id).call(...)` (e.g. to set `max_total_charge_usd` for pay-per-event Actors — see the [user report](#)).

This patch wires through the missing parameters so the SDK API matches the underlying client:

- `Actor.start` / `Actor.call` — adds `max_items`, `max_total_charge_usd`, `restart_on_error`, `force_permission_level`
- `Actor.call_task` — adds `max_items`, `restart_on_error` (the task endpoint has no `max_total_charge_usd` / `force_permission_level`)

Docstrings updated to match. No change to existing kwargs or call semantics; pyproject `apify-client` constraint unchanged.

### DIFF
--- a/src/apify/_actor.py
+++ b/src/apify/_actor.py
@@ -42,10 +42,12 @@ from apify.storages import Dataset, KeyValueStore, RequestQueue
 if TYPE_CHECKING:
     import logging
     from collections.abc import Callable
+    from decimal import Decimal
     from types import TracebackType
 
     from typing_extensions import Self
 
+    from apify_shared.consts import ActorPermissionLevel
     from crawlee._types import JsonSerializable
     from crawlee.proxy_configuration import _NewUrlFunction
 
@@ -863,8 +865,12 @@ class _ActorType:
         token: str | None = None,
         content_type: str | None = None,
         build: str | None = None,
+        max_items: int | None = None,
+        max_total_charge_usd: Decimal | None = None,
+        restart_on_error: bool | None = None,
         memory_mbytes: int | None = None,
         timeout: timedelta | None | Literal['inherit', 'RemainingTime'] = None,
+        force_permission_level: ActorPermissionLevel | None = None,
         wait_for_finish: int | None = None,
         webhooks: list[Webhook] | None = None,
     ) -> ActorRun:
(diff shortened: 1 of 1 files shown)


MatousMarik (approved):
Seems ok to me.

One concern though: this approach of manually mirroring kwargs from the
client is fragile — next time the client adds a parameter, someone has to
remember to update the SDK too. Would it be possible to extract the
"passthrough" subset into a shared TypedDict and use **kwargs:
Unpack[RunConfig]? That would make the boundary explicit and future-safe.

Also, was a full audit done against the current client signatures? This PR
fixes the reported case, but are we confident there are no other missing
args across start / call / call_task right now?

src/apify/_actor.py
```diff
@@ -863,8 +865,12 @@ async def start(
         token: str | None = None,
         content_type: str | None = None,
         build: str | None = None,
+        max_items: int | None = None,
```
janbuchar: PPR is effectively removed now. Maybe we don't need to add this parameter?

janbuchar (approved):

src/apify/_actor.py
vdusek: Not a strong opinion, but I'd prefer to mirror the behavior of the client (and the API) here. If PPR is being removed, the right path is to introduce a deprecation warning pointing users to PPE, and then drop it in the next major. So I'd rather keep it for now and follow a proper deprecation process later.

vdusek:
> One concern though: this approach of manually mirroring kwargs from the client is fragile — next time the client adds a parameter, someone has to remember to update the SDK too. Would it be possible to extract the
"passthrough" subset into a shared TypedDict and use **kwargs: Unpack[RunConfig]? That would make the boundary explicit and future-safe.

It's fragile, but that's the trade-off if we want to propagate methods and arguments downstream. This is the approach that was established here before I joined, so we're just continuing consistently. The typed-dict alternative has its own drawbacks, too. If we want to rethink this, I would challenge the "propagate methods and arguments downstream" approach itself.

> Also, was a full audit done against the current client signatures? This PR fixes the reported case, but are we confident there are no other missing args across start / call / call_task right now?

Everything should be addressed now (if I didn't overlook anything).

src/apify/_actor.py
janbuchar: But you're not keeping it, you're adding it :slightly_smiling_face:

vdusek: alrighr, removing

PR #911, opened 2026-05-26
ci: Allow manually dispatching the master Checks workflow

vdusek:
## Summary

Adds `workflow_dispatch` to `CI (master)` so the `Checks` suite can be re-run manually on the same commit when the push-triggered run is lost — e.g. during a GitHub outage. The `doc_release` / `beta_release` jobs are gated on `github.event_name == 'push'` so manual dispatches only re-run the checks and never re-publish.

The redundant `workflow_dispatch` trigger on `_checks.yaml` is removed: dispatching it directly produced check-run names without the `Checks /` prefix, so the `wait-for-checks` gate in the manual release workflows wouldn't recognize them anyway.

Mirrors apify/crawlee-python#1917.

## How to use

When a push-triggered Checks run on `master` is missing, go to **Actions → CI (master) → Run workflow** on `master`. Once it succeeds, the manual release workflows' `wait-for-checks` gate will be satisfied.

### DIFF
--- a/.github/workflows/_checks.yaml
+++ b/.github/workflows/_checks.yaml
@@ -1,16 +1,9 @@
 name: Checks
 
 on:
-  # Runs when manually triggered from the GitHub UI.
-  workflow_dispatch:
-    inputs:
-      run_tests:
-        description: Whether to run the test suites (unit, integration, E2E).
-        required: false
-        type: boolean
-        default: true
-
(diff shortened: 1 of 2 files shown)


janbuchar (approved):
LGTM, but see https://github.com/apify/apify-client-python/pull/828#discussion_r3318786270

PR #917, opened 2026-05-29
refactor!: Mark secondary arguments as keyword-only

### LINKED ISSUE #881: Mark secondary arguments as keyword-only
Reshape function/method signatures across the public API so that secondary parameters must be passed as keyword arguments. Primary "subject" arguments stay positional.

This mirrors what was done in the client: apify/apify-client-python#766.

## Why

Keyword-only parameters at API boundaries make call sites self-documenting and prevent breakage when new options are added between existing arguments.

vdusek:
## Summary

Closes #881.

Reshapes function/method signatures across the SDK public API so that secondary parameters must be passed as keyword arguments. Primary "subject" arguments (e.g. `key`, `data`, `event_name`) stay positional. Mirrors what was done in the client: apify/apify-client-python#766.

### Affected APIs

`Actor`:

- `get_value(key, *, default_value=None)`
- `push_data(data, *, charged_event_name=None)`
- `charge(event_name, *, count=1)`
- `use_state(default_value=None, *, key=None, kvs_name=None)`

`ChargingManager` / `ChargingManagerImplementation` (returned by `Actor.get_charging_manager()`):

- `charge(event_name, *, count=1)`

Crawlee-overriding methods (e.g. `ProxyConfiguration.new_proxy_info`, the storage clients) were intentionally left untouched — Crawlee's base signatures are positional, so making the overrides keyword-only would diverge from the base class.

## Why

Keyword-only parameters at API boundaries make call sites self-documenting and prevent breakage when new options are added between existing arguments.

**BREAKING CHANGE** for v4.0 — see `docs/04_upgrading/upgrading_to_v4.md` for the migration guide.


docs/04_upgrading/upgrading_to_v4.md
```diff
@@ -9,3 +9,24 @@ This guide lists the breaking changes between Apify Python SDK v3.x and v4.0.
 ## Python 3.11+ required
 
 Support for Python 3.10 has been dropped. The Apify Python SDK v4.x now requires Python 3.11 or later — make sure your environment is on a compatible version before upgrading.
+
+## Keyword-only arguments
+
+Secondary parameters on these methods can no longer be passed positionally.
```
szaganek:
```suggestion
Secondary parameters in these signatures can no longer be passed positionally:
- `Actor` — `get_value`, `push_data`, `charge`, `use_state`.
- `ChargingManager` — `charge`.
```
It's a little awkward that we start with `these methods` but don't list them right away. How about something like this?

docs/04_upgrading/upgrading_to_v4.md
```diff
@@ -9,3 +9,24 @@ This guide lists the breaking changes between Apify Python SDK v3.x and v4.0.
...
+
+```python
+# Before
```
szaganek:
```suggestion
# Before (v3)
```

docs/04_upgrading/upgrading_to_v4.md
```diff
@@ -9,3 +9,24 @@ This guide lists the breaking changes between Apify Python SDK v3.x and v4.0.
...
+value = await Actor.get_value('my-key', default_value)
+await Actor.push_data(data, 'my-event')
+await Actor.charge('my-event', 5)
+
+# After
```
szaganek:
```suggestion
# After (v4)
```

src/apify/_actor.py
```diff
@@ -625,7 +625,7 @@ async def open_request_queue(
         )
 
     @_ensure_context
-    async def push_data(self, data: dict | list[dict], charged_event_name: str | None = None) -> ChargeResult:
+    async def push_data(self, data: dict | list[dict], *, charged_event_name: str | None = None) -> ChargeResult:
```
janbuchar, line 627: Not so sure about this — `charged_event_name` is pretty long. Since this is a pretty important part of the API, can you please get an OK from someone in the Publishers team perhaps? Or whoever is responsible for PPE...

vdusek: Alright, let's try @stetizu1 - Zuzka, could you give us your opinion, please?

TL;DR: The trade-off here is making the public interface more verbose by enforcing named arguments for `charged_event_name`; in exchange, we get a more stable API, with the flexibility to add future parameters in any way without breaking changes.

vdusek:
All addressed, thanks @szaganek.

src/apify/_actor.py
stetizu1: I don't have strong opinions there. I agree it is long but overall prefer verbose / more explained things

janbuchar (approved):

PR #937, opened 2026-06-05
test: retry flaky webhook E2E test on platform-side timeouts

vdusek:
The E2E test `test_actor_adds_webhook_and_receives_event` is flaky. The server Actor blocks until the Apify platform delivers the webhook fired by the client Actor's `ACTOR_RUN_SUCCEEDED` event; when platform-side delivery is delayed past the server run's 600s timeout, the run ends as `TIMED-OUT` and the assertion fails. This has been observed on both `master` (run 26950690902) and an unrelated docs PR (run 26959362332) on the same day, each burning the full timeout with no delivery.

There is no test-logic root cause to fix — the latency is platform-side and outside the SDK's control. This adds `pytest-rerunfailures` and marks the test `@pytest.mark.flaky(reruns=3)`, so a transient delivery hiccup is retried instead of failing CI. A comment on the test documents the failure mode.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -78,7 +78,7 @@ dev = [
     "pytest-asyncio<2.0.0",
     "pytest-cov<8.0.0",
     "pytest-httpserver<2.0.0",
-    "pytest-rerunfailures<16.0.0",
+    "pytest-rerunfailures<17.0.0",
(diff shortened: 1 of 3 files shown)


vdusek:
tmp issue with webhooks on platform, closing

PR #944, opened 2026-06-05
test: accept platform auto-abort in test_actor_charge_limit

vdusek:
## Description

The E2E test `test_actor_charge_limit` was flaky: it asserted the run ended as `SUCCEEDED`, but occasionally observed `ABORTED` (e.g. [this CI run](https://github.com/apify/apify-sdk-python/actions/runs/27021762318/job/79751266145?pr=941)).

### Root cause

When an Actor reaches its `max_total_charge_usd` ceiling, the **platform automatically aborts the run** (*"Aborted automatically after reaching the maximum cost of the run $0.20"*). That auto-abort races with the Actor's own clean exit (`exit_code 0`), so the terminal status is non-deterministically `SUCCEEDED` or `ABORTED`. It's a platform-side timing race, not an SDK or test-logic bug — the actual behavior under test (the charge limit capping the run at exactly 2 of the 4 attempted events) held fine in both outcomes.

### Fix

Accept both `SUCCEEDED` and `ABORTED` as valid terminal statuses, keying the wait/assert on `charged_event_counts == {'foobar': 2}` (the real invariant).


Pijukatel (commented):
Maybe it would be better to avoid the non-determinism in how the actor can exit. `ppe_actor` could wait for all events (4) to propagate to the platform. This should ensure, that it will always be aborted with 2 events only.

To me, this condition would make more sense in the `ppe_actor` itself:

```
    run = await poll_until_condition(
        ...self.run...,
        lambda r: r.charged_event_counts == {'foobar': 4},
        timeout=30,
        poll_interval=1,
    )
```

Pijukatel (approved):

PR #719, opened 2025-12-23
refactor!: Adapt to apify-client v3

vdusek:
### Description

- This PR updates the SDK to work with `apify-python-client` v3, which introduces fully typed API clients generated from OpenAPI specifications.
- See https://github.com/apify/apify-client-python/pull/604 for more details.

### Issues

- Closes: #736
- Closes: #770
- Closes: #697
- Closes: #853

### Testing

- The existing SDK tests pass with `apify-python-client` v3.

PR #950, opened 2026-06-08
refactor: derive Pydantic model field aliases with to_camel

### LINKED ISSUE #949: Use `alias_generator=to_camel` on Pydantic models instead of manual per-field aliases
### Background

We're on Pydantic ≥2.11. Models in `src/apify/_models.py` and `src/apify/storage_clients/_apify/_models.py` map snake_case fields to the Apify API's camelCase, declaring each alias by hand:

```python
actor_id: Annotated[str | None, Field(alias='actorId')] = None
actor_task_id: Annotated[str | None, Field(alias='actorTaskId')] = None
```

### Proposal

Add `alias_generator=to_camel` to the shared config so aliases are derived:

```python
from pydantic.alias_generators import to_camel

model_config = ConfigDict(populate_by_name=True, alias_generator=to_camel)
```

Fields then become `actor_id: str | None = None` and still bind to `actorId`. Currently 19 models repeat `ConfigDict(populate_by_name=True, ...)` while 11 (events in `events/_types.py`, the request-loader input models) have no config at all. A single base model carrying the config (including the generator) would cover both gaps.

### Where it makes sense

- Apply to the API-response models that already use camelCase aliases.
- Keep explicit `Field(alias=...)` where the generated key would differ.
- Subclasses of Crawlee storage models (`ApifyKeyValueStoreMetadata`, `ApifyRequestQueueMetadata`) should follow whatever Crawlee adopts.
- Event models carry protocol-defined names, so check each before switching.

### Scope

Refactor only. Generated aliases must match the current ones, guarded by tests.

vdusek:
Closes #949.

Replaces hand-written per-field `Field(alias=...)` declarations with Pydantic's `alias_generator=to_camel`, so each model derives its camelCase JSON aliases from its snake_case field names.

Following the existing convention in this repo, the config (`ConfigDict(populate_by_name=True, alias_generator=to_camel, ...)`) is repeated per model rather than introduced through a shared base class.

### Scope

- `_charging.py` — the pricing-info and charge-event models.
- `events/_types.py` — `SystemInfoEventData` and the event wrapper models.
- `request_loaders/_apify_request_list.py` — the request-input models.
- `storage_clients/_apify/_models.py` — `ApifyKeyValueStoreMetadata`, `RequestQueueHead`, `CachedRequest`, `ApifyRequestQueueMetadata`.

### No breaking change

The generated aliases match the aliases that were declared by hand, so validation and serialization keep the same wire keys. Attribute names are unchanged (`run.usage.actor_compute_units` and friends still work).

The one field whose name differs from its wire key is `RequestQueueHead.lock_time`. The API sends `lockSecs`, while `to_camel` would derive `lockTime`. It keeps `serialization_alias='lockSecs'` so it still emits `lockSecs`, and it now validates from both `lockSecs` and `lockTime` via `AliasChoices`. That change is additive. Nothing that parsed before stops parsing.

`Configuration` is intentionally left untouched. Its fields bind to `APIFY_*` / `CRAWLEE_*` environment variables through validation aliases, so applying `to_camel` there would break env-based configuration.


vdusek:
Finish this after merging https://github.com/apify/apify-sdk-python/pull/719

vdusek:
@Pijukatel

Pijukatel (approved):

PR #951, opened 2026-06-09
fix(scrapy)!: serialize requests and HTTP cache as JSON instead of pickle

vdusek:
Scrapy requests (in the Apify request queue) and cached responses (in the Scrapy HTTP cache) were serialized with pickle, but both storages hold JSON. This switches them to JSON via a shared serializer (`apify/scrapy/_serialization.py`).

Request serialization:

- Binary fields (`body`, `headers`) are base64-encoded; pydantic models such as Crawlee's `UserData` are dumped via `model_dump`.
- A non-JSON-serializable `meta`/`cb_kwargs` is logged and the request skipped, instead of being silently corrupted.
- A serialized `_class` is honored only when it is already imported as a `scrapy.Request` subclass.

HTTP cache, so existing caches survive the upgrade:

- A legacy pickle entry fails to load, so reads treat it as a cache miss and re-fetch it, instead of crashing the download.
- The cleanup sweep now deletes stale, expired, or malformed entries. It previously overwrote them with `set_value(..., None)`, which under apify-client v3 stores JSON `null` rather than deleting, leaking dead records.

**BREAKING CHANGE:** requests and HTTP cache entries are now stored as JSON, not pickle. Entries written by older versions are ignored (re-fetched), not read.


Pijukatel (approved):

docs/03_guides/06_scrapy.mdx
```diff
@@ -46,7 +46,7 @@ The Apify SDK provides several custom components to support integration with the
...
-- <ApiLink to="class/ApifyCacheStorage">`apify.scrapy.extensions.ApifyCacheStorage`</ApiLink> - A storage backend for Scrapy's built-in [HTTP cache middleware](https://docs.scrapy.org/en/latest/topics/downloader-middleware.html#module-scrapy.downloadermiddlewares.httpcache). This backend uses Apify's [key-value store](https://docs.apify.com/platform/storage/key-value-store). Make sure to set `HTTPCACHE_ENABLED` and `HTTPCACHE_EXPIRATION_SECS` in your settings, or caching won't work.
+- <ApiLink to="class/ApifyCacheStorage">`apify.scrapy.extensions.ApifyCacheStorage`</ApiLink> - A storage backend for Scrapy's built-in [HTTP cache middleware](https://docs.scrapy.org/en/latest/topics/downloader-middleware.html#module-scrapy.downloadermiddlewares.httpcache). This backend uses Apify's [key-value store](https://docs.apify.com/platform/storage/key-value-store). Make sure to set `HTTPCACHE_ENABLED` and `HTTPCACHE_EXPIRATION_SECS` in your settings, or caching won't work. When the spider closes, expired and unreadable entries are cleaned up, up to `APIFY_HTTPCACHE_EXPIRATION_MAX_ITEMS` entries per run (default 100).
```
szaganek:
```suggestion
- <ApiLink to="class/ApifyCacheStorage">`apify.scrapy.extensions.ApifyCacheStorage`</ApiLink> - A storage backend for Scrapy's built-in [HTTP cache middleware](https://docs.scrapy.org/en/latest/topics/downloader-middleware.html#module-scrapy.downloadermiddlewares.httpcache). This backend uses Apify's [key-value store](https://docs.apify.com/platform/storage/key-value-store). To enable caching, set `HTTPCACHE_ENABLED` and `HTTPCACHE_EXPIRATION_SECS` in your settings. By default, when the spider closes, up to 100 expired and unreadable entries per run are cleaned up. To change this number, update `APIFY_HTTPCACHE_EXPIRATION_MAX_ITEMS`.
```

szaganek (approved):

PR #952, opened 2026-06-09
fix(scrapy): skip a request that fails to convert instead of crashing the run

vdusek:
`ApifyScheduler.next_request` marked an Apify request as handled before converting it to a Scrapy request, so a single malformed queue entry raised and crashed the whole run (after the entry had already been consumed). The conversion is now wrapped in a try/except: it logs the failure and skips the request (returns `None`) instead of propagating.


Pijukatel (commented):
Can you please add a test for it.

Pijukatel (approved):

PR #954, opened 2026-06-09
fix(scrapy): make logging configuration idempotent

vdusek:
`initialize_logging` re-wrapped `scrapy.utils.log.configure_logging` on every call, stacking wrappers when it ran more than once. The monkey-patch is now installed at most once and reads the current handler and level from a module-level state dict, so repeated initialization stays correct without nesting wrappers.


src/apify/scrapy/_logging_config.py
```diff
@@ -23,26 +28,39 @@ def _configure_logger(name: str | None, logging_level: str, handler: logging.Han
     logger.propagate = False
 
 
+def _configure_all_loggers() -> None:
```
Pijukatel, line 31: Maybe I am missing some context, but can't we skip the module-level dict and do it like this?

```
@lru_cache
def _configure_all_loggers(handler, level):
    if handler is None:
        return
    for logger_name in [None, *_ALL_LOGGERS]:
        _configure_logger(logger_name, level, handler)
```

vdusek: Unfortunately, we can't do that - the `_configure_all_loggers` needs to run more than once.

The expected flow of the Scrapy Actors is something like this:

1. We configure logging (our handler + levels).
2. We do some setup, during which things may already log.
3. Scrapy's own setup runs and overrides our logging configuration.
4. We re-apply our configuration on top to override it back.

So the current code is the correct way.

vdusek:
@Pijukatel

Pijukatel (approved):

PR #955, opened 2026-06-09
refactor(scrapy): make AsyncThread timeout configurable

vdusek:
`AsyncThread.run_coro` hardcoded a 60s timeout. This adds a configurable `default_timeout` constructor argument (default unchanged at 60s), used when a per-call timeout is not given, and documents why each Scrapy consumer (scheduler, HTTP cache) owns its own event-loop thread.

### DIFF
--- a/src/apify/scrapy/_async_thread.py
+++ b/src/apify/scrapy/_async_thread.py
@@ -14,13 +14,18 @@ logger = getLogger(__name__)
 
 
 class AsyncThread:
-    """Class for running an asyncio event loop in a separate thread.
+    """Run an asyncio event loop in a dedicated background thread.
 
-    This allows running asynchronous coroutines from synchronous code by executingthem on an event loop
-    that runs in its own dedicated thread.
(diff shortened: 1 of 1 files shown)


src/apify/scrapy/_async_thread.py
```diff
@@ -33,7 +36,7 @@ def __init__(self) -> None:
     def run_coro(
         self,
         coro: Coroutine,
-        timeout: timedelta = timedelta(seconds=60),
+        timeout: timedelta | None = None,
```
Pijukatel: `run_coro(..., timeout=None)` can be missleading. It suggests that the coro will be running without a timeout.

I would prefer to be more verbose about it


```
   def run_coro(
        self,
        coro: Coroutine,
        timeout: timedelta | Literal['default'] = 'default'),
```

This call is very explicit: `run_coro(..., timeout='default')` and does not leave much room for interpretation

vdusek: Yeah, this makes sense

Pijukatel (approved):

PR #957, opened 2026-06-09
fix(scrapy): Tolerate malformed HTTP cache items and delete expired ones

vdusek:
### What

Make the Scrapy `ApifyCacheStorage` resilient to cache entries it cannot deserialize, and fix the cleanup so it actually removes stale entries.

- `retrieve_response` now treats any undecodable cache item as a cache miss (`return None`) instead of letting the exception propagate. Previously a single bad entry raised out of the cache middleware and turned every request (including the initial robots.txt fetch) into a download error, so the spider scraped nothing.
- `close_spider`/`expire_kvs` now use `delete_value` instead of `set_value(..., None)` to purge expired/malformed items. Under apify-client v3, `set_value(..., None)` no longer deletes the record — it stores a JSON `null` — so bad items were never cleaned up and piled up in the shared key-value store.

### Why

The e2e test `test_actor_scrapy_title_spider` was failing with `assert 0 >= 9`. The HTTP cache (a persistent named KVS keyed only by spider name, shared across all branches/runs) contained legacy `gzip(JSON)` entries written by another branch's experiments. The pickle-based reader hit them and raised `UnpicklingError: invalid load key, '{'`, which broke the whole crawl. The ineffective `set_value(..., None)` cleanup meant the store could never self-heal.

### Tests

Added two `retrieve_response` unit tests: a malformed `gzip(JSON)` entry resolves to a cache miss, and a valid entry still returns a proper response.

### DIFF
--- a/src/apify/scrapy/extensions/_httpcache.py
+++ b/src/apify/scrapy/extensions/_httpcache.py
@@ -86,11 +86,11 @@ class ApifyCacheStorage:
                         gzip_time = read_gzip_time(value)
                     except Exception as e:
                         logger.warning(f'Malformed cache item {item.key}: {e}')
-                        await self._kvs.set_value(item.key, None)
+                        await self._kvs.delete_value(item.key)
                     else:
                         if self._expiration_secs < current_time - gzip_time:
                             logger.debug(f'Expired cache item {item.key}')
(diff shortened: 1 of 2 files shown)


vdusek:
closing in favor of https://github.com/apify/apify-sdk-python/pull/951

PR #961, opened 2026-06-11
fix(scrapy): dump pydantic models in JSON mode when serializing requests

vdusek:
Pydantic models in a Scrapy request's `meta`/`cb_kwargs` were dumped in python mode, so any model with a non-JSON-native field (`datetime`, `UUID`, `HttpUrl`, ...) failed `json.dumps` and the request was dropped. The v4 upgrade guide promises such models keep working.

- Dump models with `model_dump(mode='json', by_alias=True)`.
- Widen the serialization error guards to `except (TypeError, ValueError)` to cover pydantic's `PydanticSerializationError`.
- Add a round-trip test with a datetime-bearing model and update the v4 upgrade guide wording.

### DIFF
--- a/docs/04_upgrading/upgrading_to_v4.md
+++ b/docs/04_upgrading/upgrading_to_v4.md
@@ -218,7 +218,7 @@ Pickle could store arbitrary Python objects. JSON cannot, so the values in a req
 
 - A `tuple` comes back as a `list`.
 - Non-string `dict` keys come back as strings, so `{1: 'a'}` becomes `{'1': 'a'}`.
-- A value JSON cannot represent (`datetime`, `set`, `Decimal`, a custom object) is no longer stored silently. The request is skipped and the failure is logged. Pydantic models are still supported and are dumped with `model_dump(mode='json')`, so model fields JSON cannot natively represent (such as `datetime`) are stored in their JSON form.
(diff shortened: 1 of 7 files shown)


docs/04_upgrading/upgrading_to_v4.md
```diff
@@ -218,7 +218,7 @@ Pickle could store arbitrary Python objects. JSON cannot, so the values in a req
 
 - A `tuple` comes back as a `list`.
 - Non-string `dict` keys come back as strings, so `{1: 'a'}` becomes `{'1': 'a'}`.
-- A value JSON cannot represent (`datetime`, `set`, `Decimal`, a custom object) is no longer stored silently. The request is skipped and the failure is logged. Pydantic models are still supported and are dumped with `model_dump()`.
+- A value JSON cannot represent (`datetime`, `set`, `Decimal`, a custom object) is no longer stored silently. The request is skipped and the failure is logged. Pydantic models are still supported and are dumped with `model_dump(mode='json')`, so model fields JSON cannot natively represent (such as `datetime`) are stored in their JSON form.
```
Pijukatel: I am confused by the last sentence. It seems to me like the two parts are contradicting each other

1. `model fields JSON cannot natively represent`
2. `are stored in their JSON form`

How can I store something that can't be represented by JSON in its JSON form?

Maybe try some rewording to clear the potential confusion.

vdusek: Well, it means they are converted into JSON-compatible values. I'll improve the wording.

Pijukatel: Thanks, the new wording is making it clear.

Pijukatel (approved):

PR #962, opened 2026-06-11
refactor!: Remove wait_for_finish from Actor.start

vdusek:
## Description

- Remove the `wait_for_finish` argument from `Actor.start()`. It contradicted the method's purpose: `Actor.start()` only starts the run without waiting for it to finish. The JS SDK does not expose it either. To wait for a run to finish, use `Actor.call()`.
- Fix the `wait` docstrings of `Actor.call()` and `Actor.call_task()`, which described a `timedelta` as "maximum number of seconds".
- Document the removal in the v4 upgrading guide.


Pijukatel:
Hello, let's first align on the common understanding of the current code.

I am confused by the method doc string, the argument name and the discrepancy to the JS implementation.

Doc string of `Actor.start`

> Run an Actor on the Apify platform.
> 
> Unlike `Actor.call`, this method just starts the run without waiting for finish....


It explicitly states that there is no waiting for finish, but it has `wait_for_finish` argument

The JS implementation does not have this argument for `Actor.start`:
https://github.com/apify/apify-sdk-js/blob/master/src/actor.ts#L260

Maybe we should delete it here instead?

vdusek:
Ok, makes sense, updated.

Pijukatel (approved):

PR #964, opened 2026-06-11
fix: Preserve decorated symbol types in docs_group and docs_name

vdusek:
The `docs_group` and `docs_name` decorators in `apify._utils` were typed `Callable -> Callable`, which erased the types of every decorated public API symbol — `Configuration`, `ProxyConfiguration`, `Webhook`, the charging classes, event models, and `ApifyRequestList` all revealed as `Unknown`/`Any` to users' type checkers (`Actor` survived only thanks to an explicit cast). Both decorators are now identity-typed as `Callable[[T], T]`, reusing the module's existing bound `TypeVar` and matching how Crawlee fixed the same bug.

Also adds a regression test: a runtime identity check plus a type-level guard that relies on the `unused-ignore-comment = "error"` ty rule — if the decorators ever erase types again, the guard's suppressions become unused and `uv run poe type-check` fails.

### DIFF
--- a/tests/unit/test_utils.py
+++ b/tests/unit/test_utils.py
@@ -1,11 +1,12 @@
 from __future__ import annotations
 
 from enum import Enum
+from typing import TYPE_CHECKING
 from unittest.mock import patch
 
 import pytest
 
-from apify._utils import get_system_info, is_running_in_ipython, maybe_extract_enum_member_value
+from apify._utils import docs_group, docs_name, get_system_info, is_running_in_ipython, maybe_extract_enum_member_value
 
 
 def test_ipython_detection_when_active() -> None:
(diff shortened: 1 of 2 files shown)


tests/unit/test_utils.py
```diff
@@ -54,3 +55,30 @@ def test_maybe_extract_enum_member_value_with_non_enum() -> None:
...
+
+    @docs_group('Actor')
+    class _DocsGroupDecorated:
+        pass
+
+    @docs_name('Renamed')
+    class _DocsNameDecorated:
+        pass
+
+    _ = _DocsGroupDecorated.nonexistent  # ty: ignore[unresolved-attribute]
```
Pijukatel, line 73: This is an interesting trick.

Pijukatel (approved):

PR #965, opened 2026-06-11
refactor: consolidate empty-string env-var validators in Configuration

vdusek:
## Description

A few `Configuration` fields were already guarded against empty-string env vars with ad-hoc `BeforeValidator` lambdas. The Apify platform sometimes sets an env var to an empty string instead of leaving it unset (see #303, or this [Discord thread](https://discord.com/channels/801163717915574323/1020281045205123102/threads/1294774068528156817)), and for these fields `''` cannot be parsed into the target type, so without the guard validation would raise and crash `Actor.init()`.

This PR is a pure refactor: it consolidates those repeated lambdas into a single shared `_default_if_empty(default=...)` validator. There is no behavior change, and the empty-string handling is **not** extended to any new fields.

### Changes

- Added a shared `_default_if_empty(default=...)` validator.
- Replaced the existing ad-hoc lambdas with it on the four fields that already handled `''`:
  - `max_paid_dataset_items` (→ `None`)
  - `max_total_charge_usd` (→ `None`)
  - `timeout_at` (→ `None`)
  - `user_is_paying` (→ `False`)
- Added a parametrized regression test covering those four fields.


tests/unit/actor/test_configuration.py
```diff
@@ -392,3 +399,10 @@ def test_actor_storage_json_env_var(monkeypatch: pytest.MonkeyPatch) -> None:
     assert config.actor_storages['datasets'] == datasets
     assert config.actor_storages['request_queues'] == request_queues
     assert config.actor_storages['key_value_stores'] == key_value_stores
+
+
+def test_actor_storages_env_var_empty_string_becomes_none(monkeypatch: pytest.MonkeyPatch) -> None:
+    """Test that an empty env var for actor_storages is converted to None instead of crashing."""
+    monkeypatch.setenv('ACTOR_STORAGES_JSON', '')
```
Pijukatel: Isn't this actually less defensive?

Is there any actual use case for having 'ACTOR_STORAGES_JSON' be set as an empty string? The env variable clearly states it should be a json. So either have json there, or don't set the variable.

Code should not try to guess the intention from an ambiguous action. And here we are guessing that an empty env variable probably means the user did not want to set it up in the first place...

vdusek: It's more about robustness. In the current state, all fields that can't serialize an empty string (and would have failed before) now fall back to `None`. Previously, this fallback was only in place for some of the fields.

Since this is a robustness improvement rather than a bug fix, I changed the commit type to chore.

Let me know WDYT.

Pijukatel: I would ask this:
What is the valid use-case for those env variables to exist and be intentionally empty? 

Because I am afraid of this scenario:
Those env variables exist, but are **accidentally** empty.  
-> We should raise to let the user know that this is not a desired scenario.

vdusek: updated based on our Slack discussion

Pijukatel (approved):

PR #966, opened 2026-06-11
fix: accept arbitrary JSON userData in ApifyRequestList

vdusek:
### Description

- `_RequestDetails.user_data` was typed `dict[str, str]`, but the platform `requestListSources` input allows arbitrary JSON values in `userData`, so e.g. `{"depth": 1}` raised a `ValidationError`. Retyped to `dict[str, JsonSerializable]` (the same alias crawlee's `Request.from_url` declares).
- Fixed a related pre-existing crash: crawlee's `Request.from_url` mutates a non-empty `user_data` dict passed to it (writes `__crawlee` back into it), and `_process_remote_url` reused one dict for every URL extracted from a `requestsFromUrl` source. Any such source with non-empty `userData` and 2+ extracted URLs crashed with a `TypeError` on the second request. Each call now gets its own copy of the dict.
- Added regression tests for both cases.

### DIFF
--- a/src/apify/request_loaders/_apify_request_list.py
+++ b/src/apify/request_loaders/_apify_request_list.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import asyncio
 import re
+from copy import deepcopy
 from itertools import chain
 from typing import Annotated, Any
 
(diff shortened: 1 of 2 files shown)


src/apify/request_loaders/_apify_request_list.py
```diff
@@ -154,7 +154,9 @@ async def _process_remote_url(request_input: _RequestsFromUrlInput, http_client:
                 method=request_input.method,
                 payload=request_input.payload.encode('utf-8'),
                 headers=request_input.headers,
-                user_data=request_input.user_data,
+                # Copy the user data so that `Request.from_url` does not mutate the shared input dict,
+                # which would break creation of the subsequent requests.
+                user_data=dict(request_input.user_data),
```
Pijukatel: Should this be `deepcopy(request_input.user_data)`?

Pijukatel (approved):

PR #967, opened 2026-06-11
fix: reconnect to platform events websocket after connection drop

vdusek:
- Previously, `ApifyEventManager._process_platform_messages` connected to the platform events websocket exactly once. Any dropped connection raised `ConnectionClosedError` and permanently ended the processing task, while a graceful server-side close exited silently. In both cases, the Actor missed all subsequent platform events for the rest of the run.
- The connection now uses the `websockets` reconnect iterator, which re-establishes the connection after every drop or graceful close, with backoff on failed attempts. An abnormal drop is logged as a warning and a graceful close as info, both with the close code and reason, and a successful reconnect is logged as well.
- A `process_exception` callback keeps errors fatal before the first successful connection, so `Actor.init` still fails fast on a misconfigured URL instead of hanging. After the first connection, the default `websockets` classification decides which errors are transient and retried.
- `__aexit__` now cancels the processing task before closing the websocket. Otherwise, every clean shutdown would trigger a spurious reconnect attempt.
- Added a regression test, parametrized over graceful and abnormal closes, that drops the connection server-side and asserts the drop is logged, the client reconnects, and events keep arriving. It replaces the obsolete mid-stream disconnect test, whose premise (the task ends after a drop) no longer holds.


src/apify/events/_apify_event_manager.py
```diff
@@ -93,50 +108,121 @@ async def __aexit__(
...
+        *,
+        connection_lost: bool,
+    ) -> bool:
+        """Log the websocket close and report whether to reconnect (`False` on a non-retryable close code)."""
+        if websocket.close_code in self._NON_RETRYABLE_CLOSE_CODES:
+            logger.error(
+                f'Connection to platform events websocket was closed with a non-retryable code '
+                f'(code={websocket.close_code}, reason={websocket.close_reason!r}); not reconnecting.'
+            )
+            return False
```
Pijukatel, line 230: When a non-retryable close code is received, the loop `break`s and the processing task ends for the rest of the run — the Actor keeps running but will no longer receive **any** platform events. The most consequential are `MIGRATING` (which drives the `PERSIST_STATE` / `is_migrating=True` emit in `_handle_platform_message`) and `ABORTING`; after a permanent give-up the Actor would migrate without persisting state.

For `1008` the docstring notes two causes: an unknown/missing run ID (genuinely permanent — agreed) and an exceeded per-run connection limit (potentially transient). Is it intentional to permanently stop reconnecting in the connection-limit case, given the cost of silently losing migration handling? If the two can be told apart (e.g. via `close_reason`) it might be worth retrying the limit case — or at minimum the ERROR log could spell out that platform events (including migration) are now disabled for the remainder of the run, since that's a fairly significant degradation to surface only as "not reconnecting."

---
_Generated by [Claude Code](https://claude.ai/code)_

src/apify/events/_apify_event_manager.py
```diff
@@ -93,50 +108,121 @@ async def __aexit__(
...
 
+    def _process_connection_exception(self, exc: Exception) -> Exception | None:
+        """Decide whether a failed connection attempt to the platform websocket should be retried.
+
+        Before the first successful connection, every error is fatal so that `__aenter__` fails fast. After that,
+        the default `websockets` behavior decides which errors are transient and retried with exponential backoff.
+        """
+        if self._connected_to_platform_websocket and self._connected_to_platform_websocket.done():
+            return websockets.asyncio.client.process_exception(exc)
+        return exc
```
Pijukatel, line 140: Worth confirming this trade-off is intended. Before the first successful connection every exception is fatal, so a *transient* error at startup — connection refused / `OSError`, a `502`/`503`/`504`, or a connect timeout — fails `__aenter__` with `RuntimeError` and takes down `Actor.init()`, even though the identical error mid-run would be retried by the default classification.

I get the goal (avoid `__aenter__` hanging forever on indefinite retries, per the PR description) and a single attempt is bounded by `open_timeout`, so it won't hang. Since the reconnect machinery is now robust, a small bounded retry budget on the *initial* connect could avoid failing a whole run on a momentary blip at boot — but this also matches the prior fail-fast behavior, so keeping it as-is is reasonable if that's the intent. Mostly flagging so the choice is explicit.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/apify/events/_apify_event_manager.py
```diff
@@ -93,50 +108,121 @@ async def __aexit__(
...
+                if not self._should_reconnect_after_close(websocket, connection_lost=connection_lost):
+                    break
+
+                # Reconnect a healthy connection immediately; back off only on repeated rapid drops.
+                if time.monotonic() - connection_opened_at >= self._HEALTHY_CONNECTION_MIN_DURATION:
+                    backoff_delays = None
+                elif backoff_delays is None:
+                    backoff_delays = websockets.client.backoff()
+                else:
+                    await asyncio.sleep(next(backoff_delays))
```
Pijukatel: Minor / readability: the *first* rapid (`< MIN_DURATION`) drop hits the `elif backoff_delays is None` branch, which creates the generator but does not sleep — so the first rapid drop still reconnects immediately, and only the *second* consecutive rapid drop begins backing off (the first `next(backoff_delays)` value is consumed on that second drop, not skipped). That's a sensible "one free retry" policy and `# back off only on repeated rapid drops` hints at it, but it took a second read to follow. A one-line note like "first rapid drop reconnects once without delay; subsequent consecutive ones back off" would save the next reader the trace.

---
_Generated by [Claude Code](https://claude.ai/code)_

tests/unit/events/test_apify_event_manager.py
```diff
@@ -57,6 +70,74 @@ async def handler(websocket: websockets.asyncio.server.ServerConnection) -> None
...
+    Binds to a fixed free port (reserved up front) so a restart can reuse the same address, letting a test simulate the
+    platform server going away and coming back. Yields a control namespace with `live_clients`, a re-armable
+    `client_connected` event, a cumulative `attempts()` counter, and `stop()` / `start()` coroutines. Pass `on_connect`
+    to take over a freshly accepted connection (e.g. immediately close it with a chosen code).
+    """
+    # Reserve a fixed free port so a restart can re-serve on the same address.
+    probe = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    probe.bind(('127.0.0.1', 0))
+    port = probe.getsockname()[1]
+    probe.close()
```
Pijukatel: Nit (potential flakiness): the port is reserved by binding to `:0`, reading it, then closing the probe before `_serve()` binds it later. There's a TOCTOU window where another process can grab the port between `probe.close()` and `serve(...)`, and since the suite runs under `pytest-xdist` with multiple workers, the collision odds aren't entirely negligible. The same fixed port also has to stay free across every `stop()`/`start()` cycle.

Probably rare in practice, but holding the actual server socket open from the start (and reusing it across restarts), or wrapping `_serve()` in a small retry-on-`OSError`, would make it deterministic. Related: `start()`'s `await asyncio.sleep(0.3)` to "let the OS release the port" is a timing band-aid — `websockets.serve` sets `SO_REUSEADDR` by default on POSIX, so it may be unnecessary there (and doesn't behave the same on Windows).

---
_Generated by [Claude Code](https://claude.ai/code)_

Pijukatel (commented):
Some Claude spam...

Pijukatel (commented):
Is this implemented in JS? If not, is there a parity issue?

src/apify/events/_apify_event_manager.py
```diff
@@ -45,6 +49,17 @@ class ApifyEventManager(EventManager):
     with the event system.
     """
 
+    _NON_RETRYABLE_CLOSE_CODES = frozenset({1002, 1003, 1007, 1008, 1010})
```
Pijukatel: As a readability improvement, I would suggest saving each code in a variable/constant instead of text description in the doc block
```
protocol_error = 1002
unsupported_data = 1003
invalid_payload_data = 1007
policy_violation = 1008
mandatory_extension = 1010
...
```

src/apify/events/_apify_event_manager.py
vdusek: Kept the non-retryable behavior — `1008` stays terminal. It matches the JS SDK (which doesn't reconnect at all), and telling the connection-limit case apart from an unknown/missing run ID via `close_reason` is too fragile to branch on.

But I took the "at minimum" suggestion: the ERROR log now spells out the degradation explicitly instead of just "not reconnecting":

> No further platform events will be received for the rest of this run, so migration and abort handling (e.g. persisting state before a migration) is now disabled.



---
_Generated by [Claude Code](https://claude.ai/code)_

src/apify/events/_apify_event_manager.py
vdusek: Intentional — keeping the fail-fast. The asymmetry (retry mid-run, fail fast before the first connection) is deliberate:

- On the platform the events websocket is served by the same worker that launched the container, so it's already listening before `Actor.init()` runs. A connect failure at boot almost always means a real problem (bad URL / wrong run ID → `1008`), not a transient blip — and those wouldn't be fixed by retrying.
- A single attempt is already bounded by `open_timeout`, so there's no hang to guard against.
- Failing loudly at init is a safer degradation than a socket that limps up late and silently misses `MIGRATING`. The run-level failure is visible and recoverable.
- It also matches the prior behavior and the JS SDK.

Left the `_process_connection_exception` docstring as the explicit record of this choice.



---
_Generated by [Claude Code](https://claude.ai/code)_

src/apify/events/_apify_event_manager.py
vdusek: Good call — expanded the comment to spell out the "one free retry" policy:

> Reconnect a healthy connection immediately; back off only on repeated rapid drops. The first rapid drop reconnects once without delay (it only primes the backoff generator), and each subsequent consecutive rapid drop then sleeps for the next backoff interval. A healthy connection resets the generator, so the next rapid drop again gets that one free retry.



---
_Generated by [Claude Code](https://claude.ai/code)_

tests/unit/events/test_apify_event_manager.py
vdusek: Fixed. The probe-bind-then-close is gone. The fixture now binds the listening socket up front with `SO_REUSEADDR` and holds it open, handing it directly to `serve(sock=...)`; a restart rebinds the same port.

That closes the TOCTOU window (the port is never released for another process to grab), and because `SO_REUSEADDR` lets the restart rebind immediately without waiting out `TIME_WAIT`, I was also able to drop the `asyncio.sleep(0.3)` band-aid entirely. Verified across repeated parallel xdist runs.



---
_Generated by [Claude Code](https://claude.ai/code)_

vdusek:
> Is this implemented in JS? If not, is there a parity issue?

I'll open it.

Pijukatel (approved):

PR #976, opened 2026-06-12
fix: redirect input key in all file-system key-value store operations

vdusek:
Only `get_value` mapped the logical input key (`INPUT`) to the actual input file on disk (e.g. `INPUT.json`). `set_value`, `record_exists`, `delete_value`, and `get_public_url` operated on the raw key, so `set_value('INPUT', ...)` wrote a second input file and reopening the store raised `RuntimeError('Only one input file is allowed...')`, breaking `Actor.get_input()` in local dev.

The input-key redirect is now extracted into a shared `_resolve_input_key` helper and applied to every key-addressing record method, so they all target the single input file consistently.


Pijukatel (approved):
Please do some manual tests with CLI.

PR #978, opened 2026-06-12
fix(scrapy): Avoid mutating request userData during Scrapy-Apify conversion

vdusek:
`to_apify_request` serialized the Scrapy request **after** `Request.from_url()` had mutated the shared `meta['userData']` dict (same bug class as #832).

`Request.from_url()` injects a live `CrawleeRequestData` object under `__crawlee` into the very `user_data` dict it receives — which was the spider's own `meta['userData']`. Because `scrapy_request.to_dict()` ran afterward, two things went wrong:

- the spider's own request `meta['userData']` was mutated in place, and
- the serialized `scrapy_request` blob stored on the platform embedded redundant Crawlee internals in every request.

**Fix:** capture `scrapy_request.to_dict()` *before* calling `from_url()`, and pass `from_url()` a copy of `user_data` (`dict(user_data)` for the plain-dict branch; `model_dump()` already returns a fresh dict). The spider's request stays untouched and the stored blob is free of injected Crawlee data.

Added two regression tests covering both the no-mutation guarantee and the clean serialized blob.

### DIFF
--- a/src/apify/scrapy/requests.py
+++ b/src/apify/scrapy/requests.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from copy import deepcopy
 from logging import getLogger
 from typing import Any, cast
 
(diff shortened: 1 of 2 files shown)


src/apify/scrapy/requests.py
```diff
@@ -80,12 +80,21 @@ def to_apify_request(scrapy_request: ScrapyRequest, spider: Spider) -> ApifyRequ
...
         user_data = scrapy_request.meta.get('userData', {})
 
         # Convert UserData Pydantic model to a plain dict to prevent CrawleeRequestData objects from leaking
-        # into Request.from_url() during Scrapy-Apify roundtrips.
+        # into Request.from_url() during Scrapy-Apify roundtrips. `model_dump()` already returns a fresh dict; the
+        # plain-dict case is copied so the `pop` and `from_url()` mutations below never touch the spider's meta.
         if isinstance(user_data, UserData):
             user_data = user_data.model_dump(by_alias=True)
+        elif isinstance(user_data, dict):
+            user_data = dict(user_data)
```
Pijukatel: I am not sure how much the dict can mutate, but lets use `deepcopy` so that we do not have to think about it?

Pijukatel (approved):

PR #979, opened 2026-06-12
fix(scrapy): async-thread startup race, shutdown lifecycle, and timeout setting

vdusek:
## Description

Fixes several defects in the Scrapy integration's background event-loop thread (`AsyncThread`), the scheduler, and the HTTP cache storage, and makes the loop timeout configurable.

## Fixes

- **`run_coro` startup race** — the `is_running()` guard fired spuriously when a coroutine was submitted before the loop thread reached `run_forever()` (observed ~122/500 in `scheduler.open()`). It now guards on `is_closed()`. A coroutine queued on a not-yet-running loop runs once the loop starts; only a closed loop raises.
- **`close()` thread leak** — if task cancellation timed out or raised, the loop was never stopped or joined. Stop, join, and the forced-shutdown fallback now run in a `finally`, and the original error still propagates.
- **`close()` second call** — a repeated close raised `RuntimeError: Event loop is closed`. An `is_closed()` early-return makes it a no-op.
- **`close()` ignored its `timeout`** for the cancellation step (it used the constructor default). It now passes the caller's timeout through.
- **`run_coro` timeout** left the coroutine running. It now cancels the future on timeout.
- **HTTP cache open/cleanup thread leaks** — `open_spider` now closes the thread if opening the key-value store fails (matching `ApifyScheduler.open`). The expiration sweep runs inside `try` with `close()` in a `finally`.
- **Configurable timeout (#955)** — new `APIFY_ASYNC_THREAD_TIMEOUT_SECS` setting, wired into the scheduler (via `from_crawler`) and the cache storage.

## Error logging

The integration now follows consistent conventions for caught exceptions:

- **`except … as exc:` → `logger.warning(f'… {exc}')`, swallowed** — for *expected, recoverable* conditions handled locally: a malformed or legacy stored payload skipped as a cache/queue miss, or non-UTF-8 headers preserved in the serialized request. A short message plus the exception text, with no traceback, because it is not a bug.
- **`except Exception:` → `logger.exception('…')`, swallowed** — for *unexpected* failures handled at a terminal point: the cleanup sweep, shutdown, or skip-and-continue. `logger.exception` attaches the full traceback, and nothing re-raises because the error is handled here.
- **`except …:` → `raise` (no logging)** — when the error is re-raised and the caller or Scrapy logs it with a traceback anyway. `run_coro`'s timeout path cancels the future and re-raises without logging, so the failure is reported once.
- **`except Exception:` → `logger.exception('…'); raise`** — the boundary log, used only where local context materially helps *and* the propagated error would otherwise be logged only generically or not at all. The scheduler's `next_request` / `enqueue_request` / `has_pending_requests` are called synchronously by the Scrapy engine (not inside a Deferred), so without this log the Apify-specific context would be lost.

**Why `logger.exception` replaced `traceback.print_exc()`:** `traceback.print_exc()` writes a bare traceback straight to stderr, bypassing logging entirely. It has no level, no logger name, no message, and ignores Scrapy's and the SDK's log configuration and handlers. `logger.exception(msg)` logs at ERROR through the configured logging, so it is routed, formatted, and filterable like every other log line. It adds a message explaining *what* failed and still attaches the full traceback automatically, which makes including the exception object in the message (`{exc}`) redundant (ruff TRY401).

## Tests

New `tests/unit/scrapy/test_async_thread.py` covers the startup race, run-after-close, timeout cancellation, idempotent close, the caller timeout reaching the shutdown step, and stop/join when task cancellation fails. The scheduler and HTTP cache test modules gain coverage for the timeout setting, closing the thread on open failure, and the cleanup-failure path still closing the thread.

### DIFF
--- a/src/apify/scrapy/extensions/_httpcache.py
+++ b/src/apify/scrapy/extensions/_httpcache.py
@@ -4,6 +4,7 @@ import gzip
 import io
 import re
 import struct
+from datetime import timedelta
 from logging import getLogger
 from time import time
 from typing import TYPE_CHECKING
@@ -38,6 +39,8 @@ class ApifyCacheStorage:
         # Upper bound on how many keys the per-spider-close cleanup sweeps (best-effort; `close_spider`).
         self._expiration_max_items: int = settings.getint('APIFY_HTTPCACHE_EXPIRATION_MAX_ITEMS', 100)
         self._expiration_secs: int = settings.getint('HTTPCACHE_EXPIRATION_SECS')
+        # Caps how long each coroutine run on the background event loop may take; defaults to 60 seconds.
+        self._async_thread_timeout = timedelta(seconds=settings.getint('APIFY_ASYNC_THREAD_TIMEOUT_SECS', 60))
         self._spider: Spider | None = None
         self._kvs: KeyValueStore | None = None
         self._fingerprinter: RequestFingerprinterProtocol | None = None
@@ -62,7 +65,7 @@ class ApifyCacheStorage:
             return await KeyValueStore.open(name=kvs_name)
 
         logger.debug("Starting background thread for cache storage's event loop")
-        self._async_thread = AsyncThread()
+        self._async_thread = AsyncThread(default_timeout=self._async_thread_timeout)
         logger.debug(f"Opening cache storage's {kvs_name!r} key value store")
         self._kvs = self._async_thread.run_coro(open_kvs())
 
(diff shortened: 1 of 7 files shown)


src/apify/scrapy/extensions/_httpcache.py
```diff
@@ -62,55 +65,74 @@ async def open_kvs() -> KeyValueStore:
...
-                        gzip_time = read_gzip_time(value)
-                    except Exception as e:
-                        logger.warning(f'Malformed cache item {item.key}: {e}')
-                        await self._kvs.delete_value(item.key)
-                    else:
-                        if self._expiration_secs < current_time - gzip_time:
-                            logger.debug(f'Expired cache item {item.key}')
+        # `close` always runs in the `finally`, so neither a cleanup failure below nor an early return can leak
+        # the event-loop thread.
+        try:
```
Pijukatel, line 124: There is too much happening in this try, so it masks which is the part that can raise.

I would suggest to:
1) move arg handling out
```
if current_time is None:
    current_time = int(time()
```
2) define `expire_kvs()` as private function on `ApifyCacheStorage`


Then the `try` is more obvious, and we will see, that we have nested try like this

```
        try:
            try:
                self._async_thread.run_coro(self._expire_kvs(...))
            except Exception:
                logger.exception('Failed to clean up expired cache items.')
        finally:
            logger.debug('Closing cache storage')
            try:
                self._async_thread.close()
            except KeyboardInterrupt:
                logger.warning('Shutdown interrupted by KeyboardInterrupt!')
            except Exception:
                logger.exception('Exception occurred while shutting down cache storage')
            finally:
                logger.debug('Cache storage closed')
```
which I think can be simplified by removing one try.

src/apify/scrapy/extensions/_httpcache.py
```diff
@@ -122,7 +144,11 @@ def retrieve_response(self, _: Spider, request: Request, current_time: int | Non
             raise ValueError('Request fingerprinter not initialized')
 
         key = self._fingerprinter.fingerprint(request).hex()
-        value = self._async_thread.run_coro(self._kvs.get_value(key))
+        try:
+            value = self._async_thread.run_coro(self._kvs.get_value(key))
+        except Exception:
```
Pijukatel, line 167: This pattern is really hard to understand without more context. I think this deserves an inline comment, as the natural thing to do with this is to remove the `try/except` as it seems redundant.

src/apify/scrapy/extensions/_httpcache.py
vdusek: Yeah, good idea

src/apify/scrapy/extensions/_httpcache.py
vdusek: It is used everywhere. IIRC, there were some issues with the exception propagation from a separate thread and its event loop. This was extreme pain to make it work, so I definitely wouldn't change it.

Pijukatel: Swallowing all exceptions to add additional log and re-raise the exception is an intentional non-standard action, which is here due to some external context circumstances. Those external circumstances are not obvious on this level, so that is why I am asking for some comment, so that the reader without context can understand the intention of the code.

I am not asking to change this, but to help protect it by adding a comment explaining it.

vdusek: Ah, okay, I just thought you wanted an explanation, sorry, adding it.

Pijukatel (approved):

PR #980, opened 2026-06-12
fix(scrapy): async-thread shutdown, duplicate error logs, and timeout setting

vdusek:
Fixes several smaller defects in the Scrapy integration's background event-loop thread, scheduler, and HTTP cache storage.

- **HTTP cache cleanup could leak the event-loop thread.** In `ApifyCacheStorage.close_spider`, the expiration sweep ran outside the `try`, so a failure there skipped `AsyncThread.close()`. The sweep now runs inside `try`, with `close()` in a `finally` so it always runs.
- **`AsyncThread.close()` raised on a second call.** A repeated close (e.g. `ApifyScheduler.open()` closing on failure, then Scrapy closing again) called into the already-closed loop and raised `RuntimeError: Event loop is closed`. Added an `is_closed()` early-return guard so a second close is a no-op.
- **`AsyncThread.close()` ignored its `timeout` argument** for the task-cancellation step (it used the constructor default). It now passes the caller's `timeout` to that step too.
- **`run_coro` timeout left the coroutine running.** On timeout it now cancels the future so the coroutine does not outlive the timeout.
- **Scheduler errors were reported 2–3×.** `run_coro` logged-and-raised and each scheduler call added `traceback.print_exc()` (which bypasses `ActorLogFormatter`), on top of Scrapy's own logging. Removed the redundant `traceback.print_exc()` wrappers and `run_coro`'s log-and-raise; errors now surface once, formatted, via Scrapy.
- **#955's configurable timeout was unreachable by users.** Added the `APIFY_ASYNC_THREAD_TIMEOUT_SECS` Scrapy setting, wired into the scheduler (via `from_crawler`) and the cache storage so the `AsyncThread` timeout is actually configurable.

Adds unit tests for each fix (new `tests/unit/scrapy/test_async_thread.py` plus additions to the scheduler and HTTP cache test modules).


vdusek:
Consolidated into #979, which now carries these lifecycle fixes on top of the startup-race fix. Closing in favor of #979.

PR #982, opened 2026-06-12
docs: Fix docstring typos, add missing docs groups, and document platform purge()

vdusek:
Documentation-only portion of the v4 major-release cleanup.

- Fix docstring typos and stale claims in `Actor`: `MIIGRATING` → `MIGRATING`, drop the nonexistent `call_task` `content_type` argument, drop the "in seconds" note on `timeout` (the parameter is a `timedelta`), correct the `exit_process` default note, and fix the `task_id` description.
- Fix `Configuration` docstrings: "prising" → "pricing", and the concatenated `web_server_port` description.
- Replace mkdocs-style cross-references in `events/_types.py` with plain backticks.
- Add `@docs_group` plus a docstring to `ApifyRequestQueueMetadata`, and `@docs_group` to `ApifyFileSystemStorageClient`.
- Document in the v4 upgrading guide that platform storages raise `NotImplementedError` on `purge()` (use `drop()` instead). Local file-system storages are unaffected.

### DIFF
--- a/docs/04_upgrading/upgrading_to_v4.md
+++ b/docs/04_upgrading/upgrading_to_v4.md
@@ -83,6 +83,20 @@ run = await Actor.start('my-actor-id', wait_for_finish=60)
 run = await Actor.call('my-actor-id', wait=timedelta(seconds=60))
 ```
 
+## Purging storages on the Apify platform
+
+The Apify platform storage clients do not support `purge()`. Calling `purge()` on a dataset, key-value store, or request queue opened on the platform raises `NotImplementedError`, because the platform API has no operation that clears a storage's contents in place. To delete a storage, call `drop()` instead.
+
+```python
+# Raises NotImplementedError on the Apify platform.
+await dataset.purge()
+
+# Delete the storage instead.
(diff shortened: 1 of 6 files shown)


docs/04_upgrading/upgrading_to_v4.md
```diff
@@ -83,6 +83,20 @@ run = await Actor.start('my-actor-id', wait_for_finish=60)
 run = await Actor.call('my-actor-id', wait=timedelta(seconds=60))
 ```
 
+## Purging storages on the Apify platform
```
Pijukatel: Was this changed in v4, or was it always like this?

Pijukatel (approved):

docs/04_upgrading/upgrading_to_v4.md
vdusek: Oh, good catch. I don't know why I thought we had worked around this by having `purge` call `drop`, but it's not true. It has been raising an implementation error since 3.0.0, when it was added. I'm deleting this section.

PR #985, opened 2026-06-13
refactor: Inject API client into AliasResolver instead of creating its own

vdusek:
`AliasResolver` (and `open_by_alias`) no longer create their own `ApifyClientAsync` to read the alias mapping from the default key-value store. The API client already built for the storage operation is injected and reused, so alias resolution shares its HTTP client (connection pool and event-loop affinity) instead of spinning up — and leaking — a separate client on every call. This also removes the need for the process-global client cache that the review flagged.


tests/unit/storage_clients/test_alias_resolver.py
```diff
@@ -20,17 +28,20 @@ def test_storage_key_format() -> None:
 
 async def test_resolve_id_returns_none_for_unknown() -> None:
     """Test that resolve_id returns None for an alias not in the map."""
-    AliasResolver._alias_map = {}
```
Pijukatel, line 23: Isn't this reducing the test isolation? `AliasResolver._alias_map` could be populated by previous tests?

vdusek: We have a `prepare_test_env` fixture. So this was redundant.

Pijukatel (approved):

PR #1000, opened 2026-06-17
test: deflake request queue aliases e2e test

vdusek:
`test_rq_aliases` intermittently failed after the Actor reboot. The platform's batch-add endpoint is best-effort and can return a request as unprocessed; the singular `add_request` does not retry it (unlike `add_requests`), so the request was silently dropped and the post-reboot `fetch_next_request` returned `None`, failing the in-Actor assertion and the run.

This switches the two pre-reboot adds to `add_requests`, which retries unprocessed requests, so the requests reliably survive the reboot. Test-only change, mirroring the sibling deflake in #991.

### DIFF
--- a/tests/e2e/test_actor_request_queue.py
+++ b/tests/e2e/test_actor_request_queue.py
@@ -109,8 +109,11 @@ async def test_rq_aliases(
             assert rq_2.name is None
 
             if not was_rebooted:
-                await rq_1.add_request(Request(url='https://example.com/rq_1', unique_key='rq_1'))
-                await rq_2.add_request(Request(url='https://example.com/rq_2', unique_key='rq_2'))
+                # The platform's batch-add endpoint is best-effort: a single `add_request` can come back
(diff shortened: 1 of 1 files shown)


vdusek:
If this is the intended behavior, we should just stabilize the test. If not, we should consider updating the `add_request` method (Apify storage client).

tests/e2e/test_actor_request_queue.py
```diff
@@ -109,8 +109,11 @@ async def main() -> None:
             assert rq_2.name is None
 
             if not was_rebooted:
-                await rq_1.add_request(Request(url='https://example.com/rq_1', unique_key='rq_1'))
```
Pijukatel, line 112: Maybe worth adding retries to the `add_request` to be consistent with `add_requests`

vdusek:
Closing in favor of https://github.com/apify/crawlee-python/pull/1976

PR #1008, opened 2026-06-22
fix: prevent request queue softlock by adding `is_finished` and correcting `is_empty`

### LINKED ISSUE #987: Implement the `is_finished` method in the `ApifyStorageClient`
- Depends on https://github.com/apify/crawlee-python/issues/1966

Mantisus:
### Description

Add `is_finished` and align `is_empty` semantics in `ApifyRequestQueueClient`

Closes: #987
Follow-up: https://github.com/apify/crawlee-python/pull/1982

### DIFF
--- a/src/apify/storage_clients/_apify/_request_queue_shared_client.py
+++ b/src/apify/storage_clients/_apify/_request_queue_shared_client.py
@@ -292,15 +292,17 @@ class ApifyRequestQueueSharedClient:
         # Check _list_head.
         # Without the lock the `is_empty` is prone to falsely report True with some low probability race condition.
         async with self._fetch_lock:
-            head = await self._list_head(limit=1)
-            return len(head.items) == 0
+            return await self._is_empty()
 
     async def is_finished(self) -> bool:
         """Specific implementation of this method for the RQ shared access mode."""
-        if not await self.is_empty():
-            return False
+        async with self._fetch_lock:
+            return await self._is_empty() and not self._queue_has_locked_requests
 
-        return not self._queue_has_locked_requests
+    async def _is_empty(self) -> bool:
+        """Check whether anything is available to fetch. Lock-free core of `is_empty`, caller must hold the lock."""
+        head = await self._list_head(limit=1)
+        return len(head.items) == 0
 
     async def _get_metadata_estimate(self) -> RequestQueueMetadata:
         """Try to get cached metadata first. If multiple clients, fuse with global metadata.
--- a/src/apify/storage_clients/_apify/_request_queue_shared_client.py
+++ b/src/apify/storage_clients/_apify/_request_queue_shared_client.py
@@ -293,7 +293,14 @@ class ApifyRequestQueueSharedClient:
         # Without the lock the `is_empty` is prone to falsely report True with some low probability race condition.
         async with self._fetch_lock:
             head = await self._list_head(limit=1)
-            return len(head.items) == 0 and not self._queue_has_locked_requests
+            return len(head.items) == 0
+
+    async def is_finished(self) -> bool:
+        """Specific implementation of this method for the RQ shared access mode."""
+        if not await self.is_empty():
+            return False
+
+        return not self._queue_has_locked_requests
 
     async def _get_metadata_estimate(self) -> RequestQueueMetadata:
         """Try to get cached metadata first. If multiple clients, fuse with global metadata.
--- a/src/apify/storage_clients/_apify/_request_queue_shared_client.py
+++ b/src/apify/storage_clients/_apify/_request_queue_shared_client.py
@@ -8,6 +8,7 @@ from typing import TYPE_CHECKING, Any, Final
 
 from cachetools import LRUCache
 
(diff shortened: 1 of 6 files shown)


src/apify/storage_clients/_apify/_request_queue_shared_client.py
```diff
@@ -293,7 +293,14 @@ async def is_empty(self) -> bool:
...
             head = await self._list_head(limit=1)
-            return len(head.items) == 0 and not self._queue_has_locked_requests
+            return len(head.items) == 0
+
+    async def is_finished(self) -> bool:
+        """Specific implementation of this method for the RQ shared access mode."""
+        if not await self.is_empty():
+            return False
+
+        return not self._queue_has_locked_requests
```
vdusek: The old `is_empty` checked `not self._queue_has_locked_requests` while still inside the `_fetch_lock` block, immediately after `_list_head` updated it. With this change, `is_finished` reads it after `is_empty` has released the lock (so a concurrent `_list_head` could modify the value in between, I guess).

vdusek (changes requested):
One comment

src/apify/storage_clients/_apify/_request_queue_shared_client.py
```diff
@@ -292,8 +292,17 @@ async def is_empty(self) -> bool:
...
         # Without the lock the `is_empty` is prone to falsely report True with some low probability race condition.
         async with self._fetch_lock:
-            head = await self._list_head(limit=1)
-            return len(head.items) == 0 and not self._queue_has_locked_requests
+            return await self._is_empty()
+
+    async def is_finished(self) -> bool:
+        """Specific implementation of this method for the RQ shared access mode."""
+        async with self._fetch_lock:
+            return await self._is_empty() and not self._queue_has_locked_requests
```
vdusek, line 301: This relies on `await self._is_empty()` is being evaluated first (it calls `_list_head`, which populates `self._queue_has_locked_requests`). A reorder would break the finished detection. Maybe we could add a comment noting the ordering dependency.

src/apify/storage_clients/_apify/_request_queue_single_client.py
```diff
@@ -279,7 +279,11 @@ async def is_empty(self) -> bool:
         """Specific implementation of this method for the RQ single access mode."""
         # Without the lock the `is_empty` is prone to falsely report True with some low probability race condition.
```
vdusek: This comment is now outdated and misleading. Could you update it?

tests/integration/test_request_queue.py
```diff
@@ -651,6 +651,44 @@ async def test_empty_rq_behavior(request_queue_apify: RequestQueue, rq_poll_time
     assert metadata.pending_request_count == 0, f'metadata.pending_request_count={metadata.pending_request_count}'
 
 
+async def test_is_empty_and_is_finished(request_queue_apify: RequestQueue, rq_poll_timeout: int) -> None:
```
vdusek: For the most part, this duplicates the existing `test_request_queue_is_finished` test. Could we consolidate them into one? And parametrized it `single`+`shared`.

vdusek (commented):
Three more comments

vdusek (commented):
Could we also use a fix here please? Same reason as explained here https://github.com/apify/crawlee-python/pull/1982#pullrequestreview-4571758641

Claude suggests one of these:

- fix: prevent RQ softlock by correcting is_empty in ApifyRequestQueueClient
- fix: add is_finished and align is_empty semantics in ApifyRequestQueueClient

I am ok with both of them

src/apify/storage_clients/_apify/_request_queue_single_client.py
Mantisus: Is this an outdated comment? The behavior remains the same: without a lock, `is_empty` and `is_finished` have a small chance of returning a false `True`.

tests/integration/test_request_queue.py
Mantisus: Updated. This has already been parameterized with `single` + `shared`.

src/apify/storage_clients/_apify/_request_queue_single_client.py
vdusek: There is no lock in the single client, right? Only in the shared. I would remove it completely.

vdusek (commented):
Let's get back to this, once Crawlee 1.8.0 is released.

vdusek (commented):
Do I understand correctly that this requires a new version of Crawlee?

If yes, then this would be a breaking change on our side, and we'd need to bump the lower version boundary. Is there a way to make this non-breaking so we don't have to bump it?

Mantisus:
> If yes, then this would be a breaking change on our side, and we'd need to bump the lower version boundary. Is there a way to make this non-breaking so we don't have to bump it?

I added a small workaround. This will allow us to maintain compatibility until we're ready to bump the Crawlee version.

vdusek (approved):
LGTM

PR #1046, opened 2026-07-15
refactor!: Export Apify storage clients under their real names

vdusek:
Cleans up inconsistencies in the public export surface of `apify.storage_clients`.

**BREAKING CHANGE:** `apify.storage_clients.FileSystemStorageClient` now refers to Crawlee's `FileSystemStorageClient`, not the Apify variant. Previously the name was an alias for the Apify file-system client, which shadowed Crawlee's distinct class of the same name. Use `ApifyFileSystemStorageClient` for the Apify variant (the default local client).

- Export `ApifyFileSystemStorageClient` from `apify.storage_clients` under its real name, and re-export Crawlee's `FileSystemStorageClient` alongside it, so both distinct classes are available without shadowing. `SmartApifyStorageClient`, the tests, and the docs refer to the Apify variant by name.
- Export the previously-missing `ApifyFileSystemDatasetClient` from `apify.storage_clients._file_system`, so the sub-package now re-exports all three of its client classes.
- Document that `MemoryStorageClient` has no Apify variant, so it's re-exported from Crawlee as-is and lacks the Apify INPUT-preservation-on-purge and `PAY_PER_EVENT` charging behavior. Clarify the `ApifyFileSystemStorageClient` docstring (SDK variant with input handling and local PPE specifics).
- Fix a stale `Configuration` docstring that referenced the removed `actor.config` property (now `Actor.configuration`).


vdusek:
Closing, as we decided not to do that.

PR #1051, opened 2026-07-16
fix: Limit timeout to minimum of 1 seconds when timeout='inherit'

Pijukatel:
Limit the minimum possible timeout to 1 second when using timeout='inherit'. 1s is the smallest possible API timeout value and will result in an almost certain timeout of the called actor.

Previously, the timeout value got set to 0, but the API interprets a timeout of 0 as no timeout at all.

### DIFF
--- a/src/apify/_actor.py
+++ b/src/apify/_actor.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import asyncio
+import math
 import sys
 import warnings
 from contextlib import suppress
@@ -896,6 +897,39 @@ class _ActorType:
         env_vars = {env_var.value.lower(): env_var.name.lower() for env_var in [*ActorEnvVars, *ApifyEnvVars]}
         return {option_name: config[env_var] for env_var, option_name in env_vars.items() if env_var in config}
 
+    @overload
+    async def start(
+        self,
+        actor_id: str,
+        run_input: Any = None,
+        *,
+        token: str | None = None,
+        content_type: str | None = None,
+        build: str | None = None,
+        max_total_charge_usd: Decimal | None = None,
+        restart_on_error: bool | None = None,
+        memory_mbytes: int | None = None,
+        timeout: timedelta | None = None,
+        force_permission_level: ActorPermissionLevel | None = None,
+        webhooks: list[Webhook] | None = None,
+    ) -> Run: ...
+    @overload
+    async def start(
+        self,
+        actor_id: str,
+        run_input: Any = None,
+        *,
+        token: str | None = None,
+        content_type: str | None = None,
+        build: str | None = None,
+        max_total_charge_usd: Decimal | None = None,
+        restart_on_error: bool | None = None,
+        memory_mbytes: int | None = None,
+        timeout: Literal['inherit'],
+        force_permission_level: ActorPermissionLevel | None = None,
+        webhooks: list[Webhook] | None = None,
+    ) -> Run | None: ...
+
     @_ensure_context
     async def start(
         self,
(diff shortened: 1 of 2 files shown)


tests/unit/actor/test_actor_helpers.py
```diff
@@ -339,6 +339,50 @@ async def test_get_remaining_time_returns_positive_when_timeout_in_future() -> N
         assert result <= timedelta(minutes=5)
 
 
+async def test_get_remaining_time_rounds_up_to_whole_seconds() -> None:
+    """Test that _get_remaining_time rounds a fractional remaining time up, so it is never truncated to zero."""
+    async with Actor:
+        Actor.configuration.is_at_home = True
+        Actor.configuration.timeout_at = datetime.now(tz=UTC) + timedelta(seconds=30.5)
```
vdusek: Relying on elapsed time - a potential source of flakiness. Isn't there a more robust way to test this?

src/apify/_actor.py
```diff
@@ -931,20 +965,28 @@ async def start(
...
         Returns:
-            Info about the started Actor run
+            Info about the started Actor run, or `None` if the start was skipped because `timeout='inherit'`
+            was used and this Actor run has no time remaining before its own timeout.
         """
         client = self.new_client(token=token) if token else self.apify_client
 
         if timeout == 'inherit':
             actor_start_timeout = self._get_remaining_time()
+            if actor_start_timeout == timedelta(0):
```
vdusek: This timeout resolution logic is repeated in several places. Maybe we could extract it into a helper method?

---

Suggestion from Claude:

This whole `timeout` resolution — the `inherit` branch (`_get_remaining_time()` + the skip warning + `return None`) *and* the `elif None` / `elif timedelta` / `else: raise ValueError(...)` tail — is copy-pasted across `start` (this block), `call`, and `call_task`, differing only in the variable name and the method name in the warning.

It can collapse to ~3 lines per method by moving the entire decision into one helper and signalling "skip" with a typed sentinel (an `enum` member narrows cleanly under `ty`, so the caller stays type-safe):

```python
class _SkipRun(enum.Enum):
    """Sentinel returned by `_ActorType._resolve_run_timeout` when the run should not be started."""

    TOKEN = enum.auto()


_SKIP_RUN = _SkipRun.TOKEN
```

```python
def _resolve_run_timeout(
    self,
    timeout: timedelta | None | Literal['inherit'],
    *,
    operation: str,
) -> timedelta | None | Literal[_SkipRun.TOKEN]:
    """Resolve the run timeout for a start/call operation.

    Returns `_SKIP_RUN` if `timeout='inherit'` was used but the current run has no time remaining, in which
    case the caller should skip the operation and return `None`.
    """
    if timeout == 'inherit':
        remaining = self._get_remaining_time()
        if remaining == timedelta(0):
            self.log.warning(
                f"{operation} was skipped: `timeout='inherit'` was used, but the current run "
                'has no time remaining before its own timeout.'
            )
            return _SKIP_RUN
        return remaining
    if timeout is None:
        return None
    if isinstance(timeout, timedelta):
        return timeout
    raise ValueError(f'Invalid timeout {timeout!r}: expected `None`, `"inherit"`, or a `timedelta`.')
```

Each of the three methods then becomes just:

```python
actor_start_timeout = self._resolve_run_timeout(timeout, operation='Actor.start()')
if actor_start_timeout is _SKIP_RUN:
    return None
```

That puts the whole timeout decision (validation, `inherit` handling, skip warning, error message) in a single place instead of three. I verified this refactor locally: `ty` type-check, `ruff`, and the timeout/inherit tests all pass. Non-blocking.

tests/unit/actor/test_actor_helpers.py
Pijukatel: Replaced

src/apify/_actor.py
Pijukatel: Trying a different approach. No longer relevant.

Pijukatel:
A 1-second timeout might be more practical, as it retains the type and behavior consistency while having a kind of the same end-user behavior.

vdusek (approved):
lgtm

PR #1059, opened 2026-07-21
fix: prevent duplicate platform websocket when nesting ApifyEventManager

vdusek:
`ApifyEventManager` overrode Crawlee's `EventManager` async context manager without respecting the parent's ref counting. Crawlee ref-counts nested `async with` blocks, and `BasicCrawler._run_crawler` always re-enters the same global event manager that `async with Actor:` already entered. Because the override started the platform websocket machinery on every enter and tore it down on every exit, the standard `async with Actor:` + `await crawler.run()` pattern opened a **second** websocket connection.

On the platform this meant every platform event (`MIGRATING`, `ABORTING`, `PERSIST_STATE`) was delivered twice during a crawl, and on the inner exit the fields pointed at the second connection, so the first leaked and could never be cancelled. Each sequential `crawler.run()` added another connection; the platform caps at 10 per run and then closes with `1008` (a non-retryable code), permanently disabling platform events - so migration/abort handling silently stops for the rest of the run.

The fix mirrors the parent's ref counting: the websocket machinery is started only on the outermost enter (`0 -> 1`) and torn down only on the outermost exit (`1 -> 0`). Nested enters/exits reuse the single existing connection.

Added two regression tests: a nested enter reuses the one connection and delivers each event exactly once; a nested exit leaves the outer context's connection working while the final exit tears everything down with no leaked task.

*✍️ Drafted by Claude Code*

### DIFF
--- a/src/apify/events/_apify_event_manager.py
+++ b/src/apify/events/_apify_event_manager.py
@@ -95,6 +95,13 @@ class ApifyEventManager(EventManager):
     @override
     async def __aenter__(self) -> Self:
         await super().__aenter__()
+
+        # The parent ref-counts nested contexts (e.g. a crawler entering the same event manager). Only the
+        # outermost enter (0 -> 1) may start the platform websocket machinery; a nested enter must reuse the
+        # existing connection rather than open a second one.
+        if self._active_ref_count > 1:
+            return self
+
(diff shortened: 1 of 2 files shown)


vdusek:
Closing as a duplicate of https://github.com/apify/apify-sdk-python/pull/1082

PR #1061, opened 2026-07-21
test: add regression test for Actor.exit() deadlock from event listeners

vdusek:
`Actor.exit()` / `Actor.fail()` called from within an event listener used to deadlock into a `RecursionError`, since the cleanup path waited on the very listener task that called it.

Fixed upstream in crawlee's `EventManager`: [apify/crawlee-python#2088](https://github.com/apify/crawlee-python/pull/2088) (merged). On Python 3.11 it still deadlocks, because `asyncio.wait_for` there wraps the awaited coroutine in a separate task, which defeats crawlee's self-wait detection. This is a minor edge case, so we're not adding an SDK-side workaround just for Python 3.11.

This PR:
- Adds a regression test for `Actor.exit()` called from an `ABORTING` listener, skipped on Python 3.11.
- Bumps `uv.lock` (not the declared `pyproject.toml` constraint) to `crawlee==1.9.1b4`, so CI exercises the fix.

Follow-up: bump the declared `crawlee` constraint in `pyproject.toml` once a stable release with the fix ships.

*✍️ Drafted by Claude Code*

### DIFF
--- a/src/apify/_actor.py
+++ b/src/apify/_actor.py
@@ -4,6 +4,7 @@ import asyncio
 import math
 import sys
 import warnings
+from contextlib import contextmanager
 from dataclasses import asdict
 from datetime import UTC, datetime, timedelta
 from functools import cached_property
@@ -41,7 +42,7 @@ from apify.storages import Dataset, KeyValueStore, RequestQueue
 
 if TYPE_CHECKING:
     import logging
-    from collections.abc import Callable, MutableMapping
+    from collections.abc import Callable, Iterator, MutableMapping
     from decimal import Decimal
     from types import TracebackType
     from typing import Self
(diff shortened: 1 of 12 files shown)


src/apify/_actor.py
```diff
@@ -1488,6 +1492,28 @@ def _get_remaining_time(self) -> timedelta | None:
         )
         return None
 
+    @contextmanager
```
Pijukatel: Taking something out for a while and then returning it back feels like we should not have taken it out in the first place.  This kind of action seems very suspicious in async code, and I am worried about some potential race conditions.
 
Maybe take one step back and try to rethink this instead of fixing this one scenario.

vdusek: Good point, maybe we can resolve this on the `crawlee-python` level

vdusek:
maybe we can close this in favor of https://github.com/apify/crawlee-python/pull/2088 , let's see

vdusek:
Merge it with Crawlee 1.9.1 in the lockfile

Mantisus (approved):
LGTM

B4nan (approved):

PR #1062, opened 2026-07-21
fix: prolong and track shared request queue locks to prevent duplicate processing

vdusek:
The shared request-queue client (`access='shared'`) locked a batch of up to 25 requests for 180 s in `list_and_lock_head`, but never tracked, prolonged, or released those locks. As a batch drained slowly, locks expired mid-flight and requests got processed twice - by another consumer that re-locked an expired request, or by a second worker in this process after the request was re-listed. That defeats the class's documented "request locking to prevent duplicate processing" guarantee (double charges in PPE, duplicate dataset items).

Changes:
- Store `lock_expires_at` per cached head entry (the API returned it, we discarded it).
- Track handed-out requests in `_requests_in_progress`: skip them when re-listing the head and when fetching, and keep `is_finished` false while any is outstanding.
- In `fetch_next_request`, skip candidates whose lock already lapsed, and prolong the lock only when its remaining window is below half the lock time (`_MIN_LOCK_TIME_AT_HANDOFF`, 90 s). A request handed out from a fresh batch lock needs no prolong, so the common path adds no API write and shared mode stays at 3 writes per request.

Out of scope: handlers outliving the lock window would need a background refresh loop, and deleting leftover head locks on shutdown has no `RequestQueueClient` hook to attach to (they lapse within the lock TTL anyway).

*✍️ Drafted by Claude Code*

### DIFF
--- a/src/apify/storage_clients/_apify/_request_queue_shared_client.py
+++ b/src/apify/storage_clients/_apify/_request_queue_shared_client.py
@@ -74,11 +74,7 @@ class ApifyRequestQueueSharedClient:
         """Local cache of request IDs from the request queue head for efficient fetching."""
 
         self._requests_in_progress = set[str]()
-        """Set of request IDs currently handed to a consumer of this client and not yet handled or reclaimed.
-
-        Tracked locally so a request is never handed to a second consumer in this process, even if its platform
-        lock lapses and the queue head re-lists it.
-        """
+        """Request IDs handed to a consumer and not yet handled or reclaimed, tracked to avoid double-handing."""
 
         self._requests_cache: LRUCache[str, CachedRequest] = LRUCache(maxsize=cache_size)
         """LRU cache storing request objects, keyed by request ID."""
--- a/src/apify/storage_clients/_apify/_request_queue_shared_client.py
+++ b/src/apify/storage_clients/_apify/_request_queue_shared_client.py
@@ -267,8 +267,7 @@ class ApifyRequestQueueSharedClient:
             self._requests_in_progress.discard(next_request_id)
             return None
 
-        # A `None` response means the lock was not (re)acquired, so another consumer may hold it. Skip the request
-        # rather than hand out one whose lock we do not hold.
+        # `None` means the lock was not (re)acquired; skip the request rather than process it without a lock.
         if lock_info is None:
             logger.debug(f'Lock of request {next_request_id} could not be re-acquired, skipping it')
             self._requests_in_progress.discard(next_request_id)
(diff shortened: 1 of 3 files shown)


vdusek:
Btw. this is a different case than https://github.com/apify/apify-sdk-python/pull/887

src/apify/storage_clients/_apify/_request_queue_shared_client.py
```diff
@@ -219,12 +222,55 @@ async def fetch_next_request(self) -> Request | None:
...
-            if not self._queue_head:
+            if next_request_id is None:
                 return None
 
-            # Get the next request ID from the request queue head
-            next_request_id = self._queue_head.popleft()
+        # Prolong the lock so the consumer gets a full lock window starting now instead of sharing the batch lock
+        # acquired when the head was listed. If the lock can no longer be (re)acquired, skip the request.
+        try:
+            lock_info = await self._api_client.prolong_request_lock(
```
Pijukatel: What about the cost increase? Won't this lead to each request being locked at least twice? Maybe we could lock only if the remaining time is something very small?

I will run benchmarks with this change to quantify the cost increase. Maybe it is not serious

Pijukatel: Let's wait for them to finish:
Without change: https://github.com/apify/actor-benchmarks/actions/runs/29900513281
With change: https://github.com/apify/actor-benchmarks/actions/runs/29900596790

Pijukatel: Summary for the repeated 10 benchmark runs
Without change: 

> Overall benchmark of all runs, :aggregated_benchmark=Actor: parsel-crawler-benchmark, Valid results: 2013.6, Runtime: 59.6927 s, Costs: 0.1262127994999134 USD

With change:

> Overall benchmark of all runs, :aggregated_benchmark=Actor: parsel-crawler-benchmark, Valid results: 2013.3, Runtime: 61.2107 s, Costs: 0.14725829787564015 USD,

Pijukatel: Around 16 % cost increase seems pretty high to me. Maybe some optimization could be used to mitigate this increase.

How is it done in JS SDK?

vdusek: > What about the cost increase? Won't this lead to each request being locked at least twice?

Yes, it should be one extra RQ write per request. The 16% seems expected.

However, IMO better to be a bit more expensive and correct than cheaper and broken.

> Maybe we could lock only if the remaining time is something very small?

See the new commit. The lock is now prolonged only when its remaining windows are below half of the lock time.

> How is it done in the JS SDK?

It prolongs unconditionally on every fetch.

---

So please see the current state and try to execute the benchmarks again.

Pijukatel: Yes, much better. 

>  2026-08-19 07:14:07,839 - benchmark_logger - INFO - Overall benchmark of all runs, :aggregated_benchmark=Actor: parsel-crawler-benchmark, Valid results: 2013.7, Runtime: 67.4471 s, Costs: 0.12980576633404428 USD, .

Pijukatel (approved):

PR #1064, opened 2026-07-27
docs: Update guide on PPE pricing model

szaganek:
- Removes outdated info on Actors running indefinitely.
- Restructures the whole guide.
- Adds missing sections so both SDKs can mirror each other better (JS guide not updated yet)
 
Related to the recent updates to platform docs [#2770](https://github.com/apify/apify-docs/pull/2770).


docs/02_concepts/11_pay_per_event.mdx
```diff
@@ -1,84 +1,95 @@
...
     {ActorChargeSource}
 </RunnableCodeBlock>
 
-Then you just push your code to Apify and that's it! The SDK will even keep track of the max total charge setting for you, so you will not provide more value than what the user chose to pay for.
-
-If you need finer control over charging, you can call <ApiLink to="class/Actor#get_charging_manager">`Actor.get_charging_manager()`</ApiLink> to access the <ApiLink to="class/ChargingManager">`ChargingManager`</ApiLink>, which can provide more detailed information, for example how many events of each type can be charged before reaching the configured limit.
+For details on how to maximize your profits, see also [Best practices](https://docs.apify.com/platform/actors/publishing/monetize/pay-per-event#best-practices).
 
-### Handling the charge limit
+### Prefer per-unit charging
```
vdusek, line 31: You updated the main title from "pay-per-event" to "pay per event" but kept "per-unit charging" - is that intentional?

docs/02_concepts/11_pay_per_event.mdx
```diff
@@ -1,84 +1,95 @@
...
+## Transition from a different pricing model
 
-For comprehensive details on pay-per-event pricing and Actor monetization, see the [pay-per-event](https://docs.apify.com/platform/actors/publishing/monetize/pay-per-event) and [monetization](https://docs.apify.com/platform/actors/publishing/monetize) documentation on the Apify platform.
+When you plan to start using the pay-per-event pricing model for an Actor that is already monetized with a different pricing model, your source code must support both pricing models during the transition period enforced by the Apify platform. The most frequent case is the transition from the pay-per-result model which utilizes the `ACTOR_MAX_PAID_DATASET_ITEMS` environment variable to prevent returning unpaid dataset items. The following is an example how to handle such scenarios. The key part is the <ApiLink to="class/ChargingManager#get_pricing_info">`ChargingManager.get_pricing_info()`</ApiLink> method which returns information about the current pricing model.
+
+<RunnableCodeBlock className="language-python" language="python">
+    {ConditionalActorChargeSource}
+</RunnableCodeBlock>
```
vdusek, line 95: Shouldn't there be some conclusion?

vdusek (approved):
Two comments for consideration; otherwise LGTM.

docs/02_concepts/11_pay_per_event.mdx
szaganek: Yeah, I treated it as a proper name, but turns out it acts like a regular adjective, my bad. Lemme fix it to pay-per-event.

docs/02_concepts/11_pay_per_event.mdx
szaganek: About that conclusion, where does it come from? We don't use it anywhere in the docs except for here. Frankly, it's not really useful. Is it some legacy?

vdusek (changes requested):
let's resolve the conclusion thing first

docs/02_concepts/11_pay_per_event.mdx
vdusek: IIRC, it was already part of the existing content when I joined, so I just followed the established structure.

Removing it from a single page and leaving it on the others is inconsistent. If we decide that these sections are unnecessary, we should apply that change across the whole documentation.

As for whether they should be there at all, I'm in favor of keeping them. IMO learning-oriented content benefits from an ending that summarizes the key takeaways or points readers toward what to explore next.

szaganek: I brought it up at the docs meeting, we discussed it, and decided that we should get rid of the Conclusion section everywhere. The main reasons are consistency with the rest of the docs plus these sections don't add much value in the current form. These docs aren't a course to follow, they aren't read like a book. 

I'll review all these sections and move some of the content to the introductions, where it makes sense.

vdusek: Okay, please leave it in this PR for now and open a follow-up PR to remove them, since it's unrelated to these changes.

vdusek (approved):

PR #1068, opened 2026-07-29
docs: Delete Conslusions section from concepts

szaganek:
Deletes `Conclusions` section from all topics under `Concepts`. Applies to next and latest doc versions.


vdusek (commented):
Did you check that we are not losing any valuable information and that the conclusion contained only summaries/recaps?

Second, there is no need to duplicate this for the versioned v4 docs, we have a CI pipeline that handles the copying. Please remove the duplication.

szaganek:
@vdusek, I did check what I'm deleting from docs, no worries. Conclusions didn't carry any valuable information.

vdusek (approved):
LGTM

PR #1077, opened 2026-08-04
fix: close the platform events websocket iterator on shutdown

vdusek:
Since v4.0.0, an Actor logs `RuntimeError: aclose(): asynchronous generator is already running` right after `Actor.exit()`. Cosmetic - the exit code stays 0 and no data is lost - but it shows up in every run and looks like a failure.

`ApifyEventManager._process_platform_messages` iterates `websockets` `connect(...)` with `async for`. That iterator is an async generator, and neither `break` nor cancelling the task closes it, so the garbage collector does - during `asyncio.run` teardown, where the `aclose()` it schedules lands after `_cancel_all_tasks` took its task snapshot and is never awaited. `loop.shutdown_asyncgens()` then calls `aclose()` on a generator that is still running. Up to v3.4.1 the code used `async with connect(...)`, which unwound synchronously on cancellation.

`contextlib.aclosing` closes the iterator on every exit path - `break`, exception and cancellation alike. The rest of the diff is the loop body re-indenting, a unit test that asserts the iterator is closed, and an E2E test that asserts the error is absent from a run log.

Reported on [Discord](https://discord.com/channels/801163717915574323/1533837354018799808/1533837354018799808).

*✍️ Drafted by Claude Code*

### DIFF
--- a/src/apify/events/_apify_event_manager.py
+++ b/src/apify/events/_apify_event_manager.py
@@ -151,9 +151,8 @@ class ApifyEventManager(EventManager):
             # Used as an async iterator, `connect` reconnects with exponential backoff on failed connection attempts.
             connector = websockets.asyncio.client.connect(ws_url, process_exception=self._process_connection_exception)
 
-            # Close the iterator explicitly: neither `break` nor cancellation does it, and leaving it to the garbage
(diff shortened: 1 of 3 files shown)


Mantisus (approved):
LGTM

B4nan (approved):

PR #1078, opened 2026-08-04
fix: do not abort Actor exit when the terminal status message fails

vdusek:
In `_ActorType.__aexit__`, the terminal `set_status_message()` call was the only step in `finalize()` not wrapped in `try`/`except Exception`. A transient API error there propagated out of `__aexit__` and skipped everything after it: the event manager exit, the charging manager exit, `_save_actor_state()` and `sys.exit(self.exit_code)`. The run then died with an unrelated traceback, the wrong exit code, and unpersisted state.

The call is now guarded the same way as the three steps that follow it - the failure is logged and cleanup continues.

*✍️ Drafted by Claude Code*

### DIFF
--- a/src/apify/_actor.py
+++ b/src/apify/_actor.py
@@ -260,7 +260,10 @@ class _ActorType:
 
         async def finalize() -> None:
             if self.status_message is not None:
-                await self.set_status_message(self.status_message, is_terminal=True)
+                try:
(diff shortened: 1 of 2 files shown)


Mantisus (approved):
LGTM

B4nan (approved):

PR #1079, opened 2026-08-04
perf: allow concurrent dataset pushes by narrowing the push-path locks

vdusek:
Concurrent dataset pushes were fully serialized - N concurrent pushes cost N sequential API round-trips.

## Problem

Two locks wrapped the network I/O; either alone is enough to serialize, so both had to be narrowed:

- `Actor.push_data` held `charging_manager.charge_lock()` across the whole push + charge sequence.
- `ApifyDatasetClient.push_data` held its per-client `_lock` across the chunked `push_items` calls.

The charge lock only matters for pay-per-event runs (keeps the limit reservation and charge atomic). `ApifyDatasetClient.push_data` mutates no client state - `_serialize_chunk` is a classmethod, `push_items` is a stateless API call - so outside of charging, neither lock protects anything.

## Change

- Both call sites now go through one `charge_lock_if_charging()` helper in `_charging.py`, keyed off `charging_manager_ctx`.
- Replaces the near-duplicate `DatasetClientPpeMixin._charge_lock` and a second, divergent `is_pay_per_event` predicate - the "do we need the lock?" decision now lives in one place.
- Pay-per-event runs are unchanged and still fully serialized.

## Benchmark

`push_data` against the real API, 10 items/push, median of 5 rounds (`raw` = same payloads sent directly via `push_items`, the network floor with no lock involved):

| concurrency | master  | this PR | raw floor | speedup |
|---|---|---|---|---|
| 1  | 145 ms  | 130 ms | 130 ms | 1.1x |
| 4  | 619 ms  | 150 ms | 138 ms | 4.1x |
| 8  | 1109 ms | 153 ms | 141 ms | 7.3x |
| 16 | 2216 ms | 190 ms | 168 ms | 11.7x |
| 32 | 4502 ms | 188 ms | 186 ms | 24.0x |

Master's curve is `N x single-push latency` (full serialization); this PR flattens onto the raw floor. Re-adding just one lock (either one) around `push_data` reproduces master's numbers exactly, confirming both had to move.

## Behavior changes

- **Pushes and `drop()` are no longer mutually exclusive.** The lock only ever ordered them within one client instance in one process, so a concurrent drop was never actually safe. `_lock` still serializes `drop()` itself. Verified against the real API: 8 concurrent pushes racing a `drop()` finish in under a second, with the pushes failing `NotFoundError` once the drop wins, instead of silently completing first.
- **A push's chunks are no longer contiguous** in the dataset when other pushes run concurrently (16 concurrent multi-chunk pushes land as 224 interleaved blocks instead of 16). Item order *within* one push is still preserved (new test below); cross-push interleaving was never a documented guarantee.
- **Local file-system runs are unaffected** - still serialized by Crawlee's `FileSystemDatasetClient._lock`, which guards sequential item numbering and metadata and has no network round-trip behind it.

## Testing

- `test_concurrent_push_data_overlaps`, `test_push_data_does_not_take_charge_lock_without_pay_per_event` - each fails on `master`.
- `test_concurrent_actor_push_data_stays_within_budget` - fails if the charge lock is ever skipped for PPE runs.
- `test_concurrent_multi_chunk_pushes_preserve_per_push_order` - locks in per-push ordering on the new concurrent-chunk code path this PR enables.
- Stress-tested against the real API beyond unit tests: 2400-item/32-way concurrent integrity check (no loss, no dupes), 640-item multi-chunk check, 256-way soak (7680 items, 0 errors), and 270 concurrent-lifecycle PPE-budget iterations across explicit/synthetic/mixed charge paths - all stayed within budget, identically on master and this PR.

*✍️ Drafted by Claude Code*

### DIFF
--- a/src/apify/storage_clients/_apify/_dataset_client.py
+++ b/src/apify/storage_clients/_apify/_dataset_client.py
@@ -42,16 +42,11 @@ class ApifyDatasetClient(DatasetClient, DatasetClientPpeMixin):
         self,
         *,
         api_client: DatasetClientAsync,
-        lock: asyncio.Lock | None = None,  # noqa: ARG002
+        lock: asyncio.Lock | None = None,  # noqa: ARG002 - no-op, kept for backward compatibility
     ) -> None:
         """Initialize a new instance.
 
         Preferably use the `ApifyDatasetClient.open` class method to create a new instance.
-
-        Args:
-            api_client: The Apify dataset client for API operations.
-            lock: Unused - no operation performed by this client needs client-side locking anymore. Kept for
-                backward compatibility with existing call sites.
         """
         DatasetClient.__init__(self)
         DatasetClientPpeMixin.__init__(self)
--- a/src/apify/storage_clients/_apify/_dataset_client.py
+++ b/src/apify/storage_clients/_apify/_dataset_client.py
@@ -12,6 +12,7 @@ from crawlee.storage_clients._base import DatasetClient
 from crawlee.storage_clients.models import DatasetItemsListPage, DatasetMetadata
 
 from ._api_client_creation import create_storage_api_client
+from apify._charging import charge_lock_if_charging
 from apify.storage_clients._ppe_dataset_mixin import DatasetClientPpeMixin
 
 if TYPE_CHECKING:
@@ -54,7 +55,7 @@ class ApifyDatasetClient(DatasetClient, DatasetClientPpeMixin):
         """The Apify dataset client for API operations."""
 
         self._lock = lock
-        """A lock to ensure that only one operation is performed at a time."""
+        """A lock serializing destructive operations on the dataset."""
 
     @override
     async def get_metadata(self) -> DatasetMetadata:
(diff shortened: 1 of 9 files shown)


src/apify/storage_clients/_apify/_dataset_client.py
```diff
@@ -54,7 +55,7 @@ def __init__(
         """The Apify dataset client for API operations."""
 
         self._lock = lock
-        """A lock to ensure that only one operation is performed at a time."""
+        """A lock serializing destructive operations on the dataset."""
```
Mantisus: Does drop really need a lock now? If I'm not mistaken, it was there to keep the dataset from being dropped while a push was in progress.

tests/unit/storage_clients/test_apify_dataset_client.py
```diff
@@ -104,3 +105,48 @@ async def test_push_data_rejects_a_non_serializable_item() -> None:
...
+    received: list[tuple[int, int]] = []
+    chunk_count = 0
+
+    async def push_items(**kwargs: Any) -> None:
+        nonlocal chunk_count
+        chunk_count += 1
+        await asyncio.sleep(0)  # Yield so chunks from other concurrent pushes can land in between.
+        received.extend((item['push'], item['i']) for item in json.loads(kwargs['items']))
+
+    api_client.push_items = push_items
```
Mantisus: Let's try doing this using `unittest.mock.patch`, but I won't insist 🙂

Mantisus (approved):
LGTM. Just a couple of nits

src/apify/storage_clients/_apify/_dataset_client.py
vdusek: It shouldn't. Removing it and keeping the input argument as a no-op.

B4nan (approved):

PR #1052, opened 2026-07-16
perf: enable brotli request compression by default

vdusek:
Enables brotli compression as the default for all SDK/client ↔ platform communication.

apify-client v3 defaults to `gzip`, so enabling brotli takes two parts:

- **Dependency**: pull in the `apify-client[brotli]` extra (`apify-client[brotli]>=3.1.0,<4.0.0`) so the brotli backend is always installed. v3.1 is the first release to add request body compression.
- **Code**: both places where the SDK constructs an `ApifyClientAsync` (`Actor.new_client` and the storage-client factory `_create_api_client`) now pass `compression='brotli'`. Since the SDK hard-depends on `apify-client[brotli]`, the brotli backend is guaranteed present, so brotli is passed directly (no availability check).

Tests: a wire-level test drives a real request through an SDK-created client against a local HTTP server and asserts the body goes out with `Content-Encoding: br` and round-trips back through `brotli.decompress`. It's parametrized over both client-creation paths (`_create_api_client` and `Actor.new_client`).

Closes #1045

*✍️ Drafted by Claude Code*

PR #1081, opened 2026-08-04
perf: serialize dataset push payloads once per chunk

vdusek:
Pushing to a dataset serialized every item in its own `asyncio.to_thread` hop (through Crawlee's `json_dumps`), so a push paid one thread round-trip per item. That helper also pretty-prints with `indent=2`, and each payload was UTF-8 encoded twice — once for the per-item size check, once again while chunking.

`_check_and_serialize` and `_chunk_by_size` are replaced by a single blocking `_serialize_chunk(items, offset)` that fills one size-bounded JSON array and returns the next offset. `push_data` drives it with one `asyncio.to_thread` per chunk and pushes each chunk before serializing the next, so payloads are still streamed rather than all materialized up front.

Serialization is now local and compact (`json.dumps(..., separators=(',', ':'))`) rather than Crawlee's `json_dumps`, whose `indent=2` is deliberate for the on-disk files people actually read. Nothing reads the wire format. Dropping the indent also lets CPython take the C encoder fast path, which it only does when `indent is None`.

Measured on 20k realistic items against a mocked API client:

|        | time   | payload  |
| ------ | ------ | -------- |
| before | 1.081s | 10.78 MB |
| after  | 0.075s | 9.34 MB  |

The payload saving is shape-dependent — larger for small or deeply nested items — and compounds with the brotli work in #1052.

Two spurious API calls disappear as a side effect: the old chunker emitted an empty `[]` chunk whenever the first item of a chunk landed within 2 bytes of the limit, and it pushed `[]` rather than nothing when PPE charging drove the push limit down to 0.

Stored items are unaffected. The only observable difference is that raw request bodies are no longer pretty-printed.

*✍️ Drafted by Claude Code*

### DIFF
--- a/src/apify/storage_clients/_apify/_dataset_client.py
+++ b/src/apify/storage_clients/_apify/_dataset_client.py
@@ -214,9 +214,8 @@ class ApifyDatasetClient(DatasetClient, DatasetClientPpeMixin):
     def _serialize_chunk(cls, items: Sequence[Mapping[str, JsonSerializable]], offset: int) -> tuple[str, int]:
         """Serialize items starting at `offset` into one JSON array staying within the payload size limit.
 
-        The array holds as many consecutive items as fit below `_EFFECTIVE_LIMIT_SIZE`, always at least one.
-        The result is compact JSON - it goes on the wire, so indentation and separator padding would be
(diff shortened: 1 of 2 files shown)


Mantisus (approved):
LGTM

B4nan (approved):

PR #1082, opened 2026-08-06
fix(events): open the platform websocket only once per context

vdusek:
`ApifyEventManager` ignored the re-entrancy contract of Crawlee's `EventManager`, which tracks active contexts with `_active_ref_count`. `BasicCrawler._run_crawler` always enters the *global* event manager, and `Actor.init` registers the `ApifyEventManager` as exactly that and has already entered it — so on the platform every crawler run re-entered it and opened a second websocket.

Measured on master, with a local events server and a `BasicCrawler.run()` inside an entered `ApifyEventManager`:

- Two connections stay open for as long as the run lasts, so every platform event arrives twice. One `Migrating` became two `Migrating` listener calls and two `PersistState(is_migrating=True)` — state persisted twice per migration. The extra connection also counts against the platform limit of 10 per run.
- The crawler's exit tears down that newer connection, which leaves `__aexit__` bookkeeping pointing at an already-closed one. The Actor's own connection and its message-processing task therefore survive `Actor.exit()`, re-breaking the iterator shutdown from #1077.

The websocket is now owned by the outermost context only, mirroring `LocalEventManager`. On top of that:

- The shutdown moved into `_teardown_platform_websocket()` and the parent context is left in a `finally`, so a failed shutdown can no longer keep the manager active for good — which would mean never emitting `PersistState` again.
- Task, connection and future are reset on exit, so the context can be entered again.
- A cancelled entry cleans up as well. Without it, a later entry mistook itself for a nested context and silently returned a manager that received no platform events at all.
- The error that prevented the first connection is raised as the `__cause__` of the `RuntimeError`, not only logged.
- `Actor.reboot()` reads `_listeners_to_wrappers` with `get`, so the lookup no longer inserts entries for events nobody listens to.

Adopts the patterns from apify/crawlee-python#2100 (Crawlee 1.9.1).

*✍️ Drafted by Claude Code*

### DIFF
--- a/src/apify/_actor.py
+++ b/src/apify/_actor.py
@@ -1249,12 +1249,10 @@ class _ActorType:
             # the reboot. Typically, crawlers are listening for the MIGRATING event to stop processing new requests.
             # We can't just emit the events and wait for all listeners to finish,
             # because this method might be called from an event listener itself, and we would deadlock.
-            persist_state_listeners = flatten(
-                (self.event_manager._listeners_to_wrappers[Event.PERSIST_STATE] or {}).values()  # noqa: SLF001
-            )
-            migrating_listeners = flatten(
-                (self.event_manager._listeners_to_wrappers[Event.MIGRATING] or {}).values()  # noqa: SLF001
-            )
+            # Read the mapping with `get` - subscripting it would insert entries for events nobody listens to.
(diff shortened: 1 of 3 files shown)


Mantisus (approved):
LGTM

B4nan (approved):

PR #1087, opened 2026-08-11
fix: verify shared request queue is_finished against per-request reads

### LINKED ISSUE #808: Fetch next request doesn't return just-added or reclaimed requests in shared RQ
## Context

Discovered during review of PR #803 ([discussion](https://github.com/apify/apify-sdk-python/pull/803#discussion_r2826493851)).

The `test_request_reclaim_functionality[shared]` integration test is flaky because `fetch_next_request()` does not immediately return requests that were just added or reclaimed when using the **shared** request queue access mode. In the **single** mode, requests are added directly to the local `queue_head` deque and are available right away, but the shared client cannot do that.

## Problem

When using shared request queue mode, there is a propagation delay between:
1. Adding a request (or reclaiming one) via the API
2. That request becoming visible in `fetch_next_request()`

This means users of the shared request queue mode may call `fetch_next_request()` and get `None` even though a request was just added or reclaimed.

## Expected behavior

`fetch_next_request()` should reliably return requests that have been added or reclaimed, even in shared mode, without requiring the caller to implement their own retry/polling logic.

## Current workaround

Callers need to poll `fetch_next_request()` with retries and delays to handle the propagation delay.

vdusek:
Closes: #808

`test_request_queue_is_finished_and_is_empty[shared]` failed on master
([run 31375518173](https://github.com/apify/apify-sdk-python/actions/runs/31375518173/job/93413747584))
with `assert not True`.

In shared access mode `is_finished()` answered solely from the queue head listing, which is eventually
consistent: right after a committed `add_request` the head can come back empty with no locked requests, so the
queue briefly reported itself as finished. That is a client defect, not just a test problem - a crawler polling
`is_finished()` in that window would shut down with work still in the queue.

`is_finished()` now confirms an empty-head verdict before reporting `True`: every request this client knows
about and has not yet seen handled is re-checked with a strongly consistent `get_request` read. A request that
is missing (not yet propagated) or unhandled (pending, or locked by another client) keeps the queue unfinished;
confirmed-handled requests are remembered, so each one is verified at most once. An in-flight
`add_batch_of_requests` call also keeps the queue unfinished. Only the shared client changes - the single
client already tracks added requests in its local head estimation, so it is consistent locally.

The test-side polling from the previous revision is reverted: the strict negative `is_finished` assertions are
back and serve as regression coverage for the client-side guarantee.

Verified by injecting the propagation lag deterministically (first head listing after each add returns an
empty, lock-free head without hitting the API): the unfixed client reproduces the exact CI failure, the fixed
one passes 10/10 runs with strict assertions. New unit tests cover the verification paths, and the whole
integration module passes under `--numprocesses=auto`.

*✍️ Drafted by Claude Code*


Pijukatel:
Maybe the `is_finished` when returning `True`, should be double-checked(or inspected in other way) due to eventual consistency to prevent something like this:
https://apify.slack.com/archives/C0L33UM7Z/p1785322628993809

src/apify/storage_clients/_apify/_request_queue_shared_client.py
```diff
@@ -344,7 +344,37 @@ async def is_finished(self) -> bool:
...
+            if cached_request.was_already_handled:
+                continue
+
+            request = await self._get_request_by_id(request_id)
+            if request is None or request.handled_at is None:
+                return False
+
+            cached_request.was_already_handled = True
+
+        return True
```
Pijukatel: Discussion points:

1. In a normal scenario, are there many requests being double-checked like this, or will most of the requests already short-circuit this function due to `cached_request.was_already_handled`?
2. Maybe it would make sense to check those requests in parallel instead of one by one?
(not to be blocked by `await self._get_request_by_id(request_id)` in for loop)

vdusek: 1 - If I'm reading it correctly, most of the requests should be short-circuited. The normal lifecycle (`add_requests` -> `list_head` -> `fetch_request` -> `mark_as_handled`) ends up cached as handled. So this path shouldn't need a recheck.

2 - Yeah, good point. I used `asyncio.gather` with batches of size 10.

Pijukatel (approved):

PR #1097, opened 2026-08-21
fix(scrapy): stop silently dropping in-flight requests and redirects

vdusek:
Two request-loss bugs in the Scrapy integration, both from `next_request()` marking a request handled the moment it is fetched, before Scrapy has downloaded anything.

**In-flight requests were lost on migration.** Everything in the downloader or the parse pipeline was already `handled_at`-stamped, and `reclaim_request` was never called anywhere in the integration. A request now stays unresolved until Scrapy is done with it — tracked through `downloader.active` and `scraper.slot.active`, so the callback and the item pipeline are covered too. Those internals are undocumented, so `open()` verifies them once and fails loudly if a Scrapy version stops exposing them. Reconstruction failures are still consumed immediately.

Killing a process with three requests in flight: 3 handled / 0 pending before, all three lost; 0 handled / 3 pending after, all fetchable next run.

**Every redirect on a discovered link was dropped.** `RedirectMiddleware` derives the new request with `request.replace(url=...)`, which inherits `meta['apify_request_unique_key']`, so `add_request` answered `was_already_present=True` and `enqueue_request` returned `False` — a silent drop with no error. `to_scrapy_request` now also stamps `meta['apify_request_fingerprint']` (URL, method, body), and the key is reused only while that fingerprint matches. Headers are left out: `DefaultHeadersMiddleware` and `UserAgentMiddleware` mutate them in place before the scheduler sees the request again.

**Reactor stalls.** `next_request()` did two serial blocking API round trips per request on the reactor thread; now one. Marks go through a fire-and-forget `AsyncThread.submit_coro`, drained before `is_finished()` is consulted and before the event loop closes; one that fails after dispatch is retried on a later pass. A hard kill can still lose a mark, leaving that request pending for the next run — at-least-once is the safe direction.

*✍️ Drafted by Claude Code*


docs/03_guides/06_scrapy.mdx
```diff
@@ -104,7 +104,7 @@ The following example shows a Scrapy Actor that scrapes page titles and enqueues
 
 ## Dealing with imminent migration to another host
 
-Under some circumstances, the platform may decide to [migrate your Actor](https://docs.apify.com/academy/expert-scraping-with-apify/migrations-maintaining-state) from one piece of infrastructure to another while it's in progress. While [Crawlee](https://crawlee.dev/python)-based projects can pause and resume their work after a restart, achieving the same with a Scrapy-based project can be challenging.
+Under some circumstances, the platform may decide to [migrate your Actor](https://docs.apify.com/academy/expert-scraping-with-apify/migrations-maintaining-state) from one piece of infrastructure to another while it's in progress. Requests that Scrapy hasn't finished when the run stops stay unhandled in the request queue, so the next run picks them up and downloads them from scratch. A Scrapy-based project doesn't resume where it left off the way a [Crawlee](https://crawlee.dev/python)-based one does, so items their callbacks already pushed can land in the dataset twice.
```
szaganek:
```suggestion
Under some circumstances, the platform may decide to [migrate your Actor](https://docs.apify.com/academy/expert-scraping-with-apify/migrations-maintaining-state) from one piece of infrastructure to another while the run is in progress. Requests that Scrapy hasn't finished when the run stops stay unhandled in the request queue, so the next run picks them up and downloads them from scratch. A Scrapy-based project doesn't resume where it left off the way a [Crawlee](https://crawlee.dev/python)-based one does, so items that their callbacks already pushed can land in the dataset twice.
```
Two tiny things to make it read better

szaganek (approved):

Pijukatel (approved):

PR #1104, opened 2026-08-25
fix(scrapy): stop the crawl gracefully when the Actor run is migrated or aborted

### LINKED ISSUE #1103: Scrapy: stop the crawl gracefully on migration and abort to avoid duplicate items
## Problem

The Scrapy integration does not react to the `MIGRATING` and `ABORTING` platform events, so the process is killed with requests in flight. Since #1097 those requests stay unhandled in the request queue and the next run downloads them again, which is correct, but every item their callbacks had already pushed lands in the dataset a second time. The Scrapy guide lists this as a known limitation with the `HTTPCACHE_*` workaround (apify/actor-templates#303); that only makes the re-download cheap, it does not prevent the duplicates.

Crawlee handles exactly this case: on `MIGRATING` / `ABORTING` it pauses the pool for up to 20 s (`SAFE_MIGRATION_WAIT_MILLIS`) so that in-flight requests finish and get marked as handled.

## Proposal

1. Add a Scrapy extension, registered by `apply_apify_settings`, that subscribes to `Event.MIGRATING` and `Event.ABORTING` on `spider_opened` and calls `await crawler.stop_async()` (Scrapy >= 2.14). The engine stops scheduling new requests, waits for the in-flight ones including their callbacks and item pipelines, and only then closes the scheduler, whose `close()` marks everything finished as handled. Only callbacks still running at the hard kill can still produce duplicates.
2. Document that the default `Spider.start()` yields the start URLs with `dont_filter=True`, which the integration maps to `always_enqueue=True`, so the start URLs are crawled again after every restart. Overriding `start()` with plain `Request`s makes them deduplicate against the request queue.
3. Rewrite the "Dealing with imminent migration to another host" section of the Scrapy guide accordingly.

*✍️ Drafted by Claude Code*

vdusek:
Closes #1103.

The Scrapy integration ignored `MIGRATING` and `ABORTING`, so the process was killed with requests in flight. Those stay pending, the next run downloads them again, and their items land in the dataset twice.

## Why not `crawler.stop_async()` for both events

- A container exiting with code 0 during a migration ends the run as SUCCEEDED and it isn't restarted (only a non-zero exit with `wasMigrationInitiated` re-schedules it), so stopping the crawl on `MIGRATING` would lose the pending work.
- On a graceful abort the run is already DONE, so exiting with 0 yields ABORTED.

## Migration

- `ApifyScheduler` listens for `MIGRATING`: `next_request()` hands out nothing, and the requests in flight are marked as handled as they finish.
- The crawl stays idle until the platform kills it, so the next run continues with the pending requests only.
- `Actor.reboot()` awaits the same listener, so a reboot waits for the requests in flight to settle first.

## Abort

- New `ApifyGracefulStopExtension` (registered by `apply_apify_settings`) calls `crawler.stop_async()` on `ABORTING`.
- The scheduler marks the requests in flight as handled as they finish, and the run ends as ABORTED within the grace period.

## Docs and tests

- The Scrapy guide describes the migration behavior and the start-URL caveat: the default `Spider.start()` yields with `dont_filter=True`, so a restarted run crawls the start URLs again. The example spider overrides `start()` with plain requests.
- The start page now deduplicates against the same link on subpages, so the title-spider e2e test asserts unique items.
- New e2e tests: a graceful abort with a request in flight, and an `Actor.reboot()` mid-crawl.

*✍️ Drafted by Claude Code*


docs/03_guides/06_scrapy.mdx
```diff
@@ -47,6 +47,7 @@ The Apify SDK provides several custom components to support integration with the
...
 - <ApiLink to="class/ApifyCacheStorage">`apify.scrapy.extensions.ApifyCacheStorage`</ApiLink> - A storage backend for Scrapy's built-in [HTTP cache middleware](https://docs.scrapy.org/en/latest/topics/downloader-middleware.html#module-scrapy.downloadermiddlewares.httpcache). This backend uses Apify's [key-value store](https://docs.apify.com/platform/storage/key-value-store). To enable caching, set `HTTPCACHE_ENABLED` and `HTTPCACHE_EXPIRATION_SECS` in your settings. By default, when the spider closes, up to 100 expired and unreadable entries per run are cleaned up. To change this number, update `APIFY_HTTPCACHE_EXPIRATION_MAX_ITEMS`.
+- <ApiLink to="class/ApifyGracefulStopExtension">`apify.scrapy.extensions.ApifyGracefulStopExtension`</ApiLink> - A Scrapy [extension](https://docs.scrapy.org/en/latest/topics/extensions.html) that stops the crawl gracefully when the Actor run is aborted, so the requests in flight finish and get marked as handled. For details, see [Dealing with imminent migration to another host](#dealing-with-imminent-migration-to-another-host).
```
Pijukatel: `so the requests in flight finish`

This creates an impression that the platform will wait for them to finish, but in reality it just gives the Actor some time (30s?) to finish.

src/apify/scrapy/extensions/_graceful_stop.py
```diff
@@ -0,0 +1,54 @@
...
+logger = getLogger(__name__)
+
+
+class ApifyGracefulStopExtension:
+    """A Scrapy extension that stops the crawl gracefully when the Actor run is being aborted.
+
+    A graceful abort gives the run a moment before it is killed. The extension uses it to stop the engine: no new
+    requests are started, the ones in flight finish along with their callbacks and item pipelines, and the
+    scheduler marks them as handled in the request queue, so resurrecting the run does not download them again.
+    A migration is handled by `ApifyScheduler` instead, as the crawl must not finish on its own then.
```
Pijukatel: `as the crawl must not finish on its own then`

What does this tries to tell the reader? Is it ensured that the run will not finish on its own, or should the reader somehow enforce it?

Pijukatel (approved):
Just some nits on comments and docs

PR #1106, opened 2026-08-27
test: Speed up the unit, integration and E2E test suites

vdusek:
The CI critical path is the E2E job at ~5.9 min; it and the unit suite mostly wait rather than work (master run `32966697702`).

**E2E**

- `sdk_wheel_path` locks the wheel build on a shared file, but keeps the "already built" marker under the **per-worker** `getbasetemp()`, so all 16 workers rebuild the wheel in turn behind the lock (first worker's first test at 41s, last at 144s). The marker moves next to the lock, into the shared parent dir.
- `test_crawlee`/`test_scrapy` hold the 6 slowest tests but collect last, and xdist distributes in collection order - `test_adaptive_playwright_crawler` (167s) ran t=155s to 322s, ~100s after every other worker went idle. A `pytest_collection_modifyitems` hook puts them first.

**Unit**

- `test_reboot_proceeds_when_event_listener_exceeds_timeout`: 30.1s of 80.5s, because its listener slept 60s and leaving the `Actor` context waited out `cleanup_timeout`. It releases the listener via `asyncio.Event` instead (-> 0.17s).
- 12 tests slept 1s each for mtime granularity; they backdate the input file with `os.utime`.
- `tests_concurrency` 1 -> 4 (the suite ran at ~21% CPU on one worker), plus `--dist worksteal` so a worker that draws several slow tests stops being the critical path.

Locally 81.0s -> 15.5s at 4 workers, 572 passing. Projected CI: unit ~2 -> ~1 min, E2E 5.9 -> ~4 min. E2E is unverified locally (no platform token), so the first CI run is the check.

*✍️ Drafted by Claude Code*

### DIFF
--- a/.github/workflows/_checks.yaml
+++ b/.github/workflows/_checks.yaml
@@ -139,7 +139,7 @@ jobs:
       operating_systems: '["ubuntu-latest", "windows-latest"]'
       python_version_for_codecov: "3.14"
       operating_system_for_codecov: ubuntu-latest
-      tests_concurrency: "1"
+      tests_concurrency: "8"
 
   integration_tests:
     name: Integration tests (${{ matrix.python-version }}, ${{ matrix.os }})
@@ -161,7 +161,7 @@ jobs:
     runs-on: ${{ matrix.os }}
 
     env:
-      TESTS_CONCURRENCY: "8"
+      TESTS_CONCURRENCY: "16"
 
     steps:
       - name: Checkout repository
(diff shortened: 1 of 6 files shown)


.github/workflows/_checks.yaml
```diff
@@ -139,7 +139,7 @@ jobs:
       operating_systems: '["ubuntu-latest", "windows-latest"]'
       python_version_for_codecov: "3.14"
       operating_system_for_codecov: ubuntu-latest
-      tests_concurrency: "1"
+      tests_concurrency: "4"
```
Pijukatel, line 142: I am worried that running them in parallel might cause unpredictable platform usage-related flakiness; not sure about it, though

Pijukatel (approved):

.github/workflows/_checks.yaml
vdusek: this is only for the unit tests

PR #1123, opened 2026-09-08
feat(charging): Support the `idempotency_key` param in `Actor.charge()`

### LINKED ISSUE #1122: Support the `idempotency_key` param in `Actor.charge()`
The Python client seems to [support this param](https://github.com/apify/apify-client-python/blob/f52190d4d49042369c9c293475eebaa657766d6d/src/apify_client/_resource_clients/run.py#L386) already, so it's just the matter of wiring it right. 

Parity tracking for https://github.com/apify/apify-sdk-js/pull/724

Mantisus:
Adds support for the `idempotency_key` parameter in `Actor.charge()`.

Closes: #1122

### DIFF
--- a/src/apify/_charging.py
+++ b/src/apify/_charging.py
@@ -451,6 +451,8 @@ class ChargingManagerImplementation(ChargingManager):
                         f"'{previous.event_name}', so it cannot be reused for event '{event_name}'."
                     )
 
+                logger.debug(f"Skipped a repeated charge of event '{event_name}' under key '{idempotency_key}'.")
+
                 return ChargeResult(
                     event_charge_limit_reached=self.is_event_charge_limit_reached(event_name),
                     charged_count=previous.charged_count,
--- a/src/apify/_charging.py
+++ b/src/apify/_charging.py
@@ -210,7 +210,7 @@ class ChargingManager(Protocol):
     charge_lock: ReentrantLock
     """Lock to synchronize charge operations. Prevents race conditions between `charge` and `push_data` calls."""
 
-    async def charge(self, event_name: str, *, count: int = 1) -> ChargeResult:
+    async def charge(self, event_name: str, *, count: int = 1, idempotency_key: str | None = None) -> ChargeResult:
         """Charge for a specified number of events - sub-operations of the Actor.
 
         This is relevant only for the pay-per-event pricing model.
@@ -218,6 +218,9 @@ class ChargingManager(Protocol):
         Args:
             event_name: Name of the event to be charged for.
             count: Number of events to charge for.
+            idempotency_key: A unique key preventing a retried operation from being charged for twice. A repeat
+                under the same key is not sent to the API and reports the `charged_count` of the original call.
+                A key belongs to a single event and is only remembered for the lifetime of the run.
         """
 
     def calculate_total_charged_amount(self) -> Decimal:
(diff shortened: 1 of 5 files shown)


src/apify/_charging.py
```diff
@@ -508,9 +505,9 @@ async def charge(self, event_name: str, *, count: int = 1, idempotency_key: str
             self._charging_state[event_name].charge_count += charged_count
             self._charging_state[event_name].total_charged_amount += charged_count * pricing_info.price
 
-            # Only remember a key that stands for a charge the platform actually received. Off the platform there
-            # is no request at all and the registry is the only thing providing deduplication.
-            if idempotency_key is not None and (not self._is_at_home or charge_sent):
+            # Remember the key for every charge that was counted, including events the API never receives, such as
+            # synthetic and tier-priced ones - those are counted locally and a repeat would count them twice.
+            if idempotency_key is not None:
```
Mantisus, line 510: With this update, the `idempotency_key` will be saved even for events that do not exist.

Mantisus:
@barjin, Please take a look at the implementation.

In the TS SDK, the `idempotency_key` doesn't affect the `ChargingManager`. This will cause desynchronization in cases where the API rejects a request with a duplicate `idempotency_key`.

src/apify/_charging.py
```diff
@@ -435,6 +443,24 @@ async def charge(self, event_name: str, *, count: int = 1) -> ChargeResult:
             )
 
         async with self.charge_lock():
+            # A repeat is resolved from this registry rather than left to the platform, whose own idempotency
+            # record expires after a few minutes: a late repeat would charge a second time, and counting it here
+            # would inflate the charging state and make the run hit `max_total_charge_usd` early.
```
barjin: This is surprising, but true (the API implementation has a 3-minute expiration window on the idempotency records ([source](https://github.com/apify/apify-core/blob/44eb8d1d1c2c123f9ca22056329a8bc3959777e7/src/api/src/lib/run_charging_service.ts#L57))).

I think we should follow the Platform behaviour here and allow for this idempotency "cooldown" period. This should also allow us to remove the old local idempotency records.

src/apify/_charging.py
```diff
@@ -210,14 +210,18 @@ class ChargingManager(Protocol):
...
 
         This is relevant only for the pay-per-event pricing model.
 
         Args:
             event_name: Name of the event to be charged for.
             count: Number of events to charge for.
+            idempotency_key: A unique key preventing a retried operation from being charged for twice. A repeat
+                under the same key is not sent to the API and reports the `charged_count` of the original call.
+                Keys are remembered for the lifetime of the Actor process. A key belongs to a single event, so
+                reusing one for a different event raises `ValueError`, as does passing a blank key.
```
barjin: > A key belongs to a single event, so reusing one for a different event raises `ValueError`

This does sound logical, but it's not how the API works. Should we implement this logic here?

barjin (commented):
Thank you @Mantisus ! 

Good catch, we need this in JS SDK as well, otherwise the deduplicated charges will eventually desync the running total count. I opened an [issue here](https://github.com/apify/apify-sdk-js/issues/733).

Otherwise, I left a few ideas regarding the implementation ⬇️  Thanks again!

src/apify/_charging.py
Mantisus: I agree. Let's align this behavior with the platform's behavior.

Mantisus:
@vdusek, could you recheck this, please?

vdusek (approved):
LGTM (I just polished some docs content :slightly_smiling_face:)

vdusek:
@barjin Do you want to recheck this? :slightly_smiling_face:

PR #1130, opened 2026-09-18
fix(scrapy): restore apify.scrapy imports on crawlee 1.10.1

l2ysho:
> [!NOTE]
> **TL;DR** while fixing CLI [pipeline](https://github.com/apify/apify-cli/actions/runs/35216295708/job/105185681004) claude brought me here. This is just one-shot fix, probably just starting point (?). It make sense to fix it here as version before was working, or should I fix downstream rather?

## The break

crawlee [#2208](https://github.com/apify/crawlee-python/pull/2208) changed the signature, shipped in [1.10.1](https://github.com/apify/crawlee-python/releases/tag/v1.10.1) on 2026-09-16:

https://github.com/apify/crawlee-python/blob/c120fdf279bdc7db712302ca60613ae9e7d3ad9e/src/crawlee/_utils/try_import.py#L15

We allow `crawlee>=1.8.0,<2.0.0`, and `crawlee._utils` is private, so nothing stopped it reaching users:

```
File ".../site-packages/apify/scrapy/__init__.py", line 8, in <module>
    with _try_import(__name__, 'run_scrapy_actor'):
TypeError: try_import() missing 1 required keyword-only argument: 'extra_name'
```

The Actor dies before it can write anything. It has been red on every run and every matrix leg in apify-cli CI since 2026-09-16 14:33 UTC, after passing on every run before 2026-09-16 01:05 UTC ([last pass](https://github.com/apify/apify-cli/actions/runs/35042657434), [first fail](https://github.com/apify/apify-cli/actions/runs/35109208583)). The `python-scrapy` Actor template is broken the same way.

## Why CI did not see it

`uv.lock` pinned crawlee 1.10.0, so no job here ever installed 1.10.1. This PR bumps it, which turns the lock into the regression guard: on 1.10.1 with the source fix reverted, `tests/unit/scrapy` fails with 52 collection errors.

## Verification

- `uv run pytest tests/unit` — 586 passed on crawlee 1.10.1.
- `uv run ruff format --check && uv run ruff check` — clean.
- Reverting only `src/apify/scrapy/__init__.py` on this branch — 52 errors, all `TypeError: try_import() missing 1 required keyword-only argument`.
- End to end with the `python-scrapy` Actor template: on crawlee 1.10.0 the crawl completes and exits 0 with 10 pages scraped; on 1.10.1 without this fix the Actor never starts and writes no dataset.

## Open question — the hint names the wrong package

`_get_install_hint` hardcodes `crawlee` as the distribution:

https://github.com/apify/crawlee-python/blob/c120fdf279bdc7db712302ca60613ae9e7d3ad9e/src/crawlee/_utils/try_import.py#L32-L37

So with `extra_name='scrapy'`, a user without scrapy installed now gets:

```
No module named 'scrapy'. Install the optional 'scrapy' extra to use it: pip install 'crawlee[scrapy]'
```

crawlee has no `scrapy` extra — it is not in `provides_extra` — and `uv pip install 'crawlee[scrapy]'` silently installs nothing, so the advice leaves the user exactly where they started. The correct advice is `apify[scrapy]`.

Two ways to settle it, and I did not want to pick one unilaterally:

1. Land this as the hotfix now, and change crawlee to take the distribution name (or a ready-made hint) so downstream users get a correct message.
2. Stop using crawlee's private util here and keep a small local `try_import` in the SDK, which removes the coupling for good.

Left as a draft for that reason — the code change itself is ready.


l2ysho:
@vdusek can you pls take a look if this make sense? Either way after last release CLI pipeline is red, if this is BS and I should fix downstream, let me know.

Pijukatel:
I would probably deal with this in the downstream repo where it was broken:
https://github.com/apify/crawlee-python/pull/2247
Then we can make just a tiny update here and bump directly to 1.10.2

@vdusek is on vacation for a few more days.

janbuchar:
:point_up_2: agreed, let's release that PR and then update deps here to not allow installing the release where the parameter is not optional.

Pijukatel:
Closing in favor of https://github.com/apify/apify-sdk-python/pull/1136

l2ysho:
thx guys 🚀  when do we plan to release 1.10.2?

